PR #1, opened 2022-11-09
Full Fledged CLI and Library

Spartee:
## Description

While a generic script is useful, there was alot of opportunity to stop duplicating the same functions we've written many times. 

The goal of this is to be able to use this library across our projects and use the cli where and when needed to reduce friction.

### Schema

This also adds the ability to create flexible custom schema based on a yaml file. these fields largely follow what the ``redis-py`` library expects.

Here is a sample 
```
index:
  name: sample
  storage_type: hash
  prefix: vector

fields:
  tag:
    categories:
      separator: "|"
    year:
      separator: "|"
  vector:
    vector:
      datatype: "float32"
      algorithm: "flat"
      dims: 768
      distance_metric: "cosine"
```

See the new README for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,42 +28,44 @@ into a yaml file for the cli to read. The format of the schema is as follows
 
 ```yaml
 index:
-  name: sample
+  name: sample # index name used for querying
   storage_type: hash
-  prefix: vector
+  key_field: "id" # column name to use for key in redis
+  prefix: vector  # prefix used for all loaded docs
 
+# all fields to create index with
+# sub-items correspond to redis-py Field arguments
 fields:
   tag:
-    categories:
+    categories: # name of a tag field used for queries
       separator: "|"
-    year:
+    year: # name of a tag field used for queries
       separator: "|"
   vector:
-    vector:
+    vector: # name of the vector field used for queries
       datatype: "float32"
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -1,65 +1,89 @@
-# RediSearch Data Loader
-The purpose of this script is to assist in loading datasets to a RediSearch instance efficiently.
+# RedisVL
 
-The project is brand new and will undergo improvements over time.
+A CLI and Library to help with loading data into Redis specifically for
+usage with RediSearch and Redis Vector Search capabilities
```
tylerhutcherson, line 4: We should back link to our docs probably!

redisvl/cli/load.py
```diff
@@ -0,0 +1,60 @@
...
+
+        # TODO more data types
+        # TODO try/expect and clear data upon failure
+        asyncio.run(self.load_pandas(args))
+
+    async def load_pandas(self, args):
+
+        # Create Redis Connection
+        logger.info(f"Connecting to {args.host}:{str(args.port)}")
+        redis_conn = get_async_redis_connection(args.host, args.port, args.password)
```
tylerhutcherson: would be good for us to also support env vars too (eventually)

PR #3, opened 2022-11-11
readme and param defaults update

tylerhutcherson:
- Updated `pip install` and examples in README to point to `sample-data/`
- Updated default value for `initial_cap` param on Vector field
- Updated index deletion function to allow for future behavior customization


Spartee (approved):
LGTM

Issue #11, opened 2022-11-21
Getting started example


Spartee:
### Description

Jupyter Notebook that takes one of the current simple test examples and walks through using ``redisvl`` as a python library and a cli.

### Acceptance Criteria

- [x] Well explained notebook with example usage for simple data in a jupyter notebook.

Issue #6, opened 2022-11-21
Implement redisvl index command in CLI


Spartee:
## Description

Implement the index management command for creating, reading, and deleting an index.

### Interface

```
usage: redisvl index <command> [<args>]

Index creation and manipulation commands

Commands:
        list        List available indices
        create      Create a new index
        delete      delete an existing index


positional arguments:
  command     Subcommand to run

optional arguments:
  -h, --help  show this help message and exit
  -s SCHEMA, --schema SCHEMA
                        Path to schema file
  --host HOST           Redis host
  -p PORT, --port PORT  Redis port
  -a PASSWORD, --password PASSWORD
```

Will have to wait until ``SearchIndex`` class in the library is completed in #7 

### Acceptance Criteria

- [ ] ``redisvl index list``
- [ ] ``redisvl index create``
- [ ] ``redisvl index delete``

PR #14, opened 2023-01-10
redisvl index command

Spartee:
This PR adds the ``redisvl index <command>`` into the CLI as well as a starter Jupyter notebook example.


Spartee:
will close #11 and #6

PR #16, opened 2023-07-11
OpenAI and Huggingface Providers

Spartee:
Providers for creating embeddings with OpenAI and Hugginface.

Includes
- base class
- simple integration testing
- async methods for clients that support it.

### DIFF
--- a/redisvl/providers/__init__.py
+++ b/redisvl/providers/__init__.py
@@ -1,0 +1,9 @@
+
+from redisvl.providers.openai import OpenAIProvider
+from redisvl.providers.huggingface import HuggingfaceProvider
+
+
+__all__ = [
+    "OpenAIProvider",
+    "HuggingfaceProvider"
+]
--- a/redisvl/providers/base.py
+++ b/redisvl/providers/base.py
@@ -1,0 +1,43 @@
+from typing import Callable, Dict, List, Optional
+
+
+class BaseProvider:
+    def __init__(self, model: str, dims: int, api_config: Optional[Dict] = None):
+        self._dims = dims
+        self._model = model
+
+    @property
+    def model(self) -> str:
+        return self._model
+
+    @property
+    def dims(self) -> int:
+        return self._dims
+
+    def set_model(self, model, dims: int = None):
+        self._model = model
+        if dims:
+            self._dims = dims
+
+    def embed_many(
+        self, inputs: List[str], preprocess: callable = None, chunk_size: int = 1000
+    ) -> List[float]:
+        raise NotImplementedError
+
+    def embed(self, emb_input: str, preprocess: callable = None) -> List[float]:
+        raise NotImplementedError
+
+    async def aembed_many(
+        self, inputs: List[str], preprocess: callable = None, chunk_size: int = 1000
+    ) -> List[float]:
+        raise NotImplementedError
+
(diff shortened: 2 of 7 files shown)


redisvl/providers/__init__.py
```diff
@@ -0,0 +1,9 @@
+
+from redisvl.providers.openai import OpenAIProvider
```
tylerhutcherson, line 2: should we consider organizing it one layer deeper, under `embeddings`? that way its clear whether its for Embedding creation or something else (future proofing)

redisvl/providers/base.py
```diff
@@ -0,0 +1,43 @@
...
+
+    async def aembed_many(
+        self, inputs: List[str], preprocess: callable = None, chunk_size: int = 1000
+    ) -> List[float]:
+        raise NotImplementedError
+
+    async def aembed(self, emb_input: str, preprocess: callable = None) -> List[float]:
+        raise NotImplementedError
+
+    def batchify(self, seq: list, size: int, preprocess: Callable = None):
```
tylerhutcherson, line 38: love it

redisvl/providers/huggingface.py
```diff
@@ -0,0 +1,29 @@
+from typing import Dict, List, Optional
+
+from sentence_transformers import SentenceTransformer
```
tylerhutcherson, line 3: we could consider importing this scoped within the class -- and then we can prevent folks from having to install SentenceTransformers (torch, etc) if they don't plan to need it?

redisvl/providers/openai.py
```diff
@@ -0,0 +1,48 @@
+from typing import Dict, List, Optional
+
+import openai
```
tylerhutcherson, line 3: same point as above?

redisvl/providers/openai.py
```diff
@@ -0,0 +1,48 @@
...
+    def __init__(self, model: str, api_config: Optional[Dict] = None):
+        dims = 1536
+        super().__init__(model, dims, api_config)
+        if not api_config:
+            raise ValueError("OpenAI API key is required in api_config")
+
+        openai.api_key = api_config.get("api_key", None)
+        self._model_client = openai.Embedding
+
+    def embed_many(
```
tylerhutcherson, line 18: should we look to make this immune to rate limit issues? One pattern I have seen is to use `tenacity` and their exponential backoff decorator.

tylerhutcherson (approved):
Left some suggestions but looks great all around!

redisvl/providers/huggingface.py
Spartee: It's the default for LLMCache... should we not assume that? I do like the scoped gaurd approach.

redisvl/providers/openai.py
Spartee: @tylerhutcherson can you point me to or write a code sample for this?

redisvl/providers/__init__.py
Spartee: That's a long import but I see what you mean. A couple options

```python
from redisvl.api.embedding.openai
from redisvl.embedding.openai import OpenAIEmbedding
from redisvl.model.embedding.openai import ...

```

others?

Spartee:
addressing changes in Next PR

PR #17, opened 2023-07-11
LLMCache

Spartee:
Introduce ``LLMCache`` library api. This allows users to use a semantic cache in their applications through utilizing Redisvl.

example:

```python

from redisvl.llmcache.semantic import SemanticCache
cache = SemanticCache(redis_host="localhost", redis_port=6379, redis_password=None)

def answer_question(question: str):
    results = cache.check(question)
    if results:
        return results[0]
    else:
        answer = ask_gpt3(question)
        cache.store(question, answer)
        return answer

```

### DIFF
--- a/redisvl/llmcache/base.py
+++ b/redisvl/llmcache/base.py
@@ -1,0 +1,45 @@
+import hashlib
+import logging
+from typing import Callable, Dict, Optional
+
+
+class BaseLLMCache:
+
+    verbose: bool = True
+
+    def check(self, prompt: str) -> Optional[Dict[str, str]]:
+        raise NotImplementedError
+
+    def store(
+        self,
+        prompt: str,
+        response: str,
+        metadata: Optional[dict] = {},
+        key: Optional[str] = None,
+    ):
+        """Stores the specified key-value pair in the cache along with metadata."""
+        raise NotImplementedError
+
+    def _refresh_ttl(self, key: str):
+        """Refreshes the TTL for the specified key."""
+        raise NotImplementedError
+
+    def hash_input(self, prompt: str):
+        """Hashes the input using SHA256."""
+        return hashlib.sha256(prompt.encode("utf-8")).hexdigest()
+
+    def cache_response(self, llm_callable: Callable):
+        """Decorator method for wrapping custom callables"""
+
+        def wrapper(*args, **kwargs):
+            # Check LLM Cache first
+            key = self.hash_input(*args, **kwargs)
+            response = self.check(*args, **kwargs)
+            if response:
+                self._refresh_ttl(key)
+                return response
+            # Otherwise execute the llm callable here
+            response = llm_callable(*args, **kwargs)
+            return response
+
+        return wrapper
--- a/redisvl/llmcache/semantic.py
+++ b/redisvl/llmcache/semantic.py
@@ -1,0 +1,130 @@
+import logging
+import math
+import os
+from typing import Any, Callable, Dict, List, Optional
+
+from redis.commands.search.field import VectorField
+
+from redisvl.index import SearchIndex
+from redisvl.llmcache.base import BaseLLMCache
+from redisvl.providers import HuggingfaceProvider
+from redisvl.providers.base import BaseProvider
+from redisvl.query import create_vector_query
+from redisvl.utils.log import get_logger
+from redisvl.utils.utils import array_to_buffer
+
+_logger = get_logger(__name__)
+
+
+class SemanticCache(BaseLLMCache):
+    """Cache for Large Language Models."""
+
+    _default_fields = [
+        VectorField(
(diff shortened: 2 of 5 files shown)


redisvl/llmcache/base.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def _refresh_ttl(self, key: str):
+        """Refreshes the TTL for the specified key."""
+        raise NotImplementedError
+
+    def hash_input(self, prompt: str):
+        """Hashes the input using SHA256."""
+        return hashlib.sha256(prompt.encode("utf-8")).hexdigest()
+
+    def cache_response(self, llm_callable: Callable):
```
tylerhutcherson, line 30: glad this is still in here!

redisvl/llmcache/semantic.py
```diff
@@ -0,0 +1,126 @@
...
+    """Cache for Large Language Models."""
+
+    _default_fields = [
+        VectorField(
+            "prompt_vector",
+            "FLAT",
+            {"DIM": 768, "TYPE": "FLOAT32", "DISTANCE_METRIC": "COSINE"},
+        ),
+    ]
+    _default_provider = HuggingfaceProvider("sentence-transformers/all-mpnet-base-v2")
```
tylerhutcherson, line 26: still think we should consider `embedding_provider` or something so it's clear what its providing

redisvl/llmcache/semantic.py
```diff
@@ -0,0 +1,126 @@
...
+            prompt_vector = array_to_buffer(vector)
+        else:
+            prompt_vector = array_to_buffer(self._provider.embed(prompt)) # type: ignore
+        results = self._index.search(query, query_params={"vector": prompt_vector})
+
+        cache_hits = []
+        for doc in results.docs:
+            sim = similarity(doc.vector_score)
+            _logger.info("Similarity: %s", sim)
+            if similarity(doc.vector_score) > self.threshold:
```
tylerhutcherson: we should take advantage of a range query for this. more performant and built into the search API. I think all we need is to expose a different `create_range_query` method in addition to `create_knn_query`* maybe?

redisvl/llmcache/semantic.py
```diff
@@ -0,0 +1,126 @@
...
+    def _refresh_ttl(self, key: str):
+        """Refreshes the TTL for the specified key."""
+        client = self._index.get_client()
+        if client:
+            client.expire(key, self.ttl)
+        else:
+            raise RuntimeError("LLMCache is not connected to a Redis instance.")
+
+
+def similarity(distance):
```
tylerhutcherson: add types?

redisvl/providers/huggingface.py
```diff
@@ -10,16 +9,23 @@ def __init__(self, model: str, api_config: Optional[Dict] = None):
...
-    def embed(self, emb_input: str, preprocess: callable = None) -> List[float]:
+    def embed(self, emb_input: str, preprocess: Optional[Callable] = None) -> List[float]:
         if preprocess:
             emb_input = preprocess(emb_input)
         embedding = self._model_client.encode([emb_input])[0]
         return embedding.tolist()
 
     def embed_many(
-        self, inputs: List[str], preprocess: callable = None, chunk_size: int = 1000
+        self, inputs: List[str], preprocess: Optional[Callable] = None, chunk_size: int = 1000
```
tylerhutcherson: this notebook from google shows how to use tenacity for retry and exponential backoff to deal with rate limits: https://github.com/GoogleCloudPlatform/generative-ai/blob/8e19bc6a5de9f253444a39f82790046b1817f89d/language/examples/document-qa/question_answering_large_documents.ipynb

tylerhutcherson (approved):
nice! left a few comments

redisvl/providers/huggingface.py
Spartee: Going to make this a followup issue

redisvl/llmcache/semantic.py
Spartee: Let's chat about this. I agree, just want to make sure we are thinking of the same thing. This would be a good one for you to pick up on.

redisvl/llmcache/semantic.py
Spartee: Agreed. I'm thinking

```python
from redisvl.provider.embedding import HuggingfaceProvider
from redisvl.provider.embedding import OpenaiProvider
```

redisvl/llmcache/semantic.py
Spartee: done

redisvl/llmcache/semantic.py
Spartee: There are some tradeoffs to dicuss here though...I'm going to leave this for now.

PR #22, opened 2023-07-25
Update query interface

Spartee:
Introduce new Query design with Filters and notebooks to show examples

includes
- documenation updates
- examples
- semantic cache and SearchIndex updates
- tests


redisvl/llmcache/base.py
```diff
@@ -26,22 +26,3 @@ def _refresh_ttl(self, key: str):
     def hash_input(self, prompt: str):
         """Hashes the input using SHA256."""
         return hashlib.sha256(prompt.encode("utf-8")).hexdigest()
-
```
tylerhutcherson, line 29: don't want to kep the decorator option?

redisvl/llmcache/semantic.py
```diff
@@ -6,56 +6,94 @@
...
 class SemanticCache(BaseLLMCache):
     """Cache for Large Language Models."""
 
+    # TODO allow for user to change default fields
     _default_fields = [
         VectorField(
             "prompt_vector",
             "FLAT",
             {"DIM": 768, "TYPE": "FLOAT32", "DISTANCE_METRIC": "COSINE"},
         ),
```
tylerhutcherson, line 22: right now it looks like the default fields is "fixed" but you can alter the embeddings provider. would this cause a potential clash with the schema (embedding dimensions etc)?

redisvl/llmcache/semantic.py
```diff
@@ -74,25 +121,40 @@ def check(
...
+        """
         if not prompt and not vector:
             raise ValueError("Either prompt or vector must be specified.")
 
-        query = create_vector_query(
-            return_fields=fields,
+        if not vector:
+            vector = self._provider.embed(prompt)  # type: ignore
+
+        v = VectorQuery(
```
tylerhutcherson, line 144: I'd suggest we use a range query for this use case - less client side processing. If you agree -- we will want to flip the polarity of the threshold such that smaller is more "strict"

redisvl/llmcache/semantic.py
```diff
@@ -74,25 +121,40 @@ def check(
...
-        results = self._index.search(query, query_params={"vector": prompt_vector})
+
+        results = self._index.search(v.query, query_params=v.params)
 
         cache_hits = []
         for doc in results.docs:
             self._refresh_ttl(doc.id)
-            sim = similarity(doc.vector_score)
+            sim = similarity(doc.vector_distance)
             if sim > self.threshold:
```
tylerhutcherson, line 158: this is the part we could do without if we use range query

redisvl/query.py
```diff
@@ -1,48 +1,209 @@
...
-    """
-    base_query = f"{tags}=>[{search_type} {number_of_results} @{vector_field_name} ${vector_param_name} AS vector_score]"
-    if return_score:
-        return_fields.append("vector_score")
-    query = Query(base_query).return_fields(*return_fields).dialect(2)
-    if sort:
-        query.sort_by("vector_score")
-    return query
+
+class Filter:
```
tylerhutcherson, line 9: might suggest keeping a separate filter module

redisvl/llmcache/semantic.py
tylerhutcherson: Could also be added later

tests/integration/test_query.py
```diff
@@ -0,0 +1,220 @@
...
+    for doc in results.docs:
+        assert (doc.credit_score == "high") or (18 <= int(doc.age) <= 100)
+    assert len(results.docs) == 5
+
+    # (@credit_score:{high} ~@age:[18 100])=>[KNN 10 @user_embedding $vector AS vector_distance]
+    t = TagFilter("credit_score", "high")
+    n = NumericFilter("age", 18, 100)
+    t ^= n
+    v.set_filter(t)
+
```
tylerhutcherson, line 216: only other input on tests would be to test some more tricky character scenarios where we'd have to properly escape them to make it work

tylerhutcherson (approved):
wooo!

redisvl/llmcache/base.py
Spartee: I'll explain, needs more work on the edge cases. I found too many ways to break it for me to be comfortable for a release with it.

redisvl/llmcache/semantic.py
Spartee: There is a TODO in there about this. Struggling with the intersection between hand-holding and custom capability.

redisvl/llmcache/semantic.py
Spartee: premature Optimization is the root of all evil. Next release.

redisvl/llmcache/semantic.py
Spartee: see previous comment

redisvl/query.py
Spartee: Thought about this, but then everyone has to do

```python
from redisvl.filter import TagFilter
from redisvl.query import VectorQuery
```

instead of 

```python
from redisvl.query import TagFilter, VectorQuery
```

It also made type hints easier... but I hear you.. what do you think now?

tests/integration/test_query.py
Spartee: Good call. Examples? Feel like you know more about this than I do.

redisvl/query.py
Spartee: Actually, the first looks cleaner..I'll do this

PR #28, opened 2023-07-31
Llmcache ttl configuration

tylerhutcherson:
Enables TTL for loading data to Redis using RedisVL. This TTL is an optional kwarg for the underlying `load` function. This is particularly useful for clients and abstractions like SemanticCache that lean on Redis support for ephemeral data.

Also adds a `clear` method to the LLM cache class that invalidates all data from the cache, but does not disturb the index itself.

### DIFF
--- a/redisvl/index.py
+++ b/redisvl/index.py
@@ -254,16 +254,18 @@ class SearchIndex(SearchIndexBase):
                 containing the data to be indexed
         raises:
             redis.exceptions.ResponseError: If the index does not exist
+        # TODO -- should we return a count of the upserts? or some kind of metadata?
         """
-        if not data:
-            return
-        if not isinstance(data, Iterable):
-            if not isinstance(data[0], dict):
-                raise TypeError("data must be an iterable of dictionaries")
+        if data:
+            if not isinstance(data, Iterable):
+                if not isinstance(data[0], dict):
+                    raise TypeError("data must be an iterable of dictionaries")
 
-        for record in data:
-            key = f"{self._prefix}:{self._get_key_field(record)}"
-            self._redis_conn.hset(key, mapping=record)  # type: ignore
+            pipe = self._redis_conn.pipeline(transaction=False)
+            for record in data:
+                key = f"{self._prefix}:{self._get_key_field(record)}"
+                pipe.hset(key, mapping=record)  # type: ignore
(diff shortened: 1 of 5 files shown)


redisvl/llmcache/semantic.py
```diff
@@ -114,6 +114,10 @@ def set_threshold(self, threshold: float):
             raise ValueError("Threshold must be between 0 and 1.")
         self._threshold = float(threshold)
 
+    def clear(self):
+        """Clear the LLMCache and create a new underlying index."""
+        self._index.delete(drop=True)
```
tylerhutcherson: So this is one area to think more about... Ideally I would like to allow clearing out the data itself, but not the index. The manual way is to search for all data in the index, get list of keys, and delete them. But this is messy.

Alternatively, we could use the delete method here along with a construction of a new index. But, this gave me some issues with creating an index again with the right params/attributes. Maybe you can shed light here.

Spartee: My guess would be a ``scan_iter`` across the DB using fuzzy match and then delete from index. This is similar to the feast update approach. Don't love it, but nothing else is coming to mind currently.

tylerhutcherson: Got it - yeah it won't be too bad. Will implement now and push commit for review along with a test

redisvl/index.py
```diff
@@ -254,16 +254,23 @@ def load(self, data: Iterable[Dict[str, Any]], **kwargs):
...
-        if not data:
-            return
-        if not isinstance(data, Iterable):
-            if not isinstance(data[0], dict):
-                raise TypeError("data must be an iterable of dictionaries")
 
-        for record in data:
-            key = f"{self._prefix}:{self._get_key_field(record)}"
-            self._redis_conn.hset(key, mapping=record)  # type: ignore
+        # TODO -- should we return a count of the upserts? or some kind of metadata?
```
Spartee: Todo outside comments. these become docs

Spartee (dismissed):
Minor comments. Looks great.

redisvl/index.py
tylerhutcherson: fixed!

Spartee (approved):
LGTM

PR #33, opened 2023-08-02
add support for cache construction from an index

tylerhutcherson:
PR introduces the ability to load an LLMCache from an underlying index. Also makes some adjustments to the index class to support this.

- Add support for `SemanticCache.from_index(...)`
- Update implementation of the `SearchIndex`:
    - move `key_field` as a `load` method arg
    - add `_get_key` method to create the redis key for an object and associated test
- Update `SearchIndex.from_existing(...)` to accept a `redis_url` and `connection_args`

@Spartee this PR has some crucial decisions in it regarding user experience. I left a few TODOs and think we should make sure we are comfortable with the risks of `from_index`. What are the main use cases?


Spartee (dismissed):
Approved pending lint, mypy, and stlying.

```bash
make style
make lint
make mypy
```

tylerhutcherson:
Made storage_type optional and performed tests, linting, styling and mypy updates.

PR #34, opened 2023-08-04
Implement Geographic Filter

Spartee:
Implement Geographic filter and clean up the hybrid query notebook.


tylerhutcherson (approved):
This is awesome. The hybrid search example is incredible and very detailed at the right level of complexity for the target user. Amazing.

PR #35, opened 2023-08-04
Add optional vectorizer bytes conversion and exponential backoff

tylerhutcherson:
This PR adds support for:
- Exponential backoff for the openAI vectorizer
- Optional `as_buffer` flag that can be passed to the various vectorizer embedding creation methods. Defaults to `False` always and assumes the user wants it back as a list of floats. But now this option is enabled.


Spartee (approved):
Approved once merge conflicts fixed.

PR #43, opened 2023-08-10
add support for processing results on search and query

tylerhutcherson:
This PR introduces post processing of query/search results from Redis.

- Default processor used for `query` method and left out for `search` method.
- Ability to pass in a `Callable`, using your own processor after results are pulled. Must work on a `List[Result]` object and return a `List[Any]]` object.
- Adds tests.

### DIFF
--- a/redisvl/index.py
+++ b/redisvl/index.py
@@ -5,6 +5,7 @@ from uuid import uuid4
 if TYPE_CHECKING:
     from redis.commands.search.field import Field
     from redis.commands.search.result import Result
+    from redis.commands.search.document import Document
 
 import redis
 from redis.commands.search.indexDefinition import IndexDefinition, IndexType
@@ -22,7 +23,12 @@ from redisvl.utils.utils import check_redis_modules_exist, convert_bytes, make_d
 def _default_process_results(results: List["Result"]) -> List[Dict[str, Any]]:
     # Convert a list of Result objects into a list of document dicts
     # Process results functions must accept a list of results and return a list
-    return [doc.__dict__ for doc in results.docs]
+    def _process(doc: "Document") -> dict:
+        d = doc.__dict__
+        if "payload" in d:
+            del d["payload"]
+        return d
+    return [_process(doc) for doc in results.docs]
 
 
 class SearchIndexBase:
@@ -374,7 +380,7 @@ class SearchIndex(SearchIndexBase):
 
             # Check if outer interface passes in TTL on load
             ttl = kwargs.get("ttl")
-            with self._redis_conn.pipeline() as pipe:
+            with self._redis_conn.pipeline(transaction=False) as pipe:
                 for record in data:
                     key = self._get_key(record, key_field)
                     pipe.hset(key, mapping=record)  # type: ignore
--- a/redisvl/index.py
+++ b/redisvl/index.py
@@ -1,5 +1,5 @@
 import asyncio
-from typing import TYPE_CHECKING, Any, Dict, Iterable, List, Optional, Callable
+from typing import TYPE_CHECKING, Any, Callable, Dict, Iterable, List, Optional
 from uuid import uuid4
 
 if TYPE_CHECKING:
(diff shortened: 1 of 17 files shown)


redisvl/index.py
```diff
@@ -19,6 +19,12 @@
 from redisvl.utils.utils import check_redis_modules_exist, convert_bytes, make_dict
 
 
+def _default_process_results(results: List["Result"]) -> List[Dict[str, Any]]:
```
Spartee: Couple thoughts

1. I think  a ``processor`` class with a ``__call__`` method would be clearer here. and shorter ha
2. should we take out some of the fields like ``payload`` by default? the ``__dict__`` method here usually returns alot of cruft.

redisvl/index.py
```diff
@@ -351,13 +374,13 @@ def load(
...
             # Check if outer interface passes in TTL on load
             ttl = kwargs.get("ttl")
-            pipe = self._redis_conn.pipeline(transaction=False)
-            for record in data:
-                key = self._get_key(record, key_field)
-                pipe.hset(key, mapping=record)  # type: ignore
-                if ttl:
-                    pipe.expire(key, ttl)
-            pipe.execute()
+            with self._redis_conn.pipeline() as pipe:
```
Spartee: transaction not nessesary here right?

redisvl/llmcache/semantic.py
```diff
@@ -180,14 +184,18 @@ def check(
...
+    def _process_cache_hits(self, results: List["Result"]) -> List[str]:
+        cache_hits: List[str] = []
         for doc in results.docs:
             sim = similarity(doc.vector_distance)
             if sim > self.threshold:
                 self._refresh_ttl(doc.id)
-                cache_hits.append(doc.response)
+                cache_hits.append(
+                    doc.response
+                )  # TODO what do we actually want to return here?
```
Spartee: Allow the user to specify fields to return in ``check``? defaults to ``response`` only. 

maybe another method like ``check_and_get`` that returns a list of dicts?

Spartee (commented):
Great start. Minor comments.

redisvl/index.py
```diff
@@ -57,32 +58,33 @@ def search(self, *args, **kwargs) -> List["Result"]:
         Returns:
             List[Result]: A list of search results
         """
-        return self._redis_conn.ft(self._name).search(*args, **kwargs)  # type: ignore
+        results: List["Result"] = self._redis_conn.ft(self._name).search(
+            *args, **kwargs
+        )
+        return results
```
Spartee, line 68: We should make a note about performance being better here since the results are not post processed.

redisvl/index.py
```diff
@@ -164,6 +166,18 @@ def _get_key(self, record: Dict[str, Any], key_field: str = None) -> str:
                 raise ValueError(f"Key field {key_field} not found in record {record}")
         return f"{self._prefix}:{key}"
 
+    @staticmethod
+    def _process_results(results: List["Result"]) -> List[Dict[str, Any]]:
```
Spartee: I feel like this might be useful to some users. Let's make it a utility?

Spartee (commented):
Not sure how I comment on notebooks, but a few things

1. Let's strip out the id from the tables in the notebook? Looks better.
2. Seems like the LLMCache had a pretty significant performance degradation.. order of magnitude? 
3. I'm about to change the hybrid queries one quite a bit, so maybe just take it out and ill adjust?

Looks great.

redisvl/index.py
tylerhutcherson: yeah good call

tylerhutcherson:
> Not sure how I comment on notebooks, but a few things
> 
> 1. Let's strip out the id from the tables in the notebook? Looks better.
> 2. Seems like the LLMCache had a pretty significant performance degradation.. order of magnitude?
> 3. I'm about to change the hybrid queries one quite a bit, so maybe just take it out and ill adjust?
> 
> Looks great.

I think the notebooks definitely need a good clean up. Let me merge this in as is and then when you update for filtering, we can get those updates in (otherwise we will be dealing with nastier merge conflicts).

PR #51, opened 2023-08-16
Add FilterQuery

Spartee:
Add a ``FilterQuery`` class for running just a query with a filter and no vector search in the case where the user needs to do this and should not have to also use Redis-py

### Example

```python
>>> from redisvl.query import FilterQuery
>>> from redisvl.query.filter import Tag
>>> t = Tag("brand") == "Nike"
>>> q = FilterQuery(return_fields=["brand", "price"], filter_expression=t)
```


redisvl/cli/utils.py
```diff
@@ -18,20 +18,15 @@ def create_redis_url(args: Namespace) -> str:
...
     parser.add_argument("-p", "--port", help="Redis port", type=int, default=6379)
-    parser.add_argument(
-        "--user", help="Redis username", type=str, default="default"
-    )
+    parser.add_argument("--user", help="Redis username", type=str, default="default")
     parser.add_argument("--ssl", help="Use SSL", action="store_true")
-    parser.add_argument(
-        "-a", "--password", help="Redis password", type=str, default=""
-    )
+    parser.add_argument("-a", "--password", help="Redis password", type=str, default="")
```
tylerhutcherson, line 31: we will eventually need to add TLS support here in the form of a path to the server CA, and client/user certs (public and private) too. With the bedrock launch coming, it would be great to push users here for an easier onboarding :)

tylerhutcherson (approved):
LGTM!

PR #52, opened 2023-08-23
Add VertexAI Text vectorizer

tylerhutcherson:
Need to add gcp credentials and env vars to our github actions secret

### DIFF
--- a/.github/workflows/run_tests.yml
+++ b/.github/workflows/run_tests.yml
@@ -35,14 +35,16 @@ jobs:
       run: |
         REDIS_URL=redis://localhost:6379
         echo REDIS_URL=$REDIS_URL >> $GITHUB_ENV
+    - name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1'
+      with:
+        credentials_json: ${{ secrets.GOOGLE_CREDENTIALS }}
     - name: Run tests
       env:
         OPENAI_API_KEY: ${{ secrets.OPENAI_KEY }}
         GCP_LOCATION: ${{ secrets.GCP_LOCATION }}
         GCP_PROJECT_ID: ${{ secrets.GCP_PROJECT_ID }}
-        GOOGLE_APPLICATION_CREDENTIALS: gcp-credentials.json
       run: |
-        echo "$GCP_SERVICE_ACCOUNT_API_KEY_BASE64" | base64 --decode | jq > gcp-credentials.json
         make test-cov
     - name: Run notebooks
       env:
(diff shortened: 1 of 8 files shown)


redisvl/vectorize/text/vertexai.py
```diff
@@ -0,0 +1,142 @@
...
+        self._model_client = TextEmbeddingModel.from_pretrained(model)
+        self._dims = self._set_model_dims()
+
+    def _set_model_dims(self) -> int:
+        try:
+            embedding = self._model_client.get_embeddings(["dimension test"])[0].values
+        except (KeyError, IndexError) as ke:
+            raise ValueError(f"Unexpected response from the VertexAI API: {str(ke)}")
+        # TODO - except openai.error.AuthenticationError as ae:
+        #     raise ValueError(f"Error authenticating with the OpenAI API: {str(ae)}")
```
Spartee: this doesn't seem to be relevant

redisvl/vectorize/text/vertexai.py
```diff
@@ -0,0 +1,142 @@
+from typing import Callable, Dict, List, Optional
+
+from tenacity import retry, stop_after_attempt, wait_random_exponential
+from tenacity.retry import retry_if_not_exception_type
+
+from redisvl.vectorize.base import BaseVectorizer
+
+
+class VertexAITextVectorizer(BaseVectorizer):
```
Spartee, line 9: annoying, I know, but I like ``Vertex`` or ``Google``. Just seems cleaner

.github/workflows/run_tests.yml
```diff
@@ -35,9 +35,15 @@ jobs:
       run: |
         REDIS_URL=redis://localhost:6379
         echo REDIS_URL=$REDIS_URL >> $GITHUB_ENV
+    - name: 'Authenticate to Google Cloud'
+      uses: 'google-github-actions/auth@v1'
+      with:
+        credentials_json: ${{ secrets.GOOGLE_CREDENTIALS }}
```
Spartee, line 41: I think this needs to be read in first. I've never done this before, so a couple things I found

- https://cloud.google.com/blog/products/identity-security/enabling-keyless-authentication-from-github-actions
- https://github.com/VeraZab/nyc-stats/blob/main/.github/workflows/deploy-flow.yaml

Spartee (commented):
We can bash heads together on this tomorrow.

redisvl/vectorize/text/vertexai.py
tylerhutcherson: Hmmm I would disagree -- "Vertex" isn't a thing (see naming convention here: https://python.langchain.com/docs/integrations/llms/google_vertex_ai_palm) and "Google" is too vague. Just like OpenAI, I think we need to call it by it's name, even if verbose.

.github/workflows/run_tests.yml
tylerhutcherson: I tried this approach because the other method with base64 encoding wasn't working.

redisvl/vectorize/text/vertexai.py
tylerhutcherson: will clean up

PR #58, opened 2023-08-30
Adding the release drafter

chayim:
The release drafter is used across Redis clients to provide a baseline for valuable release notes, across clients. The goal is not to replace the need to write release notes, but rather to provide a *consistent* basis, and baseline for those.

As part of this, pull request **titles** upon merge to **master** become the comment in a release note. A merged pull request needs to have one of the labels below, thus inserting itself into the right place within the proposed, future release notes.  In no way, does this negate a pull request appearing in multiple locations, nor enforce label requirements.  The labels are as follows:

bug - A pull request with this label will appear within the bugs section.
feature - This label ensures the pull request is marked as a feature
maintenance - Documentation changes, CI changes, and basically ongoing library maintenance work, belong here.
breakingchange - Occasionally, we need to communicate with users, that a change breaks behaviour, or an API. By tagging accordingly, this appears at the top of the release notes.

The action itself, is stored in the workflows folder, whereas the configuration (and hence template) for the github action appears in .github. The proposed configuration is a direct copy from redis-py, as these are used everywhere. Finally, as an example of well generated release notes, for larger changes please see [release 4.5.5](https://github.com/redis/redis-py/releases/tag/v4.5.5)

### DIFF
--- a/.github/release-drafter-config.yml
+++ b/.github/release-drafter-config.yml
@@ -1,0 +1,48 @@
+name-template: '$NEXT_MINOR_VERSION'
+tag-template: 'v$NEXT_MINOR_VERSION'
+autolabeler:
+  - label: 'maintenance'
+    files:
+      - '*.md'
+      - '.github/*'
+  - label: 'bug'
+    branch:
+      - '/bug-.+'
+  - label: 'maintenance'
+    branch:
+      - '/maintenance-.+'
+  - label: 'feature'
+    branch:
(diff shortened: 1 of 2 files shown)


chayim:
Note @Spartee @tylerhutcherson in the ideal world, this pull request, would be labelled _maintenance_

tylerhutcherson (approved):
Works for me!

PR #62, opened 2023-09-05
CI platform support

chayim:
This pull requests extends testing to multiple versions of redis-stack. Along the way, it attempts to track all versions of python tested by redis-py, to ensure coverage. Finally, adding tests for hiredis, and updating the README accordingly.


Spartee:
Let's run this on a Cron-like nightly basis. Keep the PR CI to the latest released version of redis-stack. Eventually, we'll have a nightly release anyway.

This is a ton of CI jobs for a single PR. I'm a little worried also about testing the vectorizers this many times. eventually it'll start racking up some costs. 

@tylerhutcherson thoughts here?

tylerhutcherson:
> This is awesome.
> 
> Let's run this on a Cron-like nightly basis. Keep the PR CI to the latest released version of redis-stack. Eventually, we'll have a nightly release anyway.
> 
> This is a ton of CI jobs for a single PR. I'm a little worried also about testing the vectorizers this many times. eventually it'll start racking up some costs.
> 
> @tylerhutcherson thoughts here?

We will be fine for the basic unit and integration tests -- but I would be worried about the notebook end-to-end tests (unless those are not impacted here)?

chayim:
I generally thing CI should run against everything always. I want to know that I broke things. However, we can pick up from what Dvir + I did in vecsim and run _increased platforms_ (whatever we determine that means) based on a label. WDYT @Spartee ?

Spartee:
Well there are a couple issues

1. Seems like we have an actual dependency issue with sentence transformers in python 3.8. So we may need to address that at some point.
2. We'll run into rate limits for the vectorizers very quickly. one route around this would be to test the vectorizer tests only on one redis variant since it really doesn't have to do with Redis.

Overall though I'd say it's not entirely necessary to run that often since we never issue a command that isn't covered in Redis-py (by design). However, I think having the ability to, or having it run every so often would be nice.

Either way, something else will have to be done to either
1. modify the test and test suites to allow the vectorizer tests to run separately and in a new workflow (and possibly with rate-limit control)
2. or
3. adapt to run a new version of the test suite across all redis variant's on a semi-regular basis.

Let me know what you think @chayim

chayim:
> Seems like we have an actual dependency issue with sentence transformers in python 3.8. So we may need to address that at some point.

Which is?

> We'll run into rate limits for the vectorizers very quickly. one route around this would be to test the vectorizer tests only on one redis variant since it really doesn't have to do with Redis.

We could do a labelled PR approach. Basically run specific tests (only) when a pull request label is applied. We'd tag these tests in the codebase with `@pytest.mark.ondemand` or something, in order to make it explicit. 

> Overall though I'd say it's not entirely necessary to run that often since we never issue a command that isn't covered in Redis-py (by design). However, I think having the ability to, or having it run every so often would be nice.

What about a similar approach to libraries. There we run against edge of redis-stack-server, nightly. We could do something similar (twice weekly until we learn more?) that makes sense for this cadence.

> Either way, something else will have to be done to either
> ...

As above,  IMHO. WDYT @Spartee

chayim:
bump @Spartee @tylerhutcherson feedback?

IMHO any tests we want to be here and there, let's use those pytest markers. WDYT?

Spartee:
Merged in #73

PR #63, opened 2023-09-07
make prefix optional and update vector field args

tylerhutcherson:
This PR makes the `prefix` field optional and an empty string by default, which mirrors the default because of the RediSearch API. It also removes the vector field args for block size and initial cap as these are not user friendly fields.


Spartee (approved):
LGTM

PR #65, opened 2023-10-13
Add support for CountQuery

tylerhutcherson:
Adds support for a simple `CountQuery` which expects a `FilterExpression` and allows users to check how many records match a particular set of filters. Also adds documentation and examples for multiple tag fields and count queries.

### DIFF
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -41,6 +41,8 @@ class CountQuery(BaseQuery):
             >>> from redisvl.query.filter import Tag
             >>> t = Tag("brand") == "Nike"
             >>> q = CountQuery(filter_expression=t)
+            >>> results = index.search(q.query)
+            >>> count = results.total
         """
         self.set_filter(filter_expression)
         self._params = params
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -21,6 +21,84 @@ class BaseQuery:
         raise NotImplementedError
 
 
+class CountQuery(BaseQuery):
+    def __init__(
+        self,
+        filter_expression: FilterExpression,
+        params: Optional[Dict[str, Any]] = None,
+    ):
+        """Query for a simple count operation on a filter expression.
+
+        Args:
+            filter_expression (FilterExpression): The filter expression to query for.
+            params (Optional[Dict[str, Any]], optional): The parameters for the query. Defaults to None.
+
+        Raises:
+            TypeError: If filter_expression is not of type redisvl.query.FilterExpression
+
+        Examples:
+            >>> from redisvl.query import CountQuery
+            >>> from redisvl.query.filter import Tag
+            >>> t = Tag("brand") == "Nike"
+            >>> q = CountQuery(filter_expression=t)
+        """
+        self.set_filter(filter_expression)
+        self._params = params
+
+    def __str__(self) -> str:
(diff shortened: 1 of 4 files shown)


redisvl/query/query.py
```diff
@@ -21,6 +21,86 @@ def params(self) -> Dict[str, Any]:
...
+            >>> results = index.search(q.query)
+            >>> count = results.total
+        """
+        self.set_filter(filter_expression)
+        self._params = params
+
+    def __str__(self) -> str:
+        return " ".join([str(x) for x in self.query.get_args()])
+
+    def set_filter(self, filter_expression: FilterExpression):
```
Spartee: Can we abstract these into ``BaseQuery``?

redisvl/query/query.py
```diff
@@ -21,6 +21,86 @@ def params(self) -> Dict[str, Any]:
         raise NotImplementedError
 
 
+class CountQuery(BaseQuery):
```
Spartee, line 50: Should this also have return fields? I guess you could just do a FilterQuery and use ``len()``

redisvl/query/query.py
```diff
@@ -21,6 +21,86 @@ def params(self) -> Dict[str, Any]:
...
+
+        Raises:
+            TypeError: If filter_expression is not of type redisvl.query.FilterExpression
+
+        Examples:
+            >>> from redisvl.query import CountQuery
+            >>> from redisvl.query.filter import Tag
+            >>> t = Tag("brand") == "Nike"
+            >>> q = CountQuery(filter_expression=t)
+            >>> results = index.search(q.query)
```
Spartee: lets show using ``query()``

Spartee (approved):
Minor feedback. I like CountQuery. Introducing more of these types of abstractions is the idea. Some pieces to fix with the linting/CI but once those are figured out, feel free to merge.

redisvl/query/query.py
tylerhutcherson: ohhh probably let me check

redisvl/query/query.py
tylerhutcherson: The idea of count query is that it returns `no_content()` thus is a more efficient query than doing a filter query with len

redisvl/query/query.py
tylerhutcherson: so this will take some more refactor as the `query` method currently processes a list of result objects from redisearch. we need to just return an int. I will probably an approach here for your review

redisvl/query/query.py
tylerhutcherson: indeed -- fixed!

redisvl/query/query.py
tylerhutcherson: this is how I figured we can handle it.. specifically on the `query()` method ONLY:

```python
    async def query(self, query: "BaseQuery") -> List[Dict[str, Any]]:
        """Run a query on this index.

        This is similar to the search method, but takes a BaseQuery
        object directly (does not allow for the usage of a raw
        redis query string) and post-processes results of the search.

        Args:
            query (BaseQuery): The query to run.

        Returns:
            List[Result]: A list of search results.
        """
        results = await self.search(query.query, query_params=query.params)
        if isinstance(query, CountQuery):
            return results.total
        return process_results(results)
```

PR #66, opened 2023-10-16
Update key prefix and access method

tylerhutcherson:
After working through arxiv example, this is a small adjustment to help with usability:
- adds an explicit `key` method help function that combines the index key prefix + in addition to the records identifier to create the key name
- updates the other private method from `_get_key` to `_create_key` to be more specific

### DIFF
--- a/redisvl/index.py
+++ b/redisvl/index.py
@@ -29,12 +29,12 @@ class SearchIndexBase:
     def __init__(
         self,
         name: str,
-        prefix: str = "rvl",
+        key_prefix: str = "rvl",
         storage_type: str = "hash",
         fields: Optional[List["Field"]] = None,
     ):
         self._name = name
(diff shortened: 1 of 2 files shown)


redisvl/index.py
```diff
@@ -151,7 +151,12 @@ def disconnect(self):
         """Disconnect from the Redis instance"""
         self._redis_conn = None
 
-    def _get_key(self, record: Dict[str, Any], key_field: Optional[str] = None) -> str:
+    def key(self, key_value: str) -> str:
```
Spartee, line 154: Docstring and this is good.

Spartee (approved):
Approved with addition of the docstring and lint/mypy check

redisvl/index.py
tylerhutcherson: will do!

PR #67, opened 2023-10-16
Add optional data load preprocessor hook

tylerhutcherson:
Another finding from working the arxiv example.. often you need to unpack or edit a record before writing to the source. If you do this before invoking redisvl, you add an additional loop, one that is unnecessary in the end. So this allows devs to optionally add a preprocessor method on load to call against each record.

### DIFF
--- a/redisvl/index.py
+++ b/redisvl/index.py
@@ -403,9 +403,13 @@ class SearchIndex(SearchIndexBase):
                         try:
                             record = preprocess(record)
                         except Exception as e:
-                            raise RuntimeError("Error while preprocessing records on load") from e
+                            raise RuntimeError(
+                                "Error while preprocessing records on load"
+                            ) from e
                     if not isinstance(record, dict):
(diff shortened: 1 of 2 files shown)


Spartee (changes requested):
A couple comments

1. What happens if the process function fails? Should we be catching errors there? leave it to the user?
2. What happens if it doesn't return a dictionary type? I think we should throw and error there at least right?
3. maybe put an example into the load function with a simple process function. 

lastly, this seems to overlap commits with #66 just FYI.

tylerhutcherson:
> A couple comments
> 
> 1. What happens if the process function fails? Should we be catching errors there? leave it to the user?
> 2. What happens if it doesn't return a dictionary type? I think we should throw and error there at least right?
> 3. maybe put an example into the load function with a simple process function.
> 
> lastly, this seems to overlap commits with #66 just FYI.

Good feedback -- i will address these!

Spartee (approved):
Thanks for addressing this feedback! LGTM

PR #68, opened 2023-10-18
Improve VectorField schema default args and tests

tylerhutcherson:
By default, the `VectorField`'s in Redis do NOT need to have the block size or initial cap args set. This change allows for those params to be set, and only included in the field args if so. Otherwise, they are ignored. Also include a small refactor on the schema classes for vectors as well as new schema unit tests.


Spartee (approved):
Tested locally, worked well.

PR #69, opened 2023-10-18
Add tests for token escaper class

tylerhutcherson:
Adds a set of unit tests on both the underlying token escaping class as well as the `Tag` filterable fields that utilize it.


Spartee (approved):
Added some test names, but otherwise looks great!

PR #70, opened 2023-10-24
redis-py bump

chayim:
- release drafter
- redis-py bump


Spartee (commented):
How does this work with the workflow_dispatch release mechanism we have today? Will this automatically supplant that?

tylerhutcherson:
@chayim looks good to me -- can you reply to Sam here just to close the loop and then I can merge :)

tylerhutcherson (commented):
It looks like we already have the release-drafter config and workflow on the main branch? But maybe I am missing something. Overall - LGTM.

tylerhutcherson (approved):

PR #72, opened 2023-10-30
Improve RedisVL filter handling, docs, and examples

tylerhutcherson:
Addresses https://github.com/RedisVentures/redisvl/issues/71 which highlights the need to support more edge case queries like "empty queries" or more sophisticated text LIKE query patterns.

- Improves module readability and testability for the individual filter classes.
- Adds support for various filter edge cases (described in the issue above).
- Also fixes a bug in the index creation routine that originally ignored the `overwrite=False` case.
- Adds more docstrings, unit tests and integration tests to improve our coverage on these cases.
- Improves documentation coverage for hybrid queries and getting started.


redisvl/index.py
```diff
@@ -510,14 +520,26 @@ async def create(self, overwrite: Optional[bool] = False):
         Raises:
             redis.exceptions.ResponseError: If the index already exists.
         """
-        exists = await self.exists()
-        if exists and overwrite:
+        # TODO - enable async version of this
+        # check_redis_modules_exist(self._redis_conn)
+
+        if not self._fields:
```
Spartee, line 526: should this be allowed?

redisvl/index.py
```diff
@@ -499,6 +508,7 @@ def connect(self, url: Optional[str] = None, **kwargs):
             ValueError: If no Redis URL is provided and REDIS_URL env var is not set.
         """
         self._redis_conn = get_async_redis_connection(url, **kwargs)
+        return self
```
Spartee, line 511: why?

redisvl/index.py
```diff
@@ -317,6 +318,7 @@ def connect(self, url: Optional[str] = None, **kwargs):
             ValueError: If the REDIS_URL env var is not set and url is not provided.
         """
         self._redis_conn = get_redis_connection(url, **kwargs)
+        return self
```
Spartee, line 321: this feels odd given that then... it's a classmethod ... kindaof

Is this to make things more fluent?

Spartee (dismissed):
I'm approving, however, I think we should leave out the index updating (or maybe move somewhere other than getting started). I think it makes things seem more complicated and I think we want that to be the lowest bar to cross possible.

Other than that, I pushed up a few changes. Biggest thing is the change of ``_set_value``. Essentially, I don't think it's necessary to treat anything except for tag differently. I've made a few tests for this as well and updated the hybrid query notebook.

redisvl/index.py
tylerhutcherson: yes, it reduces the potential lines of code for a user by allowing `index = SearchIndex.from_dict(...).connect(...)`

redisvl/index.py
tylerhutcherson: See above

redisvl/index.py
tylerhutcherson: This is the current behavior of the synchro version of the `create` method, and was just missing here

Spartee (approved):
LGTM

PR #73, opened 2023-11-01
Test multiple versions of Python, Redis Stack, and hiredis option

Spartee:
Building off of Chayims Pr in #62, This PR extends the test suite capability to more variations supported by Redis.


tylerhutcherson:
Just pushed a small commit / fix that should handle the failing tests. LGTM thanks guys.

PR #78, opened 2023-11-17
Refactor Redis Storage and add JSON support

tylerhutcherson:
PR introduces a new `BaseStorage` class, subclassed by `HashStorage` and `JsonStorage`, to handle the underlying data structures (as well as I/O) for Redis. Includes:
- New storage classes and type enums
- New `key_separator` param for better use-customizability when constructing redis keys
- Updated documentation and new user guide
- Updated docstrings and examples


redisvl/index.py
```diff
@@ -1,91 +1,160 @@
...
+        Args:
+            name (str): Index name.
+            prefix (str, optional): Key prefix associated with the index. Defaults to "rvl".
+            storage_type (str, optional): Underlying Redis storage type (hash or json). Defaults to "hash".
+            key_separator (str, optional): : Separator character to combine
+                prefix and key value for constructing redis keys. Defaults to ":".
+            fields (Optional[List[Field]], optional): List of Redis fields to index. Defaults to None.
+        """
+        # configure index and storage specs
+        # @ Tyler: I think we keep the init args in this release
```
Spartee: comment

redisvl/schema.py
```diff
@@ -17,6 +18,10 @@
 class BaseField(BaseModel):
     name: str = Field(...)
     sortable: Optional[bool] = False
+    as_name: Optional[str] = None
+
+
+# @Tyler Do we want to include Extra Field in this PR?
```
Spartee: comment removeal

tylerhutcherson:
@Spartee good to go from my end.

docs/api/searchindex.rst
```diff
@@ -13,18 +13,23 @@ SearchIndex
 .. autosummary::
 
     SearchIndex.__init__
+    SearchIndex.client
+    SearchIndex.name
+    SearchIndex.prefix
+    SearchIndex.key_separator
+    SearchIndex.storage
```
Spartee: Shouldn't be documented.

redisvl/index.py
```diff
@@ -371,61 +486,94 @@ def delete(self, drop: bool = True):
         self._redis_conn.ft(self._name).dropindex(delete_documents=drop)  # type: ignore
 
     @check_connected("_redis_conn")
+    @check_modules_present("_redis_conn")
```
Spartee, line 489: Can we use ``lru_cache`` to cache the calls to this within an invocated client?

Spartee (commented):
Nearly there. Looks good to me. Running tests now.

redisvl/index.py
tylerhutcherson: hmm i've not done that. are you saying cache the module list client side? I was going to recommend some kind of private attribute on the class like `_modules` that gets populated first time you run it. Then next time the check is called, it looks at this attribute first, and then executes the redis command. Something like that. Lmk what you're thinking? Feels like a later optimization to me

docs/api/searchindex.rst
tylerhutcherson: fixing

Spartee (approved):
LGTM

PR #80, opened 2023-11-21
Semantic cache interface update

tylerhutcherson:
Goals of this PR focus on improving the semantic cache interface in RedisVL:

1. Unify interface

    - Remove the ability to pass in an index. This is not the preferred methodology and currently breaks if the schema doesn't perfectly match up. Remove references from Docs.
    - Naming Conventions:
        - ``index_name`` ➝ ``name``.  Add deprecation warning.
        - ``threshold`` --> ``distance_threshold``. Add deprecation warning.
        - For the ``check()`` method, change ``fields`` --> ``return_fields``. Add deprecation warning.
    - Update all main cache interfaces, docs, and tests with these changes.

2. Improve doc strings and error handling

3. Wrap comments. no longer than 80 lines.

**FUTURE WORK**:
- Add support for creating a cache from an index, or passing in a custom schema.
- Add async semantic cache option.
- Build a more real-world use case example.

### DIFF
--- a/docs/user_guide/llmcache_03.ipynb
+++ b/docs/user_guide/llmcache_03.ipynb
@@ -6,16 +6,16 @@
    "source": [
     "# Semantic Caching\n",
     "\n",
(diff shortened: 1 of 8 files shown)


tylerhutcherson:
@Spartee 2 quick things:

I rebased off of JSON support. Once that's in I will update this branch from tip of main and get it right.

Check for a few TODO comments where I'd like your input / take and ensure we're aligned on approach.

PR #85, opened 2023-11-28
IndexSchema refactor and async collapse

tylerhutcherson:
This PR introduces a new user space class, `IndexSchema`. This effort to refactor the index schema is key for us to be able to enable "wrappability" and extensibility of RedisVL with other integration efforts (i.e. LangChain), and also improve the UX of working with the tool overall.

A few changes:
- SearchIndex now expects the schema to be provided on init, or built with some of the constructors (common).
- SearchIndex now optionally expects redis_url or redis_client
- An initial MVP of the RedisConnection class was built to support working with both sync and async connections from redis-py. A better implementation will be coming in the near term.
- Async and Sync search index were combined. No more `AsyncSearchIndex` class here.
- Other updates and docs improvements.

Adding link to current notes on the schema spec: https://docs.google.com/document/d/1vkShf9PoQd1FnjYa2PpKhcgk8FW9WZGi_KpCuoDtvEY/edit?usp=sharing


redisvl/schema/fields.py
```diff
@@ -0,0 +1,115 @@
...
+            "TYPE": self.datatype,
+            "DIM": self.dims,
+            "DISTANCE_METRIC": self.distance_metric,
+        }
+        if self.initial_cap is not None:  # Only include it if it's set
+            field_data["INITIAL_CAP"] = self.initial_cap
+        return field_data
+
+
+class FlatVectorFieldSchema(BaseVectorField):
```
Spartee: Shouldn't these be kept as just ``xxField``? Fields go into the schema?

redisvl/schema/schema.py
```diff
@@ -0,0 +1,306 @@
...
+    if not fp.exists():
+        raise FileNotFoundError(f"Schema file {file_path} does not exist")
+
+    with open(fp, "r") as f:
+        schema = yaml.safe_load(f)
+
+    return Schema(**schema)
+
+
+class SchemaGenerator:
```
Spartee: I think this should be a classmethod that returns a ``Schema`` object. Any reason this should be a long lived object?

redisvl/schema/schema.py
```diff
@@ -0,0 +1,306 @@
...
+
+class FieldsModel(BaseModel):
+    tag: Optional[List[TagFieldSchema]] = None
+    text: Optional[List[TextFieldSchema]] = None
+    numeric: Optional[List[NumericFieldSchema]] = None
+    geo: Optional[List[GeoFieldSchema]] = None
+    vector: Optional[List[Union[FlatVectorFieldSchema, HNSWVectorFieldSchema]]] = None
+
+
+class SchemaValidationError(Exception):
```
Spartee: ok so the thoughts on validation
- we may need to validate on more than just init. i.e. after a new vector field is added. 
- We should ideally be relying on the pydantic validators here and throwing our exception if we choose to make it cleaner.
- starting our own line of exceptions takes on a whole new part of this library that wasn't in the original design. Are we sure we don't just want to throw ``ValueError``? If we do take on exceptions, I think it should be in a separate PR where we introduce docs and use them where appropriate.

redisvl/schema/schema.py
```diff
@@ -0,0 +1,306 @@
...
+    make_dict,
+)
+
+
+class StorageType(Enum):
+    HASH = "hash"
+    JSON = "json"
+
+
+class IndexModel(BaseModel):
```
Spartee: At this point I think we just collapse ``IndexModel`` and ``FieldsModel`` into the Schema object. No reason to nest them from what I see here.

redisvl/schema/__init__.py
```diff
@@ -0,0 +1,38 @@
...
+    SchemaValidationError,
+)
+
+
+__all__ = [
+    "BaseField",
+    "TagFieldSchema",
+    "TextFieldSchema",
+    "GeoFieldSchema",
+    "NumericFieldSchema",
```
Spartee: do these fields need to be exported? I'm assuming only 
- ``Schema``
- ``StorageType``

need to be exported. We want to maintain a wall between what we see and the user sees in the importable space.

redisvl/llmcache/semantic.py
```diff
@@ -7,22 +7,22 @@
...
-    _vector_field_name: str = "prompt_vector"
-    _default_fields: List[Field] = [
-        VectorField(
-            _vector_field_name,
-            "FLAT",
-            {"DIM": 768, "TYPE": "FLOAT32", "DISTANCE_METRIC": "COSINE"},
-        ),
-    ]
+    _default_vector_field_name: str = "prompt_vector"
+    _default_vector_field: Dict[str, str] = {
```
Spartee: I think a default initialized ``Schema`` object makes sense here. Any user changes should use the user interface like we expect the user to when they manipulate the ``Schema`` object.

redisvl/index.py
```diff
@@ -169,78 +139,62 @@ class SearchIndexBase:
...
-        self._prefix = prefix
-        self._key_separator = key_separator
-        self._storage_type = StorageType(storage_type)
-        self._fields = fields
-
-        # configure storage layer
-        self._storage = self.STORAGE_MAP[self._storage_type](  # type: ignore
-            self._prefix, self._key_separator
+        # final validation on schema object
+        if not schema or not isinstance(schema, Schema):
```
Spartee: pass ``**kwargs`` to the schema init so it's not breaking from 0.0.5?

redisvl/index.py
```diff
@@ -282,29 +235,16 @@ def from_dict(cls, schema_dict: Dict[str, Any]):
...
-        cls,
-        name: str,
-        redis_url: Optional[str] = None,
-        key_separator: str = ":",
-        fields: Optional[List["Field"]] = None,
-        **kwargs,
-    ):
-        raise NotImplementedError
+        schema = Schema(**schema_dict)
+        return cls(schema=schema, **kwargs)
```
Spartee: connection_args?

redisvl/index.py
```diff
@@ -253,13 +207,13 @@ def from_yaml(cls, schema_path: str):
             >>> index.create(overwrite=True)
 
         Returns:
-            SearchIndex: A SearchIndex object.
+            SearchIndex: A RedisVL SearchIndex object.
         """
         schema = read_schema(schema_path)
-        return cls(fields=schema.index_fields, **schema.index.dict())
+        return cls(schema=schema, **kwargs)
```
Spartee: connection_args?

redisvl/index.py
```diff
@@ -379,48 +315,6 @@ class SearchIndex(SearchIndexBase):
         >>> index.load(data) # data is an iterable of dictionaries
     """
 
-    @classmethod
-    def from_existing(
```
Spartee, line 383: deprecate instead?

redisvl/index.py
```diff
@@ -462,14 +356,13 @@ def create(self, overwrite: bool = False) -> None:
             self.delete()
 
         # Create the index with the specified fields and settings.
-        self._redis_conn.ft(self._name).create_index(  # type: ignore
-            fields=self._fields,
+        self._redis_conn.ft(self.schema.index_name).create_index(  # type: ignore
```
Spartee: ``self.name``

redisvl/index.py
```diff
@@ -462,14 +356,13 @@ def create(self, overwrite: bool = False) -> None:
...
+            fields=index_fields,
             definition=IndexDefinition(
-                prefix=[self._prefix], index_type=self._storage.type
+                prefix=[self.schema.index_prefix], index_type=self._storage.type
```
Spartee: ``self.prefix``

redisvl/index.py
```diff
@@ -525,7 +417,7 @@ def load(
             >>> index.load(data, preprocess=func)
         """
         self._storage.write(
-            self.client,
+            self._redis_conn,
```
Spartee: why are we getting rid of the ``client`` property?

Spartee (commented):
Great start. Lots of thoughts on here. Would be good to sync on this and what you're doing with collapsing the interfaces. 

Biggest theme here is reduction in code duplication, I think we can collapse
- ``SchemaModel``
- ``FieldsModel``
- ``SchemaGenerator``

into
- ``Schema``

Let me know what you think. Great start!

redisvl/schema/fields.py
tylerhutcherson: So the other field types all already had `xxxSchema`, I changed the vector fields to match since they were misnamed in that context. We could strip this from all fields (which makes sense to me).

redisvl/schema/schema.py
tylerhutcherson: I agree. Was waiting to do this before we had a change to get some more alignment. Will do today

redisvl/schema/schema.py
tylerhutcherson: I think the benefit to keeping them here is for pydantic validation since Schema is not a pydantic object. Each has separate conditions and checks. Also makes it cleaner than accessing args from a dictionary like `self.index['prefix']`. Obviously a few routes to handle this:

1. remove the model classes and have everything be a first class citizen within the Schema object namespace `schema.name` schema.prefix` `schema.tags` etc
2. get rid of the models, but keep them bundles as dictionaries for organizational purposes and maybe expose property methods for each.
3. keep it how it is

redisvl/schema/__init__.py
tylerhutcherson: Yep will redact what I can!

redisvl/llmcache/semantic.py
tylerhutcherson: So we could have everything set up pre-initializer up here. default schema etc.

But when user initializes the cache with different params than those in the default, what is your suggestion for how we handle that? We don't currently have any schema "update" / "manipulation" concepts in the code.

redisvl/index.py
tylerhutcherson: force it? i.e. throw a warning but allow it? Or raise an exception that forces the user's hand to change their approach? I think it's safe to break it here, but would be ok with a depwarning for sure

redisvl/index.py
tylerhutcherson: we're not! it's use the internal version of the method for consistency, since that's what we have been doing everywhere else. but we could change that

redisvl/schema/schema.py
tylerhutcherson: agreed. will move to ValueError for now.

we should maybe take on exceptions later, and agreed in a different PR.

so if we remove the `IndexModel` and `FieldsModel` then we don't have validators anymore

redisvl/schema/schema.py
tylerhutcherson: I think there are some advantages to keeping it. Will push a new version of code and lets review together so we can discuss

redisvl/index.py
tylerhutcherson: what would this look like?

redisvl/index.py
Spartee: ok with either. Break is fine as long as we update to 0.1.0 which is the plan

redisvl/index.py
Spartee: ok

redisvl/index.py
Spartee: ensuring that we can pass it the existing init method paramteres and it still works. breaking the init method even in a minor release is tough for users.

redisvl/llmcache/semantic.py
Spartee: I think we have to have schema update/manipulation if it's a user space object.

redisvl/schema/schema.py
```diff
@@ -0,0 +1,299 @@
...
+
+class FieldsModel(BaseModel):
+    tag: List[TagField] = []
+    text: List[TextField] = []
+    numeric: List[NumericField] = []
+    geo: List[GeoField] = []
+    vector: List[Union[FlatVectorField, HNSWVectorField]] = []
+
+    def add(self, field: Union[BaseField, BaseVectorField]):
+        if isinstance(field, TagField):
```
Spartee: create a mapping instead? I do't like chained conditionals

redisvl/schema/schema.py
```diff
@@ -36,10 +36,28 @@ def get_vector_type(**field_data: Dict[str, Any]) -> Union[FlatVectorField, HNSW
...
+        "tag": TagField,
+        "text": TextField,
+        "numeric": NumericField,
+        "geo": GeoField,
+        "vector": get_vector_type
+    }
+
+    @staticmethod
+    def create_field(field_type: str, name: str, **kwargs) -> BaseField:
+        field_class = FieldFactory.FIELD_TYPE_MAP.get(field_type)
```
tylerhutcherson: is this the right way to do this?

PR #101, opened 2024-01-23
Upgrade to use Pydantic 2.x.x

tylerhutcherson:
With the shift in the ecosystem to Pydantic 2.x.x, we need to get `redisvl` to the same level.

Referenced here: https://github.com/RedisVentures/redisvl/issues/86

Fortunately, there is not a ton of complex usage of Pydantic; which makes this easy! Changes include:

- Shifting the storage classes to use Pydantic to avoid manual (and ugly) field validation. Now, the subclasses inherit from the base and params are all typed as expected.
- Shifting the vectorizer classes to use Pydantic for the same reasons.
- Bump and pin pydantic version in the packge requirements to a reasonable range `pydantic>=2.0.0,<3`
- use `v1` shim for now for safety

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -30,6 +30,39 @@ Here's how to get started with your code contribution:
 ### Dev Environment
 There is a provided `requirements.txt` and `requirements-dev.txt` file you can use to install required libraries with `pip` into your virtual environment.
 
+Or use the local package editable install method:
+```bash
+python -m venv .venv
+source .venv/bin/activate
+pip install -e .[all,dev]
+```
+
+Then to deactivate the env:
+```
+source deactivate
+```
+
+### Linting and Tests
+
+Formatting and linting checks:
+```bash
+make format
+make sort-imports
+make mypy
+```
+
+Tests with vectorizers:
+```bash
+make test-cov
+```
+
+Tests w/out vectorizers:
+```bash
+SKIP_VECTORIZERS=true make test-cov
+```
+
+> Dev requirements are needed here to be able to run tests and linting.
+
 ### Docker Tips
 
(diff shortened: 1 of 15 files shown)


CONTRIBUTING.md
```diff
@@ -30,6 +30,39 @@ Here's how to get started with your code contribution:
...
+Then to deactivate the env:
+```
+source deactivate
+```
+
+### Linting and Tests
+
+Formatting and linting checks:
+```bash
+make format
```
Spartee: make check?

requirements.txt
```diff
@@ -1,7 +1,7 @@
 numpy
-redis>=5.0.0
 pyyaml
 coloredlogs
-pydantic<2.0.0
+redis>=5.0.0
+pydantic==2.5.3
```
Spartee: may think about relaxing this a bit. Pinning to a patch version for this popular of a library is bound to cause some issues. more on this in the top level comment

Spartee (dismissed):
Undoubtly will cause some pain with the langchain integration. However, I think as long as there is no inheritance chain that utilizes pydantic v2 with langchain internals we should be fine. the langchain integration should just use ``langchain.pydantic_v1`` for any instantiated classes. As well, any base classes we expect to use in that integration should also utilize ``langchain.pydantic_v1`` if at all possible.

Also, as this is a library we expect to have people use as a dependency within their projects, I would expect to have a wider range of support for pydantic versions even if we have to test multiple versions (we shouldnt, but you never know).

nit: might think about breaking up some of these non-related changes just in case you need to roll this back if it causes major issues at some point down the road. 

Up to you, otherwise looks good

requirements.txt
tylerhutcherson: Will do. Good call

PR #104, opened 2024-01-31
Redis connection refactor

tylerhutcherson:
For ultimate clarity and safety of our library, the decision was made to separate the sync and async search index modules for now until we can better utilize the redis-py client to serve both needs simultaneously.

This PR splits `SearchIndex` and `AsyncSearchIndex` into two classes. It also refactors the way RedisVL handles Redis connections (much more lightweight) and makes some major improvements to the overall documentation on these topics.


tylerhutcherson:
Will address the concerns from https://github.com/RedisVentures/redisvl/issues/88

PR #108, opened 2024-02-01
Changes to contributions page

bsbodden:
- Highlight the need for `REDIS_URL` env var
- Add quotes around `pip install -e` for compatibility with all shells


tylerhutcherson (approved):
perfect

PR #119, opened 2024-02-13
Update and pin package dependency ranges

tylerhutcherson:
Loosen restriction on `tenacity` and `tabulate` to make it easier to integrate.


bsbodden (approved):
LGTM! 🚀 Ship it!

PR #126, opened 2024-02-29
OpenAI API client upgrade and code migration

bsbodden:
Upgrade OpenAI Python lib to >1.0.0 and migrates the code following https://github.com/openai/openai-python/discussions/742#async-client

Also, conditional skipping logic for vectorizers has been fixed (it was always skipping)


docs/user_guide/llmcache_03.ipynb
```diff
@@ -35,11 +35,13 @@
     "\n",
     "api_key = os.getenv(\"OPENAI_API_KEY\") or getpass.getpass(\"Enter your OpenAI API key: \")\n",
     "\n",
-    "openai.api_key = api_key\n",
+    "client = OpenAI(\n",
+    "  api_key=os.environ['OPENAI_API_KEY'],  # this is also the default, it can be omitted\n",
```
tylerhutcherson: We need to pass the `api_key` object extracted from above into here

requirements-all.txt
```diff
@@ -1,4 +1,4 @@
-openai<=0.28.1
+openai<=1.13.3
```
tylerhutcherson: Wondering if it makes sense to pin in the opposite direction? Maybe >= `1.13.0`

tests/integration/test_vectorizers.py
```diff
@@ -9,17 +9,12 @@
...
 def skip_vectorizer() -> bool:
     # os.getenv returns a string
     v = os.getenv("SKIP_VECTORIZERS", "False").lower() == "true"
     print(v, flush=True)
-    return v
-
-
-skip_vectorizer_test = lambda: pytest.config.getfixturevalue("skip_vectorizer")
-
+    return True
```
tylerhutcherson: Wouldn't this always be true and thus never run the test?

tylerhutcherson (changes requested):

requirements-all.txt
bsbodden: Indeed, I missed that

tests/integration/test_vectorizers.py
bsbodden: Bob Saget! sorry, that was me debugging and leaving the test behind

tylerhutcherson (approved):

tylerhutcherson:
Once CI passes, good to go. Will merge.

PR #129, opened 2024-03-12
fix connection pool for SSL

tylerhutcherson:
When an Async Redis SSLConnection (ACRE for example) is used, the method to check modules was failing. The module check has to be done with a sync connection because it is a decorator around certain `SearchIndex` methods.


bsbodden (commented):
LGTM

PR #134, opened 2024-04-04
Poetry for Dependency Management and TestContainers for Integration Testing

bsbodden:
In this PR:
- Uses Poetry for Dependency Management
- Replaces the Makefile with Python scripts that can be run from Poetry
- Implements TestContainers with a Docker Compose file for integration testing


bsbodden:
@tylerhutcherson Test will likely fail in CI since we would need to configure Github Actions correctly to use TestContainers. I will look into that soon.

pyproject.toml
```diff
@@ -1,3 +1,71 @@
+[tool.poetry]
+name = "redisvl"
+version = "0.1.2"
+description = "Python client library and CLI for using Redis as a vector database"
+authors = ["Redis Inc. <sam.partee@redis.com>"]
```
tylerhutcherson: we should change this, maybe the field engineering alias?

pyproject.toml
```diff
@@ -1,3 +1,71 @@
...
+license = "MIT"
+readme = "README.md"
+homepage = "https://github.com/RedisVentures/redisvl"
+repository = "https://github.com/RedisVentures/redisvl"
+documentation = "https://www.redisvl.com"
+keywords = ["ai", "redis", "redis-client", "vector-database", "vector-search"]
+classifiers = [
+  "Programming Language :: Python :: 3.7",
```
tylerhutcherson: Thinking we might consider EOL Python 3.7 support...

tylerhutcherson: We can update that later

pyproject.toml
```diff
@@ -1,3 +1,71 @@
...
+[tool.poetry.dependencies]
+python = ">=3.9,<4.0"
+numpy = "*"
+pyyaml = "*"
+coloredlogs = "*"
+redis = ">=5.0.0"
+pydantic = { version = ">=2,<3" }
+tenacity = ">=8.2.2"
+tabulate = { version = ">=0.9.0,<1" }
+openai = ">=1.13.0"
```
tylerhutcherson: These from openai and below need to be carved out into extras. They aren't required for `redisvl` to be installed today.

```
pip install redisvl[cohere]
```

for example

pyproject.toml
```diff
@@ -1,3 +1,71 @@
...
+google-cloud-aiplatform = ">=1.26"
+cohere = ">=4.44"
+
+[tool.poetry.dev-dependencies]
+black = ">=20.8b1"
+isort = ">=5.6.4"
+pylint = ">=2.6.0"
+pytest = ">=6.0.0"
+pytest-cov = ">=2.10.1"
+pytest-asyncio = "*"
```
tylerhutcherson: maybe we should look at pinning these to be safe. We have had some dev env flops where it's just a matter of a different mypy version locally than in GH actions

scripts.py
```diff
@@ -0,0 +1,36 @@
+import subprocess
```
tylerhutcherson, line 1: Is this the typical pattern with Poetry (subprocess calls?)? If so looks good. I see some places that still use Make (langchain) for example. Seems to be a matter of preference these days ;)

tests/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: "3.9"
```
tylerhutcherson, line 1: Curious how the test containers will integrate with GH Actions and it's matrix support

scripts.py
bsbodden: Yeah, I've seen both, I just went with throw more Python at the problem approach. I'll check if there is a new fangled way to do this better in Poetry

.github/workflows/lint.yml
```diff
@@ -30,19 +30,17 @@ jobs:
...
-
+        poetry run check-sort-imports
     - name: check-black-format
       run: |
-        make check-format
-
-    - name: check-mypy
-      run: |
-        make mypy
+        poetry run check-format
```
bsbodden, line 43: @tylerhutcherson I forgot I commented `mypy` out since it was failing both locally and on CI. I'll investigate what the deal is.

bsbodden: This is fixed now

.github/workflows/publish.yml
```diff
@@ -18,10 +18,12 @@ jobs:
...
+        uses: snok/install-poetry@v1
+
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip
-          python -m pip install setuptools wheel
+          poetry install --all-extras
 
       - name: create python package
         run: python setup.py sdist bdist_wheel
```
tylerhutcherson: We need to address this either here or in a quick follow PR (`setup.py` no longer exists) and this will not release to pypi currently!

bsbodden: Oh duh, let me address that, shouldn't be long

bsbodden: Just pushed a change to the publish.yml, totally missed the create python package, that should be just a `poetry publish` - let me know if the changes do the trick.

tylerhutcherson:
@bsbodden I am getting some env issues locally when trying to `poetry install --all-extras` here in a clean virtual env:
The first time it ran most of the way but got caught on `docopt` and then when run again, it picks back up, and same error message. Any ideas here?
```
$ poetry install --all-extras
Installing dependencies from lock file

Package operations: 19 installs, 0 updates, 0 removals

  • Installing docopt (0.6.2): Failed

  TypeError

  Can't instantiate abstract class IsolatedEnv with abstract methods make_extra_environ, python_executable

  at ~/miniconda3/lib/python3.9/site-packages/poetry/installation/chef.py:109 in _prepare
      105│     ) -> Path:
      106│         from subprocess import CalledProcessError
      107│ 
      108│         with ephemeral_environment(self._env.python) as venv:
    → 109│             env = IsolatedEnv(venv, self._pool)
      110│             builder = ProjectBuilder(
      111│                 directory,
      112│                 python_executable=env.executable,
      113│                 scripts_dir=env.scripts_dir,
```

pyproject.toml
```diff
@@ -1,5 +1,78 @@
...
+readme = "README.md"
+homepage = "https://github.com/RedisVentures/redisvl"
+repository = "https://github.com/RedisVentures/redisvl"
+documentation = "https://www.redisvl.com"
+keywords = ["ai", "redis", "redis-client", "vector-database", "vector-search"]
+classifiers = [
+  "Programming Language :: Python :: 3.9",
+  "Programming Language :: Python :: 3.10",
+  "Programming Language :: Python :: 3.11",
+  "License :: OSI Approved :: MIT License",
```
rbs333, line 16: this needed the OSI approved thing: https://pypi.org/classifiers/

.github/workflows/publish.yml
```diff
@@ -36,8 +38,7 @@ jobs:
...
 
-      - name: deploy
-        uses: pypa/gh-action-pypi-publish@release/v1
-        with:
-          user: __token__
-          password: ${{ secrets.PYPI }}
\ No newline at end of file
+      - name: Publish to PyPI
+        env:
+          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI_TOKEN }}
```
tylerhutcherson:
```suggestion
          POETRY_PYPI_TOKEN_PYPI: ${{ secrets.PYPI }}
```

.github/workflows/test-publish.yml
```diff
@@ -11,20 +11,22 @@ jobs:
...
-          python -m pip install --upgrade pip
-          python -m pip install setuptools wheel
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
 
-      - name: create python package
-        run: python setup.py sdist bdist_wheel
+      - name: Build package
+        run: |
+          poetry version ${{ github.event.inputs.version }}
```
tylerhutcherson, line 28: this version command is used here when it's not used in the other one?

.github/workflows/test-publish.yml
```diff
@@ -36,9 +38,9 @@ jobs:
           tag: ${{ github.event.inputs.version }}
           commit: main
 
-      - name: deploy
+      - name: Publish to TestPyPI
         uses: pypa/gh-action-pypi-publish@release/v1
         with:
           user: __token__
-          password: ${{ secrets.PYPI }}
+          password: ${{ secrets.TEST_PYPI_API_TOKEN }}
```
tylerhutcherson, line 45: We don't have a test PYPI token yet -- but also this step looks different than the one from the prod release?

PR #139, opened 2024-04-17
Initial reranker integration in RedisVL

tylerhutcherson:
This PR introduces another utility class into RedisVL to improve the querying experience, and result quality, through integration with 3rd party rerankers. We have started with Cohere's reranking API and will also add additional options later on.

TODO (with Cohere):

- [x] Confirm proper Cohere API param usage. (Clarify best way to use `max_chunks_per_doc`)
- [x] Confirm backwards compatibility between Cohere v2 and v3. Best way to add this here?
- [x] Confirm public method names. 
- [x] Finalize doc strings and examples.
- [x] Add to docs.
- [x] Add initial testing suite.


redisvl/utils/rerank/cohere.py
```diff
@@ -0,0 +1,115 @@
...
+            model=model,
+            rank_by=rank_by,
+            limit=limit,
+            return_score=return_score,
+        )
+
+    @staticmethod
+    def _preprocess(results: List[Dict[str, Any]], rank_by: str) -> List[str]:
+        try:
+            docs = [result[rank_by] for result in results]
```
nabsabraham: Technically, we don't need to preprocess documents to slice out the json fields. Rerankv3 takes in json docs and a `rank_fields` param (list of strings) and our api slices the json doc by those rank fields to then pass to rerank model. 

We could keep this preprocess function to validate the data. IE, if the data **is** a list of json documents, you **must** have `rank_fields`. All fields needed for reranking must be explicitly passed in the rerank call (for now). 

If instead the documents are a list of strings, then `rank_fields` should not be passed to the rerank call.

redisvl/utils/rerank/cohere.py
```diff
@@ -0,0 +1,115 @@
...
+        max_chunks_per_doc: Optional[int] = None,
+        **kwargs,
+    ) -> List[Dict[str, Any]]:
+        limit = kwargs.get("limit", self.limit)
+        return_score = kwargs.get("return_score", self.return_score)
+        rank_by = kwargs.get("rank_by", self.rank_by)
+
+        docs = self._preprocess(results, rank_by)
+
+        rankings = self._client.rerank(
```
nabsabraham: I think we'd want `rank_fields` to equal the `rank_by` param

redisvl/utils/rerank/cohere.py
```diff
@@ -0,0 +1,115 @@
...
+            result = results[item.index]
+            if return_score:
+                result["score"] = item.relevance_score
+            reranked_results.append(result)
+        return reranked_results
+
+    def rank(
+        self,
+        query: str,
+        results: List[Dict[str, Any]],
```
nabsabraham: would this always be a `List[Dict[str, Any]]` or do you see it being more like `Union[List[str], List[Dict[str, Any]]`
If its the former, then we would always expect a `rank_by` parameter because its suggesting the documents are always a list of documents and rerank will expect `rank_fields` to be passed for json documents

nabsabraham (changes requested):
left some suggestions on what the documents need to look like to pass to the rerank call. TLDR: 
* If we want, we are free to send `rerank` json docs (ie, list of dicts) but we just need to make sure to pass `rerank` a list of `rank_fields` that are the keys in the json docs 
* If we want to instead handle the slicing of json fields on the client side, we can do so and then pass rerank a list of strings and then omit passing `rank_fields` 
* We should default to `rerank-english-v3.0`

Apologies if I'm missing some of the requirements we discussed a few months ago regarding the integration of the new reranker in redis.

tylerhutcherson:
> left some suggestions on what the documents need to look like to pass to the rerank call. TLDR:
> 
> * If we want, we are free to send `rerank` json docs (ie, list of dicts) but we just need to make sure to pass `rerank` a list of `rank_fields` that are the keys in the json docs
> * If we want to instead handle the slicing of json fields on the client side, we can do so and then pass rerank a list of strings and then omit passing `rank_fields`
> * We should default to `rerank-english-v3.0`
> 
> Apologies if I'm missing some of the requirements we discussed a few months ago regarding the integration of the new reranker in redis.

Great feedback. I definitely missed the API spec update for v3. My concern with moving to rank_fields was backwards compatibility and the right way to do that. But I think I can updated `rank_by` to be a list of strings. That way it can be flexible across different reranking implementations. I will work on an update here shortly, and tag you again :) thanks!

redisvl/utils/rerank/cohere.py
tylerhutcherson: Great input, I can make that a bit more flexible and then just check the data before calling API

redisvl/utils/rerank/cohere.py
tylerhutcherson: Will do

redisvl/utils/rerank/cohere.py
```diff
@@ -0,0 +1,185 @@
...
+        reranker_kwargs = {
+            "model": self.model,
+            "query": query,
+            "top_n": limit,
+            "documents": docs,
+            "max_chunks_per_doc": max_chunks_per_doc,
+        }
+        # if we are working with list of dicts
+        if all(isinstance(doc, dict) for doc in docs):
+            if rank_by:
```
nabsabraham, line 112: as a sanity check, it might be good to validate the `rank_by` fields are actually keys in the list of dicts -- on the cohere side, if the keys are not the rank_fields, it'll complain that you sent in an empty text input which might be confusing

tests/integration/test_rerankers.py
```diff
@@ -0,0 +1,66 @@
...
+
+    reranked_docs, scores = await reranker.arank(query, docs)
+
+    assert isinstance(reranked_docs, list)
+    assert len(reranked_docs) == len(docs)  # Ensure we get back as many docs as we sent
+    assert all(isinstance(score, float) for score in scores)  # Scores should be floats
+
+
+# Test handling of bad input
+def test_bad_input(reranker, skip_reranker):
```
nabsabraham: love!

nabsabraham (approved):
very clean! lgtm 🎉

PR #141, opened 2024-04-23
Introduce the LLM session manager classes

justin-cechmanek:
Here's the proposed class for LLM session management. It support recent or full conversation storage & retrieval, as well as relevance based conversation section retrieval.


redisvl/extensions/sessionmanager/__init__.py
```diff
@@ -0,0 +1,3 @@
+from redisvl.extensions.sessionmanager.session import SessionManager
```
tylerhutcherson: wondering if we should name the module `session_manager` for better readability. WDYT?

redisvl/extensions/sessionmanager/session.py
```diff
@@ -0,0 +1,129 @@
...
+        as exchanges.
+
+        Args:
+            from_file Optional[str]: File to intiaialize the cache index with.
+            vectorizer Vectorizer: The vectorizer to create embeddings with.
+
+        The proposed schema will support a single combined vector embedding constructed
+        from the prompt & response in a single string.
+
+        schema = IndexSchema.from_dict({"index": {"name": name, "prefix": prefix}})
```
tylerhutcherson: what are you thinking regarding the index "partitioning"? I imagine we will want to give the end user the ability to optionally set their own index name -- potentially for a different application, use case, or set of users

tylerhutcherson: Also will this be HASH or JSON? should it be configurable?

redisvl/extensions/sessionmanager/session.py
```diff
@@ -0,0 +1,129 @@
...
+        schema.add_fields(
+            [
+                {"name": prompt_field, "type": "text"},
```
tylerhutcherson: Will also need the other fields like timestamp, user id, and session id to support query operations that can segment to those levels

redisvl/extensions/sessionmanager/session.py
```diff
@@ -0,0 +1,129 @@
...
+                          or list of alternating prompts and responses.
+
+        Returns:
+            Union[str, List[str]]: A single string transcription of the session
+                                   or list of strings if as_text is false.
+        """
+        pass
+
+
+    def _order_by(self, exchanges: List[str], recency: bool = True) ->  List[str]:
```
tylerhutcherson: this _should_ be achievable on the Redis search API using the `.sort_by()` operator, but depends on how or where this is needed in the flow

redisvl/extensions/sessionmanager/session.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+    def load_previous_sessions(self):
+        """ Checks the Redis instance for previous sessions from this user and
+            loads them into session memory if they exist.
+        """
+        pass
+
+
+    def store(self, exchange: Tuple[str, str]):
```
tylerhutcherson: we should also think about efficient loading of batches or bulk conversation history

tylerhutcherson (commented):
Left some feedback! Great start. From an interface design perspective, I'd think about what are the key public methods that the user will have access to in order to manipulate or read from the session state. this is what I was able to derive :) 

- `conversation_history` - loads full conversation history. What about just last 10 messages (the naive session retrieval approach)?
- `fetch_context` - fetches session data based on semantic similarity to text input
- `load_previous_sessions` - loads previous session data (how is this diff than the first method)
- `store` - store exchanges into the db!

redisvl/extensions/sessionmanager/__init__.py
justin-cechmanek: I wasn't sure what style to follow, but yeah, I agree `session_manager` is better.

redisvl/extensions/sessionmanager/session.py
justin-cechmanek: The plan is to use Hashes. I don't think we'll need a nested structure for this data.

redisvl/extensions/sessionmanager/session.py
justin-cechmanek: Added `timestamp` and `user_id`. I'm rethinking the need for `session_id` given we have timestamps. Let's chat about this.

redisvl/extensions/session_manager/session.py
```diff
@@ -28,22 +46,40 @@ def __init__(self,
...
+            prefix = name
+        if not session_id:
+            self._session_id = "any_session"
+        if not user_id:
+            self._user_id = "any_id"
+        if not application_id:
+            self._application_id = "any_app"
+
+        if vectorizer is None:
+            self._vectorizer = HFTextVectorizer(
```
tylerhutcherson: This is a great starting point for the vectorization. Curious if there are any open source sentence transformers that might be ideal for this kind of task

redisvl/extensions/session_manager/session.py
```diff
@@ -0,0 +1,293 @@
...
+                    response.
+
+            Returns:
+                str: The Redis key for the entry added to the database.
+        """
+        count_key = ":".join([self._application_id, self._user_id, self._session_id, "count"])
+        count = self._redis_client.incr(count_key)
+        vector = self._vectorizer.embed(exchange[0] + exchange[1])
+        timestamp = int(datetime.now().timestamp())
+        payload = {
```
tylerhutcherson: We should look at using Pydantic models for these -- helps with modeling and automatic type validation / checks

redisvl/extensions/session_manager/session.py
```diff
@@ -0,0 +1,293 @@
...
+            "timestamp",
+        ]
+
+        count_key = ":".join([self._application_id, self._user_id, self._session_id, "count"])
+        count = self._redis_client.get(count_key) or 0
+        last_k_filter = Num("count") > int(count) - top_k
+        combined = self._tag_filter & last_k_filter
+
+        query = FilterQuery(
+                return_fields=return_fields,
```
tylerhutcherson: indention

redisvl/extensions/session_manager/session.py
```diff
@@ -0,0 +1,293 @@
...
+
+            Args:
+                exchange Tuple[str, str]: The user prompt and corresponding LLM
+                    response.
+
+            Returns:
+                str: The Redis key for the entry added to the database.
+        """
+        count_key = ":".join([self._application_id, self._user_id, self._session_id, "count"])
+        count = self._redis_client.incr(count_key)
```
tylerhutcherson: there should be a way for us to get this from this search index without a separate key! also, I wonder if we can just use the search API to grab last k. we might have to dip into the base redis py search capabilities, or extend the redisvl query classes to support the sorting mechanism you need here!

redisvl/extensions/session_manager/session.py
```diff
@@ -0,0 +1,293 @@
...
+            user_filter = Tag("user_id") == self._user_id
+            self._tag_filter = self._tag_filter & user_filter & session_filter
+
+
+    def clear(self):
+        """ Clears the chat session history. """
+        pass
+
+
+    def fetch_context(
```
tylerhutcherson: love it! I would push for `fetch_relevant` and `fetch_recent` for example to align the purposes of the interface and methods to end user (or something similar)

redisvl/extensions/session_manager/session.py
```diff
@@ -0,0 +1,293 @@
...
+from redisvl.index import SearchIndex
+from redisvl.query import FilterQuery, RangeQuery
+from redisvl.query.filter import Tag, Num
+from redisvl.redis.utils import array_to_buffer
+from redisvl.schema.schema import IndexSchema
+from redisvl.utils.vectorize import BaseVectorizer, HFTextVectorizer
+
+class SessionManager:
+    def __init__(self,
+                 name: str,
```
tylerhutcherson: indentation

tylerhutcherson (commented):
looking solid, left some more thoughts here

redisvl/extensions/sessionmanager/session.py
justin-cechmanek: Sorting is needed because sometimes we return results based on relevance, and sometimes by recency. It's being done as a final step currently,  but I plan to move it into `query.py` as a parameter.

redisvl/extensions/session_manager/session.py
justin-cechmanek: I can look around for some and add them as options.

justin-cechmanek: Sbert has a few models that are specifically for asymetric prompt:response retrieval, and are tuned for cosine similarity. I've changed the default here.

redisvl/extensions/session_manager/__init__.py
```diff
@@ -0,0 +1,3 @@
+from redisvl.extensions.session_manager.session import SessionManager
```
rbs333: Quick hitter @justin-cechmanek I noticed when pulling down the code it seems like this branch is behind the poetry changes that went in last week. I'm not sure why the diff isn't in here but we don't want to merge until that's resolved.

redisvl/extensions/session_manager/session.py
justin-cechmanek: I wasn't sure how to do this with hash docs. With lists we can just append and fetch a set number K entries. Is there a way to know the ordered of hashed entries without a separate key?

redisvl/extensions/session_manager/base_session.py
```diff
@@ -0,0 +1,145 @@
...
+
+from redisvl.redis.utils import array_to_buffer
+
+
+class BaseSessionManager:
+    def __init__(
+        self,
+        name: str,
+        session_id: str,
+        user_id: str,
```
rbs333: I wonder if we might want to call this user_tag instead? I think it aligns closer with its function. If I see user_id, I might think that's like an authentication thing etc. since there is a lot of preconception to that name.

redisvl/extensions/session_manager/base_session.py
```diff
@@ -0,0 +1,145 @@
...
+        LLM responses to allow for enriching future prompts with session
+        context. Session history is stored in prompt:response pairs referred to
+        as exchanges.
+
+        Args:
+            name (str): The name of the session manager index.
+            session_id (str): Tag to be added to entries to link to a specific
+                session.
+            user_id (str): Tag to be added to entries to link to a specific user.
+            preamble (str): System level prompt to be included in all context.
```
rbs333: I also prefer system_prompt over preamble because that aligns with OpenAI who I think is the most well known API.

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,307 @@
...
+        if vectorizer is None:
+            self._vectorizer = HFTextVectorizer(
+                model="sentence-transformers/msmarco-distilbert-cos-v5"
+            )
+
+        self.set_distance_threshold(distance_threshold)
+
+        schema = IndexSchema.from_dict({"index": {"name": name, "prefix": prefix}})
+
+        schema.add_fields(
```
rbs333, line 69: Style comment: might make a function for creating index and adding fields outside the __init__

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,307 @@
...
+        id_field = ":".join([self._user_id, self._session_id, str(timestamp)])
+        hash_id = self.hash_input(id_field)
+        payload = {
+            "id_field": hash_id,
+            "prompt": prompt,
+            "response": response,
+            "timestamp": timestamp,
+            "session_id": self._session_id,
+            "user_id": self._user_id,
+            "token_count": 1,  # TODO get actual token count
```
rbs333: pydantic? Also, I think this todo should be addressed.

redisvl/extensions/session_manager/standard_session.py
```diff
@@ -0,0 +1,168 @@
...
+            user_id str: Tag to be added to entries to link to a specific user.
+            as_text bool: Whether to return the conversation as a single string,
+                or list of alternating prompts and responses.
+            raw bool: Whether to return the full Redis hash entry or just the
+                prompt and response
+        Returns:
+            Union[str, List[str]]: A single string transcription of the session
+                or list of strings if as_text is false.
+        """
+        if top_k == 0:
```
rbs333, line 131: this logic confuses me a bit. 

If 0 => return self.format_context(<empty>)
if -1 => set zero and continue? 

would top_k <= 0 => return self.format_context() be more flexible or am I missing some functionality you are going for?

redisvl/extensions/session_manager/standard_session.py
```diff
@@ -0,0 +1,168 @@
...
+            prompt str: The user prompt to the LLM.
+            response str: The corresponding LLM response.
+        """
+        timestamp = datetime.now().timestamp()
+        payload = {
+            "id_field": ":".join([self._user_id, self._session_id, str(timestamp)]),
+            "prompt": prompt,
+            "response": response,
+            "timestamp": timestamp,
+            "token_count": 1,  # TODO get actual token count
```
rbs333: same as other comment

rbs333 (commented):
Left a couple comments mainly for clarification and style. In general, looks good!

redisvl/extensions/session_manager/base_session.py
tylerhutcherson: I hear ya. However, I don't think we should bleed the internals of Redis data structures through this interface. I would assert that though it could collide with authentication systems.. that's sort of the intention. That end users (who choose to include a user scope) could do so with an existing "schema". A good docstring here will pay off well for us too

redisvl/extensions/session_manager/base_session.py
tylerhutcherson: What if the preamble/system prompt is empty? When plugging into langchain or llama.. this notion would be handled client side, not in Redis

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,307 @@
...
+        The proposed schema will support a single combined vector embedding
+        constructed from the prompt & response in a single string.
+
+        """
+        super().__init__(name, session_id, user_id, preamble)
+
+        prefix = prefix or name
+
+        if vectorizer is None:
+            self._vectorizer = HFTextVectorizer(
```
tylerhutcherson: We will need to update how vectorization works in order to integrate upstream. Instead of a vectorizer class, we need to support a vectorizer callable. That way any embedding generation tool will do.

We could get clever and have a simple default way of handling this, but langchain vectorizers and llama index vectorizers (or custom) will need to be supported.

redisvl/extensions/session_manager/semantic_session.py
tylerhutcherson: agreed we should use pydantic to handle all of the validation here. It will add a tinge of latency, but we should be able to measure that.

tylerhutcherson (commented):
Left some comments too! Looking nice. Because we are introducing a new capability here into the library, we will need to discuss how to evaluate further and eventually roll this out.

redisvl/extensions/session_manager/base_session.py
justin-cechmanek: Good point. I'll swap _id to _tag for users and sessions. We don't have any requirement that this align with other application user_ids so tag feels more permissive.

redisvl/extensions/session_manager/standard_session.py
justin-cechmanek: The goal is to support `top_k=-1` returning the full conversation history, which means passing 0 (or -0) to lrange() as the first index.

rbs333: Cool I can roll with that. Tbh not the most declarative imo and I tend to bias in that direction

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,364 @@
...
+        The proposed schema will support a single vector embedding constructed
+        from either the prompt or response in a single string.
+
+        """
+        super().__init__(name, session_tag, user_tag)
+
+        prefix = prefix or name
+
+        if vectorizer is None:
+            self._vectorizer = HFTextVectorizer(
```
tylerhutcherson: It would seem that there's no way to actually set your own vectorizer here. What if a user providers a `vectorizer` object?

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def drop(self, id_field: Optional[str] = None) -> None:
+        """Remove a specific exchange from the conversation history.
+
+        Args:
+            id_field Optional[str]: The id_field of the entry to delete.
+                If None then the last entry is deleted.
+        """
+        if id_field:
+            key = ":".join([self._index.schema.index.name, id_field])
```
tylerhutcherson: `key_separator` is an index schema property that should be used here. The default is `:` but it could be other things. Safest to use the schema's value here as a best practice.

tylerhutcherson (approved):
Great work. Left two small notes, but after we address those, let's get this merged!

redisvl/extensions/session_manager/semantic_session.py
justin-cechmanek: That's a mistake. Nice catch.

redisvl/extensions/session_manager/base_session.py
```diff
@@ -0,0 +1,173 @@
...
+        self,
+        name: str,
+        session_tag: str,
+        user_tag: str,
+    ):
+        """Initialize session memory with index
+
+        Session Manager stores the current and previous user text prompts and
+        LLM responses to allow for enriching future prompts with session
+        context. Session history is stored in prompt:response pairs referred to
```
rbs333: I believe this was made more flexible to not just store in prompt:response pairs? Might just be update to comment.

redisvl/extensions/session_manager/base_session.py
```diff
@@ -0,0 +1,173 @@
...
+        Raises:
+            ValueError: If top_k is not an integer greater than or equal to 0.
+        """
+        raise NotImplementedError
+
+    def _format_context(
+        self, hits: List[Dict[str, Any]], as_text: bool
+    ) -> Union[List[str], List[Dict[str, str]]]:
+        """Extracts the prompt and response fields from the Redis hashes and
+           formats them as either flat dictionaries oor strings.
```
rbs333: oor

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -0,0 +1,365 @@
...
+                        "distance_metric": "cosine",
+                        "algorithm": "flat",
+                    },
+                },
+            ]
+        )
+
+        self._index = SearchIndex(schema=schema)
+
+        if redis_client:
```
rbs333, line 92: Not a blocker just thinking out loud. 

What are the advantages of passing a redis_client vs just a redis_url? If I'm a user I think I'd rather just provide the connection string and let the class do the rest. Then this code gets simplier because it's just redis_url or "redis://localhost:6379" 

I'm also wondering if we should overwrite by default in case of collision issues... but could go either way.

redisvl/extensions/session_manager/standard_session.py
```diff
@@ -0,0 +1,193 @@
...
+        The proposed schema will support a single combined vector embedding
+        constructed from the prompt & response in a single string.
+
+        """
+        super().__init__(name, session_tag, user_tag)
+
+        if redis_client:
+            self._client = redis_client
+        else:
+            # TODO make this configurable
```
rbs333: kinda related to other comment

rbs333 (commented):
Looking good left a few thoughts. Since this PR is pretty large and involved it might be worth using 15 minutes at the end of standup to make sure we all really get it before merge otherwise probably pretty much there

tylerhutcherson (approved):
Great work -- pulling this into main and we can scope a few updates over the next few weeks before our next release.

PR #145, opened 2024-04-30
Robert/build/tests

rbs333:
Notes:

- adding changes for poetry build cycle and push to test-pypi
- updates test_cov script to properly fail with failed tests

todo before merge:
- remove push branch option
- replace pip install with poetry
- grok test-pypi cache problem with version


pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [tool.poetry]
-name = "redisvl"
+name = "redisvl-test"
```
tylerhutcherson: just marking so we revert

scripts.py
```diff
@@ -20,13 +20,13 @@ def mypy():
     subprocess.run(["python", "-m", "mypy", "./redisvl"])
 
 def test():
-    subprocess.run(["python", "-m", "pytest", "--log-level=CRITICAL"])
+    subprocess.run(["python", "-m", "pytest", "--log-level=CRITICAL"], check=True)
 
 def test_verbose():
     subprocess.run(["python", "-m", "pytest", "-vv", "-s", "--log-level=CRITICAL"])
```
tylerhutcherson: let's go ahead and add check=True to all

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+  #     - name: Install Poetry
+  #       uses: snok/install-poetry@v1
+
+  #     - uses: actions/download-artifact@v4
+  #       with:
+  #         name: dist
+  #         path: dist/
+
+  #     - name: Publish to PyPI
+  #       env:
```
tylerhutcherson, line 62: Will need to review this step to make sure we follow a similar flow from test pypi

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+        uses: actions/setup-python@v4
+        with:
+          python-version: ${{ env.PYTHON_VERSION }}
+
+      - name: Install Poetry
+        uses: snok/install-poetry@v1
+
+      - name: Install dependencies
+        run: |
+          poetry install --all-extras
```
tylerhutcherson: we needed this step afterall?

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
...
+
+      - uses: actions/download-artifact@v4
+        with:
+          name: dist
+          path: dist/
+
+      - name: Publish to TestPyPI
+        env:
+          POETRY_PYPI_TOKEN_TESTPYPI: ${{ secrets.TESTPYPI }}
+        run: poetry config repositories.test-pypi https://test.pypi.org/legacy/; poetry config pypi-token.test-pypi $POETRY_PYPI_TOKEN_TESTPYPI; poetry publish --repository test-pypi
```
tylerhutcherson: it dawns on me that if I can't get access to the official `redisvl` test-pypi index, we will have to comment out the pre-checks here because right now this leans on the custom `redisvl-test` package which we won't be able to support. Although maybe there is a setting for this in the pyproject.toml

.github/workflows/release.yml
```diff
@@ -0,0 +1,145 @@
+name: Release
+
+on:
+  # TODO: remove
+  push:
+    branches: [ robert/build/tests ]
```
tylerhutcherson: yep just marking so we remember to remove this :)

tylerhutcherson (changes requested):

.github/workflows/release.yml
rbs333: we had this as a todo

.github/workflows/release.yml
rbs333: yeah we're unfortunately still reliant on Sam unless we can get that oicd work around to happen

.github/workflows/release.yml
rbs333: image

So dropping this we get no model named pytest when we try and run tests.

.github/workflows/release.yml
rbs333: Yeah I'm not sure if we'll have to do the config step stuff with this version since I think that was partially just to get it off the default but would have to test

tylerhutcherson (approved):

PR #148, opened 2024-05-11
feat: adds optional sorting parameter to FilterQuery

justin-cechmanek:
This PR adds the ability to sort results by a hash field if the field name is passed as a parameter to filter queries.

### DIFF
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -136,6 +136,7 @@ class FilterQuery(BaseQuery):
         return_fields: Optional[List[str]] = None,
         num_results: int = 10,
         dialect: int = 2,
+        sort_by: Optional[str] = None,
         params: Optional[Dict[str, Any]] = None,
     ):
         """A query for a running a filtered search with a filter expression.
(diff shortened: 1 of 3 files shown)


redisvl/query/query.py
```diff
@@ -136,6 +136,7 @@ def __init__(
         return_fields: Optional[List[str]] = None,
         num_results: int = 10,
         dialect: int = 2,
+        sort_by: Optional[str] = None,
```
tylerhutcherson, line 139: let's also add this support to vector queries too

tylerhutcherson (changes requested):
let's add to vector queries as well since you can sort the resulting set by any given field in the index

redisvl/query/query.py
justin-cechmanek: Added to Vector and Range queries also

tylerhutcherson: thanks!

tylerhutcherson (approved):

PR #150, opened 2024-05-13
Reranker implementation using Cross Encoders (HuggingFace / SentenceT…

bsbodden:
…ransformers)

Inspired by [LangChain's Cross Encoder Reranker](https://python.langchain.com/v0.1/docs/integrations/document_transformers/cross_encoder_reranker/)

### DIFF
--- a/docs/user_guide/rerankers_06.ipynb
+++ b/docs/user_guide/rerankers_06.ipynb
@@ -9,7 +9,10 @@
     "\n",
     "In this notebook, we will show how to use RedisVL to rerank search results\n",
     "(documents or chunks or records) based on the input query. Today RedisVL\n",
-    "supports reranking through the [Cohere /rerank API](https://docs.cohere.com/docs/rerank-2).\n",
+    "supports reranking through: \n",
+    "\n",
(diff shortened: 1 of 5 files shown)


redisvl/utils/rerank/cross_encoder.py
```diff
@@ -0,0 +1,129 @@
+from typing import Any, Dict, List, Optional, Tuple, Union
+
+from sentence_transformers import CrossEncoder
+
+from redisvl.utils.rerank.base import BaseReranker
+
+
+class CrossEncoderReranker(BaseReranker):
```
tylerhutcherson: Given the pattern of vectorizers and rerankers thus far, I think we should use `HFReranker` or somehow combine HF and CrossEncoder in the name. The names of these classes have been indicating who the integration is with and the "model" param gives the ability to select the "implementation". So I think as long as we call out HF, we should be good since it will require those dependencies.

tylerhutcherson (commented):
LGTM! Left one suggestion around naming :)

redisvl/utils/rerank/cross_encoder.py
bsbodden: `HFCrossEncoderReranker` or fully spelled out `HuggingFaceCrossEncoderReranker`?

bsbodden: Done!

tylerhutcherson (approved):
Love it!

PR #151, opened 2024-05-13
Fix: Issue with search dialect 3 and JSON (resolves #140)

### LINKED ISSUE #140: Issue with search dialect 3 and JSON
The dialect 3 api changes the outputs for JSON:

```python
q = VectorQuery(
    vector=[0.23, 0.12, -0.03, 0.98],
    vector_field_name="embedding",
    filter_expression=filter_expression,
    return_fields=["name", "description", "price"],
    dialect=3
)
index.query(q)
```
returns:
```
[{'id': 'product:36edbfd1372144759975f01fe6968bbf',
  'vector_distance': '1.08839428425',
  'name': '["Wireless earbuds"]',
  'description': '["Wireless Bluetooth in-ear headphones"]',
  'price': '[64.99]'}]
```

BUT dialect 2 returns:
```
[{'id': 'product:36edbfd1372144759975f01fe6968bbf',
  'vector_distance': '1.08839428425',
  'name': 'Wireless earbuds',
  'description': 'Wireless Bluetooth in-ear headphones',
  'price': '64.99'}]
```
(which is the correct and user expected format.

Additionally `FilterQuery` types break when using dialect=3... 

Need to fix the parsing layer in RedisVL.

bsbodden:
Attempt at addressing issue #140

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -32,6 +32,38 @@ from redisvl.utils.log import get_logger
 logger = get_logger(__name__)
 
 
+def _handle_dialect_3(result: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Handle dialect 3 responses by converting JSON-encoded list values to strings.
+
+    Each JSON-encoded string in the result that is a list will be converted:
+    - If the list has one item, it is unpacked.
+    - If the list has multiple items, they are joined into a single comma-separated string.
+
+    Args:
+        result (Dict[str, Any]): The dictionary containing the results to process.
+
+    Returns:
+        Dict[str, Any]: The processed dictionary with updated values.
+    """
+    for field, value in result.items():
+        if isinstance(value, str):
+            try:
+                parsed_value = json.loads(value)
+            except json.JSONDecodeError:
+                continue  # Skip processing if value is not valid JSON
+
+            if isinstance(parsed_value, list):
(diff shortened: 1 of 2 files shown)


redisvl/index/index.py
```diff
@@ -32,6 +32,38 @@
 logger = get_logger(__name__)
 
 
+def _handle_dialect_3(result: Dict[str, Any]) -> Dict[str, Any]:
+    """
+    Handle dialect 3 responses by converting JSON-encoded list values to strings.
+
+    Each JSON-encoded string in the result that is a list will be converted:
+    - If the list has one item, it is unpacked.
+    - If the list has multiple items, they are joined into a single comma-separated string.
```
rbs333, line 41: It seems like this is the desired behavior from @tylerhutcherson but I feel like I'd rather have a list than a comma-separated string in this situation. I may be missing some deeper context though

rbs333 (approved):
Left one style question but works as described when I run local

redisvl/index/index.py
bsbodden: @tylerhutcherson what do you think?

bsbodden:
@tylerhutcherson is this still needed?

tylerhutcherson:
Dialect 2 is becoming the standard for search.

PR #155, opened 2024-06-04
Add client info reporting

tylerhutcherson:
We need to be able to track the usage of clients for measuring adoption and impact. The `CLIENT SETINFO` command can be used for this purpose. The required string format is documented [here](https://redis.io/docs/latest/commands/client-setinfo/#:~:text=The%20CLIENT%20SETINFO%20command%20assigns,CLIENT%20LIST%20and%20CLIENT%20INFO%20).

### Variants

- Standalone RedisVL:
```
CLIENT SETINFO LIB-NAME redis-py(redisvl_v0.2.0)
CLIENT SETINFO LIB-VER 5.0.4
```

- Abstraction layers (like LlamaIndex or LangChain):
```
CLIENT SETINFO LIB-NAME redis-py(redisvl_v0.2.0;llama-index-vector-stores-redis_v0.1.0)
CLIENT SETINFO LIB-VER 5.0.4
```

### Constraints
- RedisVL uses both Async connection and standard connection instances from Redis Py
	- So the technique to run this command needs to properly handle this...
- Other wrappers around RedisVL like LangChain and LlamaIndex will need to pass through their lib_name
- These libraries generally support the notion of providing your own client instance OR providing the connection string and performing the connection on your behalf -- which also adds some difficulty.

### Planned Route

In order to build the proper name string, all clients that wrap redis-py will need to use the following format:
```
{package-name}_v{version}
```

RedisVL will implement the default which is `redisvl_v0.x.x`, but outer wrappers of RedisVL can implement their own by using a lib_name `kwarg` on the index class like 

```
SearchIndex(lib_name="langchain_v0.2.1")
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -38,62 +38,62 @@ jobs:
           name: dist
           path: dist/
 
-  test-pypi-publish:
-    needs: build
-    runs-on: ubuntu-latest
+  # test-pypi-publish:
+  #   needs: build
+  #   runs-on: ubuntu-latest
 
-    steps:
-      - uses: actions/checkout@v4
+  #   steps:
+  #     - uses: actions/checkout@v4
 
-      - name: Set up Python
-        uses: actions/setup-python@v4
-        with:
-          python-version: ${{ env.PYTHON_VERSION }}
+  #     - name: Set up Python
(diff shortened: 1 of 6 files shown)


bsbodden (dismissed):
This seems to fit the bill! LGTM

PR #177, opened 2024-07-02
adds check to ensure correct embedding vector dimensions are used

justin-cechmanek:
Currently our semantic cache allows for specifying the vector in calls to store() and check(), but if the vector dimension does not match the schema dimensions this fails silently. This PR adds a check to verify correct vector dimensions and raises an error if they do not match.


bsbodden (approved):
LGTM!

tylerhutcherson (approved):

PR #178, opened 2024-07-08
Improve caching ttl tests

tylerhutcherson:
Small improvement to the TTL tests for semantic cache. Based on feedback from an SA in the field.


justin-cechmanek (approved):
Looks good

justin-cechmanek (changes requested):
test_ttl_expiration_after_update is still finding a match

tylerhutcherson:
> test_ttl_expiration_after_update is still finding a match

Figured out the issue. I wrote the test incorrectly because of the refresh after hit policy that's in place by default.

PR #179, opened 2024-07-08
Add ability to clear ALL data associated with an index

tylerhutcherson:
This PR introduces the `clear()` method to the core SearchEngine class in RedisVL. This now allows for the ability to clear out all data, while leaving the index in place.

Useful for manual cache eviction, manual session clearing, and more.

This PR also updates the extension classes to use the new `clear()` method as opposed to the SCAN ITER approach.

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -492,8 +492,8 @@ class SearchIndex(BaseSearchIndex):
             FilterQuery(FilterExpression("*"), return_fields=["id"]), page_size=500
         ):
             batch_keys = [record["id"] for record in batch]
-            self._redis_client.delete(*batch_keys)  # type: ignore
-            total_records_deleted += len(batch_keys)
+            record_deleted = self._redis_client.delete(*batch_keys)  # type: ignore
+            total_records_deleted += record_deleted  # type: ignore
 
         return total_records_deleted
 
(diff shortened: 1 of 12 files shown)


redisvl/index/index.py
bsbodden, line 1: Maybe we should let them control the batching parameter but defaulted to 500?

redisvl/index/index.py
```diff
@@ -476,6 +477,26 @@ def delete(self, drop: bool = True):
...
+
+        Returns:
+            int: Count of records deleted from Redis.
+        """
+        # Track deleted records
+        total_records_deleted: int = 0
+
+        # Paginate using queries and delete in batches
+        for batch in self.paginate(
+            FilterQuery(FilterExpression("*"), return_fields=["id"]), page_size=500
```
bsbodden, line 492: Unlikely scenario, but what if I want to get the (some) data out before destroying the index, e.g. `return_fields` being configurable also?

bsbodden (commented):
LGTM, left a couple of comments

redisvl/index/index.py
tylerhutcherson: I agree with the notion that someone might want to be able to export data from an index. However, I think that should be it's own clear feature. For example, an `index.export(file_path="file.json")` method or something similar?

I think `clear()` should just be a simple method for clearing data out of your index

redisvl/index/index.py
tylerhutcherson: I think if someone's goal is to just clear all of the data from an index, we should implement the best practice under the hood and not make the user worry about it? But it certainly woudln't be hard to add an optional arg in the future if folks need it in the future!

PR #180, opened 2024-07-10
adds semantic cache scoped access and additional functionality

justin-cechmanek:
This PR is the first of a few PRs that adds to RedisVL’s existing semantic cache class more functionality around dropping and updating cache entries. It also adds scoped access control and meta data fields.


redisvl/extensions/llmcache/semantic.py
```diff
@@ -183,6 +191,20 @@ def delete(self) -> None:
         index."""
         self._index.delete(drop=True)
 
+    def drop(self, document_ids: Union[str, List[str]]) -> None:
+        """Remove a specific entry or entries from the cache by it's ID.
+
+        Args:
+            document_ids (Union[str, List[str]]): The document ID or IDs to remove from the cache.
+        """
+        if isinstance(document_ids, List):
```
tylerhutcherson: In a similar theme to the work around `clear()`, to me this feels like an addition to that method by supporting an optional list of ids? by default clear removes everything. Maybe if this list is provided, instead of clearing everything it will just focus on using this pipeline code to clear the subset?

redisvl/extensions/llmcache/semantic.py
```diff
@@ -16,6 +18,9 @@ class SemanticCache(BaseLLMCache):
     entry_id_field_name: str = "id"
     prompt_field_name: str = "prompt"
     vector_field_name: str = "prompt_vector"
+    inserted_at_field_name: str = "inserted_at"
+    updated_at_field_name: str = "updated_at"
+    tag_field_name: str = "scope_tag"
```
tylerhutcherson: Just want to understand this data model a bit more. As it stands here, it looks like there would be 1 field for scoping (`scope_tag`) and filtering. What if I have both year, location, and user ID data that I'd like to filter my cache check on?

I think we might need to generalize a bit more and support a set of optional/customizable filterable fields (providable to the class on init??)

So if staying with Hash, the schema could be something like:
- `id`
- `prompt`
- `prompt_vector`
- `response`
- `inserted_at`
- `updated_at`
- `metadata`
- `scope_tag`???
- <all other fields as customizable filter fields included in the cache schema>?

Just thinking out loud!

redisvl/extensions/llmcache/semantic.py
```diff
@@ -77,6 +82,9 @@ def __init__(
             [
                 {"name": self.prompt_field_name, "type": "text"},
                 {"name": self.response_field_name, "type": "text"},
+                {"name": self.inserted_at_field_name, "type": "numeric"},
```
tylerhutcherson: Checkout how I do the schema encapsulation / wrapping in my routing branch. I think it's the strategy we will want to move towards with all of our interfaces. I am happy to put this in a commit so you can see what it would look like!

redisvl/extensions/llmcache/semantic.py
```diff
@@ -248,6 +276,7 @@ def check(
         vector: Optional[List[float]] = None,
         num_results: int = 1,
         return_fields: Optional[List[str]] = None,
+        tags: Optional[Union[List[str], str]] = None,
```
tylerhutcherson: I like the idea here of allowing for these tags to filter. Wonder if we could lean on the more generic concept and allow user to pass a `FilterExpression` object which then gets applied to the `RangeQuery` or `VectorQuery`? This would align with the notion of letting users customize their filters broadly and use the redisvl tools to shape the filter at run time!

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: I like the idea of combining methods. What about something like 
```
clear(keys: Optional[List[str], str]):
"""Remove all or specific entries from the cache by it's ID.
    Args:
        document_ids (Union[str, List[str]]): The document ID or IDs to remove from the cache. If none are provided then all documents are removed.
"""
```

justin-cechmanek: Will move this pipeline code into index.

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: I like that idea, especially since we already have good support and examples of how to construct filters.

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: You mean your `SemanticRouterIndexSchema` class with `from_params()` method?
We can bring that over here

redisvl/extensions/llmcache/semantic.py
```diff
@@ -333,19 +366,66 @@ def store(
         self._check_vector_dims(vector)
 
         # Construct semantic cache payload
+        now = time()
```
tylerhutcherson: We should use `datetime.utcnow()` to generate a UTC tz and then use the numeric version of that. Let's also make this a utility function that can be imported from `redisvl.utils` module. We will use this for multiple things across extensions.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -333,19 +366,66 @@ def store(
...
             self.response_field_name: response,
             self.vector_field_name: array_to_buffer(vector),
+            self.inserted_at_field_name: now,
+            self.updated_at_field_name: now,
         }
         if metadata is not None:
             if not isinstance(metadata, dict):
                 raise TypeError("If specified, cached metadata must be a dictionary.")
             # Serialize the metadata dict and add to cache payload
             payload[self.metadata_field_name] = self.serialize(metadata)
```
tylerhutcherson: Ditto, I think we can move serialization to/from JSON string to utils so we can use the same technique and unit test the module.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -333,19 +366,66 @@ def store(
...
+
+            # Check for metadata and deserialize
+            if _key == self.metadata_field_name:
+                if isinstance(val, dict):
+                    kwargs[_key] = self.serialize(val)
+                else:
+                    raise TypeError(
+                        "If specified, cached metadata must be a dictionary."
+                    )
+        kwargs.update({self.updated_at_field_name: time()})
```
tylerhutcherson: replace with the time helper method

redisvl/extensions/llmcache/semantic.py
```diff
@@ -333,19 +366,66 @@ def store(
...
+        self._index.client.hset(key, mapping=kwargs)  # type: ignore
```
tylerhutcherson, line 430: Do we refresh TTL here too?

redisvl/extensions/llmcache/semantic.py
```diff
@@ -1,21 +1,54 @@
...
 
+class SemanticCacheIndexSchema(IndexSchema):
+
+    @classmethod
+    def from_params(cls, name: str, vector_dims: int):
+
+        return cls(
+            index={"name": name, "prefix": name},  # type: ignore
+            fields=[  # type: ignore
+                {"name": "cache_name", "type": "tag"},
```
tylerhutcherson: Is there a reason we need to index a `cache_name` field? Doesn't seem like it's used.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -1,21 +1,54 @@
...
+                {"name": "prompt", "type": "text"},
+                {"name": "response", "type": "text"},
+                {"name": "inserted_at", "type": "numeric"},
+                {"name": "updated_at", "type": "numeric"},
+                {"name": "label", "type": "tag"},
```
tylerhutcherson, line 27: Just to confirm -- `label` field intent here is to enable arbitrary filtering on label-like tags? I like the simplicity.

What about other forms of filtering though? Is there a plan to be able to support more generic metadata indexing and filtering on a SemanticCache?

justin-cechmanek: With the adding of FilterExpressions we can do simple filters on labels, and also complex combined text or range filters. I added some examples in the tests.

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: good catch. We should

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: It's not used. Removed it.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -182,6 +194,14 @@ def delete(self) -> None:
         index."""
         self._index.delete(drop=True)
 
+    def drop(self, document_ids: Union[str, List[str]]) -> None:
```
tylerhutcherson, line 197: So for this, are we expecting users to pass the full redis key? or just the id portion (without prefix)? I think the terminology we try to use for this throughout the library is `id` when we are referring to the part without the prefix. So maybe we just use `ids` instead of `document_ids`?

tylerhutcherson (approved):
I left a comment, but I will address it later in my PR this evening!

PR #181, opened 2024-07-10
resets cache ttl when no arg is passed

justin-cechmanek:
This PR resets the cache entry TTL to infinite/no expiration if we call `set_ttl()` with no arguments. Otherwise once a TTL is set, it can only be changed to another integer.


tylerhutcherson (approved):
lgtm

PR #183, opened 2024-07-10
adds drop_keys and async drop_keys methods to index

justin-cechmanek:
From our chat today I'm pulling out this method from the semantic cache and moving it into index. Rather can modify the `clear()` method to take an optional list of keys it's a new method.

The goal is to stop the accidental calling of clear() with no args, thinking it will clear no keys, but actually delete everything.

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -497,6 +497,19 @@ class SearchIndex(BaseSearchIndex):
 
         return total_records_deleted
 
+    def drop_keys(self, keys: Union[str, List[str]]) -> None:
+        """Remove a specific entry or entries from the index by it's key ID.
+        Args:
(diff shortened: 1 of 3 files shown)


rbs333 (dismissed):

redisvl/index/index.py
```diff
@@ -935,6 +945,16 @@ async def clear(self) -> int:
 
         return total_records_deleted
 
+    async def drop_keys(self, keys: Union[str, List[str]]) -> None:
```
tylerhutcherson: Sorry one quick thing. `delete` returns the count of keys dropped. With the `clear` method, I though it might be valuable to return the count here. Should be easy to do the same

tylerhutcherson (approved):
Looks great -- one small thought about returning the count

tylerhutcherson (approved):

PR #187, opened 2024-07-24
Enhance semantic router loading

tylerhutcherson:
This PR introduces some serialization methods for the `SemanticRouter` so that it can be loaded to and from file or dict.


rbs333 (approved):
LGTM. Some nice ergonomic adds with yaml and dict helpers

PR #188, opened 2024-07-26
Standardize extensions redis init

tylerhutcherson:
Standardizes the way our extension modules connect to redis.

- First check for client
- Then check for URL
- Then raise a CLEAR error if neither are provided

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -27,7 +27,7 @@ class SemanticCache(BaseLLMCache):
         ttl: Optional[int] = None,
         vectorizer: Optional[BaseVectorizer] = None,
         redis_client: Optional[Redis] = None,
-        redis_url: Optional[str] = None,
+        redis_url: str = "redis://localhost:6379",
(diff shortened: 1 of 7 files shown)


tylerhutcherson:
(I need to fix the failing test on the notebook)

justin-cechmanek:
Looks good. Just curious why we don't want to have a default redis url anymore

tylerhutcherson:
> Looks good. Just curious why we don't want to have a default redis url anymore

Certainly open to discussing this with everyone more. Let me ping the rest of the team to see what the collective thoughts are. My preference is for these to be less magicy and focus on clarity. But if it'a a common of enough pattern, we can go for it.

tylerhutcherson:
> > Looks good. Just curious why we don't want to have a default redis url anymore
> 
> Certainly open to discussing this with everyone more. Let me ping the rest of the team to see what the collective thoughts are. My preference is for these to be less magicy and focus on clarity. But if it'a a common of enough pattern, we can go for it.

After reviewing with team, we went back to defaulting to localhost Redis.

justin-cechmanek (approved):
Looks good!

PR #190, opened 2024-07-29
Support inorder query param

tylerhutcherson:


justin-cechmanek:
More of a question, and not a critique: what are your opinions on defaulting to `None` for the in_order bool instead of `False`? Looks like the logic and behavior is identical either way.

tylerhutcherson:
> More of a question, and not a critique: what are your opinions on defaulting to `None` for the in_order bool instead of `False`? Looks like the logic and behavior is identical either way.

You're exactly right -- oversight by me. Updated it in the latest commit.

justin-cechmanek (approved):
LGTM

PR #192, opened 2024-07-30
Refactor search index to improve connection handling

tylerhutcherson:
We were leaning on a hack to run some async code in a sync setting. This was both dangerous and an anti-pattern. We also needed to refactor some of the shared and non-shared content between the BaseSearchIndex and derivatives.


rbs333 (approved):
Looks like improvement to me. Key takeaway to use the new decorated .set_client

tylerhutcherson:
> Looks like improvement to me. Key takeaway to use the new decorated .set_client

Yes, for `AsyncSearchIndex`. Because we have to run a bit of async code to properly, and safely set it up, that's the correct way rather than the initial hack we had in place.

PR #193, opened 2024-07-30
Makes setting scope fully optional in session managers

justin-cechmanek:
This PR removes the requirement that user and session tags be specified on session manager initialization. Session tags can be added when storing messages and filter expressions can be used to retrieve specific chat histories. A default uuid is used when no session is provided.

### DIFF
--- a/redisvl/extensions/session_manager/base_session.py
+++ b/redisvl/extensions/session_manager/base_session.py
@@ -7,7 +7,7 @@ from redisvl.query.filter import FilterExpression
 
 
 class BaseSessionManager:
-    id_field_name: str = "id_field"
+    id_field_name: str = "_id"
     role_field_name: str = "role"
     content_field_name: str = "content"
     tool_field_name: str = "tool_call_id"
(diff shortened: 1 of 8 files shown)


redisvl/extensions/session_manager/base_session.py
```diff
@@ -75,23 +59,21 @@ def messages(self) -> Union[List[str], List[Dict[str, str]]]:
     def get_recent(
         self,
         top_k: int = 5,
-        session_tag: Optional[str] = None,
-        user_tag: Optional[str] = None,
         as_text: bool = False,
         raw: bool = False,
+        tag_filter: Optional[FilterExpression] = None,
```
tylerhutcherson: my concern here is we blur the conceptual lines between a `tag_filter` and a `session_tag`. Also we "own" the schema. Users wouldn't know they are supposed to use this field for session filtering. Ideally we make it so it's super clear :)

We could make it clear and ask them to pass the `session_tag: Optional[str] = None` here and WE wrap it into a Tag filter since we own the schema and know how to apply it.

justin-cechmanek:
inside search_cache() what if instead of:
      id = hit[self.entry_id_field_name]
      self._refresh_ttl(self._index.key(id))
we did:
      id = hit['id']
      self._refresh_ttl(self._index.key(id))

tylerhutcherson (approved):

PR #194, opened 2024-07-31
Use ChatMessage pydantic model

tylerhutcherson:
Uses pydantic for validation, docs, and custom serialization of chat message objects on the way into and out of Redis.


justin-cechmanek (approved):
added a few simple doc strings to the utils, but otherwise it looks good!

PR #195, opened 2024-08-01
Use pydantic for cache entries and hits

tylerhutcherson:
This PR handles 2 things:

1) Uses pydantic for both `CacheEntry` and `CacheHit` models. It ended up being better and cleaner to have two separate models.... one for what we write, and the other for what we load.

2) Filtering. @justin-cechmanek I want your feedback on this. I landed on a technique to allow for the user to define a list of `filterable_fields` as part of the semantic cache class init. What this does is give us a way to support arbitrary filters (scope, permissions, tags, numerics.... anything) within reason. Then you can create any `FilterExpression` and pass through at query time. This extends your initial implementation.

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -3,7 +3,11 @@ from typing import Any, Dict, List, Optional, Union
 from redis import Redis
 
 from redisvl.extensions.llmcache.base import BaseLLMCache
-from redisvl.extensions.llmcache.schema import CacheEntry, CacheHit, SemanticCacheIndexSchema
+from redisvl.extensions.llmcache.schema import (
+    CacheEntry,
+    CacheHit,
+    SemanticCacheIndexSchema,
+)
 from redisvl.index import SearchIndex
 from redisvl.query import RangeQuery
 from redisvl.query.filter import FilterExpression, Tag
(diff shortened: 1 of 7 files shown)


redisvl/extensions/llmcache/semantic.py
```diff
@@ -68,9 +49,6 @@ def __init__(
         Args:
             name (str, optional): The name of the semantic cache search index.
                 Defaults to "llmcache".
-            prefix (Optional[str], optional): The prefix for Redis keys
-                associated with the semantic cache search index. Defaults to
-                None, and the index name will be used as the key prefix.
```
justin-cechmanek, line 73: Add a description for filterable_fields?

And I don't think this raises a Value error if the index name is not provided, so we can drop that comment line.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -101,25 +81,42 @@ def __init__(
...
+        if filterable_fields is not None:
+            for filter_field in filterable_fields:
+                if (
+                    filter_field["name"] in self.return_fields
+                    or filter_field["name"] == "key"
+                ):
+                    raise ValueError(
+                        f'{filter_field["name"]} is a reserved field name for the semantic cache schema'
+                    )
+                schema.add_field(filter_field)
```
justin-cechmanek: I really like the idea of allowing fully customizable fields for people to add and filter on!

justin-cechmanek (approved):
Looks great! Just one note about updating the doc string

PR #199, opened 2024-08-05
rename files for order

rbs333:
It bothered me how these were numbered but wouldn't show up in order 
image

I also changed the name from llmcache to semantic_cache


tylerhutcherson:
@rbs333 Also need to update the sidebar javascript that references these filenames in the docs

PR #201, opened 2024-08-12
Fix semantic cache with empty metadata

tylerhutcherson:
We found a bug where `metadata={}` was not getting handled correctly.

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -300,8 +300,6 @@ class SemanticCache(BaseLLMCache):
             key = cache_search_result["id"]
             self._refresh_ttl(key)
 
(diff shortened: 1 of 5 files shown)


redisvl/extensions/llmcache/semantic.py
```diff
@@ -300,7 +300,7 @@ def check(
             key = cache_search_result["id"]
             self._refresh_ttl(key)
 
-            print(cache_search_result, flush=True)
+            # print(cache_search_result, flush=True)
```
justin-cechmanek: I think we can just remove this line entirely.

justin-cechmanek (changes requested):
Looks good! Just the one comment we can trim

rbs333 (approved):

PR #202, opened 2024-08-12
Support distance threshold override

tylerhutcherson:
At runtime, we should extend the ability to override the distance threshold if provided. This is also how the semantic router works. Parity between the extensions is key here (in terms of functionality)


justin-cechmanek (approved):
I like it

redisvl/extensions/llmcache/semantic.py
```diff
@@ -300,7 +306,7 @@ def check(
             key = cache_search_result["id"]
             self._refresh_ttl(key)
 
-            print(cache_search_result, flush=True)
+            # print(cache_search_result, flush=True)
```
rbs333: hanging comment

rbs333 (commented):
One small thing and I think there's now a validation error from the pydantic class in the test

PR #205, opened 2024-08-16
Fix notebook CI

tylerhutcherson:
Because our notebooks are run in parallel using threads, for expedited testing purposes, and using a single redis instance, there were situations where different caches were created with the same name... causing collisions.


bsbodden (approved):
LGTM!

PR #211, opened 2024-08-30
Refactor query classes

tylerhutcherson:
In order to support more advanced query params/settings of the Redis query engine, we needed a more flexible and customizable way to build RedisVL queries.

This PR introduces a simple refactor to inherit all redis-py Query module capabilities for each query type.

Source: https://github.com/redis/redis-py/blob/master/redis/commands/search/query.py

Updated user guide is still in flight **

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -296,13 +296,8 @@ class SemanticCache(BaseLLMCache):
 
         # overrides
         distance_threshold = distance_threshold or self._distance_threshold
-        return_fields = return_fields or self.return_fields
         vector = vector or self._vectorize_prompt(prompt)
-
         self._check_vector_dims(vector)
-
-        if not isinstance(return_fields, list):
-            raise TypeError("return_fields must be a list of field names")
 
         query = RangeQuery(
             vector=vector,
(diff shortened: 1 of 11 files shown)


redisvl/extensions/llmcache/semantic.py
```diff
@@ -320,15 +315,18 @@ def check(
         cache_search_results = self._index.query(query)
 
         for cache_search_result in cache_search_results:
-            key = cache_search_result["id"]
-            self._refresh_ttl(key)
+            redis_key = cache_search_result.pop("id")
```
rbs333, line 318: is this related of just a bonus part of pr?

redisvl/query/query.py
```diff
@@ -1,234 +1,179 @@
...
-        self._in_order = in_order
+class BaseQuery(RedisQuery):
+    """Base query class used to subclass many query types."""
+
+    _params: Dict[str, Any] = {}
+    _filter_expression: FilterExpression = FilterExpression("*")
+
+    def __init__(self, query_string: str = "*"):
+        """
+        Initialize the BaseQuery class.
```
rbs333, line 18: love how compact this gets

redisvl/query/query.py
```diff
@@ -1,234 +1,179 @@
...
+            self._params = params
 
-        .. code-block:: python
+        self._num_results = num_results
 
-            from redisvl.query import CountQuery
-            from redisvl.query.filter import Tag
+        # Initialize the base query with the full query string constructed from the filter expression
+        query_string = self._build_query_string()
+        super().__init__(query_string)
```
rbs333, line 106: we should maybe call out somewhere that we are extending the RedisQuery now which is why it is like this

rbs333 (approved):
LGTM left a couple comments if we want to make a couple things more clear

redisvl/extensions/llmcache/semantic.py
tylerhutcherson: I found a small hidden bug that broke my tests after the query refactor work. The new query refactor work changes the way the return_fields are generated/stored on the class and that exposed an issue in how the cache checks are done

rbs333 (dismissed):

rbs333 (approved):

PR #214, opened 2024-09-06
Add async support to semantic cache extension

tylerhutcherson:
This PR introduces new async compliant methods to the semantic cache class using lazy index construction. Because the `AsyncSearchIndex` requires an async redis python client, we needed to construct that class lazily upon first usage within the semantic cache class. This PR fixes some unclosed connection errors and is also in support of https://github.com/BerriAI/litellm/pull/5412 at LiteLLM.

### DIFF
--- a/redisvl/utils/vectorize/base.py
+++ b/redisvl/utils/vectorize/base.py
@@ -61,6 +61,7 @@ class BaseVectorizer(BaseModel, ABC):
         as_buffer: bool = False,
         **kwargs,
     ) -> List[List[float]]:
+        # Fallback to standard embedding call if no async support
         return self.embed_many(texts, preprocess, batch_size, as_buffer, **kwargs)
 
     async def aembed(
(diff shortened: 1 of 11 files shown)


redisvl/utils/vectorize/base.py
```diff
@@ -62,17 +61,16 @@ async def aembed_many(
         as_buffer: bool = False,
         **kwargs,
     ) -> List[List[float]]:
-        raise NotImplementedError
+        return self.embed_many(texts, preprocess, batch_size, as_buffer, **kwargs)
```
tylerhutcherson, line 65:
```suggestion
        # Fallback to standard embedding call if no async support
        return self.embed_many(texts, preprocess, batch_size, as_buffer, **kwargs)
```

redisvl/utils/vectorize/base.py
```diff
@@ -62,17 +61,16 @@ async def aembed_many(
...
-    @abstractmethod
     async def aembed(
         self,
         text: str,
         preprocess: Optional[Callable] = None,
         as_buffer: bool = False,
         **kwargs,
     ) -> List[float]:
-        raise NotImplementedError
+        return self.embed(text, preprocess, as_buffer, **kwargs)
```
tylerhutcherson, line 75:
```suggestion
        # Fallback to standard embedding call if no async support
        return self.embed(text, preprocess, as_buffer, **kwargs)
```

rbs333 (approved):
good stuff! Nice solve

rbs333 (approved):

PR #216, opened 2024-09-09
Update cache check logic

tylerhutcherson:
Refactors the cache check logic into a single helper method. Adjusts the cache check and processing logic to better handle `return_fields` configurations.


bsbodden (approved):
LGTM!

PR #221, opened 2024-09-17
DOC-4198: add TCEs to the vector query page

dwdougherty:
Tabbed code examples (TCEs) for the vector query page.


doctests/query_vector.py
```diff
@@ -0,0 +1,68 @@
...
+res = index.query(vquery)
+print(res) # >>> [{'id': 'bicycle:6bcb1bb4...', 'vector_distance': '0.399111807346', 'description': 'Kids want...
+# REMOVE_START
+assert len(res) == 2
+# REMOVE_END
+# STEP_END
+
+# REMOVE_START
+# destroy index and data
+r.ft("idx:bicycle").dropindex(delete_documents=True)
```
tylerhutcherson:
```suggestion
index.delete(drop=True)
```

tylerhutcherson (changes requested):

doctests/query_vector.py
dwdougherty: That works. I used the former in all the other search examples, so didn't think to use a more idiomatic statement.

tylerhutcherson (approved):
Great!

PR #222, opened 2024-09-20
adds support for new vector data types

justin-cechmanek:
Adds support for bfloat16, float16, and float64 data types.

Users can now specify the desired dtype for their schema when creating semantic caches, semantic routers, and semantic session managers.

When not specified data type still defaults to float32


pyproject.toml
```diff
@@ -31,6 +31,7 @@ sentence-transformers = { version = ">=2.2.2", optional = true }
 google-cloud-aiplatform = { version = ">=1.26", optional = true }
 cohere = { version = ">=4.44", optional = true }
 mistralai = { version = ">=0.2.0", optional = true }
+ml-dtypes = "^0.4.0"
```
tylerhutcherson: Can you move this above the optional dependencies?

redisvl/extensions/llmcache/schema.py
```diff
@@ -26,6 +26,8 @@ class CacheEntry(BaseModel):
     """Optional metadata stored on the cache entry"""
     filters: Optional[Dict[str, Any]] = Field(default=None)
     """Optional filter data stored on the cache entry for customizing retrieval"""
+    dtype: str
```
tylerhutcherson: This should be something that's passed to the serialization method `to_dict`. `CacheEntry` itself doesn't need to hold onto this state.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -103,7 +103,10 @@ def __init__(
         ]
 
         # Create semantic cache schema and index
-        schema = SemanticCacheIndexSchema.from_params(name, prefix, vectorizer.dims)
+        dtype = kwargs.get("dtype", "float32")
```
tylerhutcherson, line 107: I like that this is `kwargs` here. It's overridable, but not necessary recommended ;)

redisvl/extensions/router/semantic.py
```diff
@@ -40,6 +40,7 @@ class SemanticRouter(BaseModel):
     """The vectorizer used to embed route references."""
     routing_config: RoutingConfig = Field(default_factory=RoutingConfig)
     """Configuration for routing behavior."""
+    vector_field_name: str = "vector"
```
tylerhutcherson: Can we just use the reference to the `_index.schema` data in order to get to the field name here? Otherwise this will allow for a user to customize the vector field name, which we don't want

redisvl/extensions/session_manager/schema.py
```diff
@@ -24,6 +24,8 @@ class ChatMessage(BaseModel):
     """An optional identifier for a tool call associated with the message."""
     vector_field: Optional[List[float]] = Field(default=None)
     """The vector representation of the message content."""
+    dtype: Optional[str] = Field(default="float32")
```
tylerhutcherson: Same note as the other pydantic model about having this passed in by the extension class when needed

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -51,6 +52,7 @@ def __init__(
             redis_url (str, optional): The redis url. Defaults to redis://localhost:6379.
             connection_kwargs (Dict[str, Any]): The connection arguments
                 for the redis client. Defaults to empty {}.
+            dtype (str): The data type for the prompt vector. Defaults to "float32".
```
tylerhutcherson: Should this be in the docstring? if so, should be in all extension __init__ docstrings

redisvl/index/index.py
```diff
@@ -354,7 +354,7 @@ def from_existing(
 
         # Validate modules
         installed_modules = RedisConnectionFactory.get_modules(redis_client)
-        validate_modules(installed_modules, [{"name": "search", "ver": 20810}])
+        validate_modules(installed_modules, [{"name": "search", "ver": 21005}])
```
tylerhutcherson: Why do we need to validate the module version up to `21005` for use of the `from_existing` method? The ability to read the index info using ft.info was available as of `20810`

redisvl/schema/fields.py
```diff
@@ -26,6 +26,8 @@ class VectorDistanceMetric(str, Enum):
 
 
 class VectorDataType(str, Enum):
```
tylerhutcherson, line 28: Is there a way to combine this with the usage of the `VectorDataTypes` dictionary in utils?

redisvl/index/storage.py
```diff
@@ -393,18 +394,30 @@ class HashStorage(BaseStorage):
...
 
         Raises:
             TypeError: If the object is not a dictionary.
+            ValueError: if the vector byte string is not of correct datatype.
         """
         if not isinstance(obj, dict):
             raise TypeError("Object must be a dictionary.")
 
+        try:
+            byte_string = obj["prompt_vector"]
```
tylerhutcherson: specific to semantic caching?

tylerhutcherson: If we need for v1 we could remove this kind of validation logic for now

tylerhutcherson (changes requested):
Nice! A few things we can adjust, but looking pretty solid

redisvl/index/index.py
justin-cechmanek: blfoat16 and float16 support was added in [RediSearch 2.10.5](https://redis.io/docs/latest/operate/oss_and_stack/stack-with-enterprise/release-notes/redisearch/redisearch-2.10-release-notes/#:~:text=What%27s%20new%20in%202.10.5)

justin-cechmanek: Ah, I see that's not needed if just reading the index. I'll switch it back.

redisvl/index/storage.py
justin-cechmanek: That's leftover from an early attempt at adding dtypes. Removed it now.

redisvl/extensions/llmcache/schema.py
justin-cechmanek: Shouldn't dtype stay bound to wherever we have the vector, to avoid any potential conflicts between the byte string and dtype?

redisvl/extensions/router/semantic.py
justin-cechmanek: I would like to keep things somewhat uniform across our cache, router, and session classes. `SemanticCache` and `SemanticSessionManager` both have this field. Should we remove it there also, or keep things different?

tylerhutcherson: That's a good point, I hadn't looked there.  Issue is with pydantic here. Because this is a `BaseModel` it will allow a user to set the vector_field_name param to be some str, but will default to `vector`. We don't want users to customize this variable. So if you can find a way to achieve that with pydantic or other, go for it!

redisvl/extensions/llmcache/schema.py
tylerhutcherson: I'd like to think that our extension classes are responsible for the write/read of the data.. and thus elements in the CacheEntry don't need to know how they were stored. we can chat about this!

redisvl/extensions/session_manager/semantic_session.py
justin-cechmanek: I pulled it from the doc string as it's not a named arg, but we probably should have some way to let users know this is an option.
Maybe include it in the docs and/or notebook examples?

redisvl/schema/fields.py
justin-cechmanek: I removed `VectorDataTypes` dict entirely now. Just needed to import this into `utils` and do some string casting.

redisvl/redis/utils.py
```diff
@@ -30,14 +33,22 @@ def convert_bytes(data: Any) -> Any:
     return data
 
 
-def array_to_buffer(array: List[float], dtype: Any = np.float32) -> bytes:
+def array_to_buffer(array: List[float], dtype: str) -> bytes:
     """Convert a list of floats into a numpy byte string."""
-    return np.array(array).astype(dtype).tobytes()
+    if dtype.upper() not in {v.value for v in VectorDataType}:
```
tylerhutcherson: enums can be used to check membership --> `VectorDataType(dtype.upper())`

this will throw a `ValueError` if the value is not a member of the enum.

justin-cechmanek: This makes membership check cleaner. I still want to raise our own value error so we can list out the accepted dtypes as the default ValueError doesn't do that.

tylerhutcherson: Yep -- like this?
```python
try:
    VectorDataType(dtype.upper())
except:
    raise ValueError("custom error msg")
```

redisvl/extensions/llmcache/semantic.py
```diff
@@ -23,7 +25,6 @@ class SemanticCache(BaseLLMCache):
     entry_id_field_name: str = "entry_id"
```
tylerhutcherson: We should use the same pattern to define constant field names, beyond just the vector field.

redisvl/extensions/session_manager/schema.py
tylerhutcherson: flagging to update using the same technique in semantic router + semantic cache

redisvl/extensions/llmcache/semantic.py
justin-cechmanek: I'd like to pull out the changing of class fields to constants into a new PR. Should be a quick one.

justin-cechmanek: I'd like to move the changing of all these class params into a constants file to another PR. Should be a quick one.

tylerhutcherson (approved):

PR #224, opened 2024-09-25
adds filter fields to cache key name hash

justin-cechmanek:
This PR aims to prevent cache entries being overwritten when the exact same prompt string is used, but other cache entry data - like user id, or some other field - is different.

### DIFF
--- a/redisvl/redis/utils.py
+++ b/redisvl/redis/utils.py
@@ -43,6 +43,6 @@ def buffer_to_array(buffer: bytes, dtype: Any = np.float32) -> List[float]:
 def hashify(content: str, extras: Optional[Dict[str, Any]] = None) -> str:
     """Create a secure hash of some arbitrary input text and optional dictionary."""
     if extras:
-        extra_string = " ".join([str(k) + str(v) for k, v in extras.items()])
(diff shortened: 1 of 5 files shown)


redisvl/redis/utils.py
```diff
@@ -40,6 +40,9 @@ def buffer_to_array(buffer: bytes, dtype: Any = np.float32) -> List[float]:
     return np.frombuffer(buffer, dtype=dtype).tolist()
 
 
-def hashify(content: str) -> str:
-    """Create a secure hash of some arbitrary input text."""
+def hashify(content: str, extras: Optional[Dict[str, Any]] = None) -> str:
+    """Create a secure hash of some arbitrary input text and optional dictionary."""
+    if extras:
```
tylerhutcherson, line 45: Small preference here to keep `hashify` dumb and move the filter string processing logic for this to within the `generate_id` method and pass the final string to hashify.

tylerhutcherson (approved):
LGTM! one small note, but good to go when you're ready. Once merged, can you close out that issue and follow up internally too?

rbs333 (approved):

PR #225, opened 2024-09-26
moves extension field names into constants file

justin-cechmanek:
Currently our cache, session, and router classes and their corresponding schema have fields like “vector_field_name” and “timestamp” hard coded into them. Since they must match between class and schema and we don’t want users to modify them they should be moved to a separate constants.py file.

### DIFF
--- a/redisvl/extensions/session_manager/schema.py
+++ b/redisvl/extensions/session_manager/schema.py
@@ -2,6 +2,15 @@ from typing import Dict, List, Optional
 
 from pydantic.v1 import BaseModel, Field, root_validator
 
+from redisvl.extensions.constants import (
+    CONTENT_FIELD_NAME,
+    ID_FIELD_NAME,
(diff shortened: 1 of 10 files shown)


redisvl/extensions/session_manager/schema.py
```diff
@@ -31,18 +40,28 @@ class Config:
     @root_validator(pre=True)
     @classmethod
     def generate_id(cls, values):
-        if "timestamp" not in values:
-            values["timestamp"] = current_timestamp()
-        if "entry_id" not in values:
-            values["entry_id"] = f'{values["session_tag"]}:{values["timestamp"]}'
+        ###if "timestamp" not in values:
```
tylerhutcherson: Remove comments?

tylerhutcherson (approved):
Looks good! Just one small comment cleanup, then merge away!

Issue #226, opened 2024-10-01
Redis Modules Identification missing for "searchlight" version of search module


bsbodden:
When we do a check for capabilities we should also check for the module named "searchlight" which is present instead of "search"  in certain deployments
![image](https://github.com/user-attachments/assets/c5af710d-d503-42c3-a667-6f9c1114f5cf)

bsbodden:
We already have the constant for it:
```py
DEFAULT_REQUIRED_MODULES = [
    {"name": "search", "ver": 20600},
    {"name": "searchlight", "ver": 20600},
]
```
There are two call to `validate_modules` like:
```py
validate_modules(installed_modules, [{"name": "search", "ver": 20810}])
```
My concern is the version number in the defaults being `20600` versus the `20810` being used in `index.py`

Potentially I simple fix would be to have a way to do an OR in, e.g. passing a flag (bool):
```py
def validate_modules(
    installed_modules: Dict[str, Any],
    required_modules: Optional[List[Dict[str, Any]]] = None,
) -> None:
```
or, introduce an alias like:
```py
DEFAULT_REQUIRED_MODULES = [
    {"name": "search", "alias": "searchlight", "ver": 20600},
]
```
assuming that it is safe to equate "search" and "searchlight"

@tylerhutcherson @justin-cechmanek @rbs333

tylerhutcherson:
Following up on this thread -- the simple solution is to provide the additional `searchlight` module+version as another item in the list passed to `validate_modules`. That method does a check to see if either/or of the items in the provided list are present. Otherwise, an exception is raised.

tylerhutcherson:
Fixed here --> https://github.com/redis/redis-vl-python/pull/227

hassan-baig:
I'm seeing the following issue 

It's Mac M1 Pro. 
Redis is running on Docker, I also tried directly.

I'm on the latest package 0.3.4

ERROR - app.services.events.event_manager.EventManager: Unexpected error in __event_listener: Required Redis db module search >= 20600 OR searchlight >= 20600 not installed. See Redis Stack docs at https://redis.io/docs/latest/operate/oss_and_stack/install/install-stack/.

PR #227, opened 2024-10-01
Fix module version checks

tylerhutcherson:
Fixes issue: https://github.com/redis/redis-vl-python/issues/226 related to the search module version checks. Also introduces a clean exception module for adding RedisVL specific exceptions.

### DIFF
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -9,6 +9,7 @@ from redis.asyncio import SSLConnection as AsyncSSLConnection
 from redis.connection import AbstractConnection, SSLConnection
 from redis.exceptions import ResponseError
 
+from redisvl.exceptions import RedisModuleVersionError
 from redisvl.redis.constants import DEFAULT_REQUIRED_MODULES
 from redisvl.redis.utils import convert_bytes
 from redisvl.version import __version__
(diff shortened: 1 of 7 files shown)


redisvl/redis/connection.py
```diff
@@ -143,11 +144,18 @@ def validate_modules(
...
-        f"Required Redis database module {required_module['name']} with version >= {required_module['ver']} not installed. "
-        "See Redis Stack documentation: https://redis.io/docs/stack/"
+    error_message = "Required Redis db module "
+
+    error_message += " OR ".join(
+        [f'{module["name"]} >= {module["ver"]}' for module in required_modules]
+    )
+
+    error_message += (
+        " not installed. See Redis Stack docs at https://redis.io/docs/stack/."
```
justin-cechmanek: this page now redirects to https://redis.io/docs/latest/operate/oss_and_stack/
 is this still the page we want to direct users to?

justin-cechmanek (dismissed):
Looks good! Just one question about the docs page we point to.

redisvl/redis/connection.py
tylerhutcherson: Good catch. Updating!

Issue #232, opened 2024-10-09
Expose show_progress_bar option for Hugging Face vectorizer


antonum:
Currently, hf.embed() and hf.embed_many() always display the tqdm progress bar while building embeddings, even if it's just for one embedding. 
```
# Embed a sentence
hf = HFTextVectorizer(model="sentence-transformers/all-MiniLM-L6-v2")
test = hf.embed("This is a test sentence.")
```

Ask: expose `show_progress_bar` parameter of the underlying HuggingFace model, so one can choose if to display a progress bar.
```
model = SentenceTransformer('sentence-transformers/all-MiniLM-L6-v2')
model.encode(["This is a test sentence."], show_progress_bar=False)
```

So one can do:
```
# Embed a sentence
hf = HFTextVectorizer(model="sentence-transformers/all-MiniLM-L6-v2")
test = hf.embed("This is a test sentence.", show_progress_bar=False)
```

tylerhutcherson:
We can accomplish this by exposing optional `**kwargs` on the `.embed()` + `.embed_many()` (and async variants). Each vectorizer would be responsible for parsing and passing through kwargs as needed. @antonum is this something you might be able to contribute if you're looking for a little OSS action ;) ?

tylerhutcherson:
Solved thanks to https://github.com/redis/redis-vl-python/pull/236

PR #236, opened 2024-10-11
Expose show_progress_bar

antonum:
Address https://github.com/redis/redis-vl-python/issues/232

Now  **kwargs are being passed from `hf.embed()` and `hf.embed_many()` to the underlying `model.encode()`

```
from redisvl.utils.vectorize import HFTextVectorizer
from tqdm.auto import tqdm
hf = HFTextVectorizer(model="sentence-transformers/all-MiniLM-L6-v2")
# Embed a sentence
test = hf.embed("This is a test sentence.", show_progress_bar=True) #progress bar would show
test = hf.embed("This is a test sentence.") #progress bar would show (default behavior as before)
test = hf.embed("This is a test sentence.", show_progress_bar=False) #progress bar would NOT show

# Uncomment to see vector embedding output
print(test[:10])
```

### DIFF
--- a/redisvl/utils/vectorize/text/huggingface.py
+++ b/redisvl/utils/vectorize/text/huggingface.py
@@ -99,7 +99,7 @@ class HFTextVectorizer(BaseVectorizer):
 
         if preprocess:
             text = preprocess(text)
-        embedding = self._client.encode([text])[0]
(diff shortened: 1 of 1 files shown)


tylerhutcherson (approved):
Thanks!

PR #237, opened 2024-10-21
Fix dtype parsing from vectorizer kwargs

tylerhutcherson:
Fixing a bug in how we handled kwargs within the vectorizer classes after a recent change was introduced.

### DIFF
--- a/redisvl/utils/vectorize/base.py
+++ b/redisvl/utils/vectorize/base.py
@@ -81,11 +81,11 @@ class BaseVectorizer(BaseModel, ABC):
             else:
                 yield seq[pos : pos + size]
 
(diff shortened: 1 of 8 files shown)


rbs333 (approved):
Makes sense to me assuming we didn't have any other kwargs for process. Does seem like there's a need to run the linter tho

tylerhutcherson:
> Makes sense to me assuming we didn't have any other kwargs for process. Does seem like there's a need to run the linter tho

`kwargs` for the vectorizers are to be consumed by design by the client doing the vectorization. `dtype` however is reserved for the process_embedding helper, which does not need anything beyond the three explicit args. Thus, needs to get popped from the `kwargs` before processing.

PR #239, opened 2024-10-29
manually parse special attributes that have no associated value

justin-cechmanek:
When converting existing index info into a schema certain attributes have a name and value, while others are indicated as True/False only by their presence in the attributes list. These have to be removed and parsed manually before parsing the remaining items.


justin-cechmanek:
@tylerhutcherson do we have a full list of possible attributes we need to check for?

tylerhutcherson (approved):
LGTM!

PR #248, opened 2024-11-25
Add Amazon Bedrock Text vectorizer

bsbodden:
Add a Vectorizer for Amazon Bedrock, along with tests and doc updates.


redisvl/utils/vectorize/text/bedrock.py
```diff
@@ -0,0 +1,201 @@
...
+
+    .. code-block:: python
+
+        # Initialize with explicit credentials
+        vectorizer = AmazonBedrockTextVectorizer(
+            model_id="amazon.titan-embed-text-v2:0",
+            api_config={
+                "aws_access_key_id": "your_access_key",
+                "aws_secret_access_key": "your_secret_key",
+                "region_name": "us-east-1"
```
tylerhutcherson: I think we should keep the naming convention consistent so that it isn't confusing for end users: `aws_region` might be best for this given the env var is `AWS_REGION`

redisvl/utils/vectorize/text/bedrock.py
```diff
@@ -0,0 +1,201 @@
...
+        if texts and not isinstance(texts[0], str):
+            raise TypeError("Texts must be a list of strings")
+
+        embeddings: List[List[float]] = []
+        dtype = kwargs.pop("dtype", None)
+
+        for batch in self.batchify(texts, batch_size, preprocess):
+            for text in batch:
+                response = self._client.invoke_model(
+                    modelId=self.model, body=json.dumps({"inputText": text})
```
tylerhutcherson, line 192: Is there a way to submit a batch of embedding requests in one call?

redisvl/utils/vectorize/text/bedrock.py
```diff
@@ -0,0 +1,201 @@
...
+        # Generate embeddings
+        embedding = vectorizer.embed("Hello, world!")
+        embeddings = vectorizer.embed_many(["Hello", "World"], batch_size=2)
+    """
+
+    _client: Any = PrivateAttr()
+
+    def __init__(
+        self,
+        model_id: str = "amazon.titan-embed-text-v2:0",
```
tylerhutcherson: Let's use the same init signature as other vectorizers, `model` instead of `model_id`.

tylerhutcherson (changes requested):
Great work. Left a few comments!

redisvl/utils/vectorize/text/bedrock.py
bsbodden: addressed

docs/api/vectorizer.rst
```diff
@@ -61,6 +61,41 @@ CohereTextVectorizer
...
+
+.. _bedrocktextvectorizer_api:
+
+.. currentmodule:: redisvl.utils.vectorize.text.bedrock
+
+.. autoclass:: BedrockTextVectorizer
+   :show-inheritance:
+   :members:
+
+   The BedrockTextVectorizer class utilizes Amazon Bedrock's API to generate
```
tylerhutcherson: You can omit this portion since it is pulled from the docstring of the class (see other vectorizer docs for reference).

docs/api/vectorizer.rst
```diff
@@ -61,6 +61,41 @@ CohereTextVectorizer
    :show-inheritance:
    :members:
 
+BedrockTextVectorizer
+====================
```
tylerhutcherson: Need one more `=` to make it line up with the line length of the `BedrockTextVectorizer`. Otherwise, sphinx will have some issues there parsing

tylerhutcherson (commented):
Docs comment -- overall code looks great. Need tests to pass.

tylerhutcherson (approved):
Nice!

PR #253, opened 2024-12-12
sets the default datatype in our vectorizers to float32 if not specified by users

justin-cechmanek:


tests/integration/test_vectorizers.py
```diff
@@ -238,6 +240,29 @@ def bad_return_type(text: str) -> str:
...
+def test_dtypes(vectorizer):
+    words = "hello"
+
+    raw = vectorizer.embed(words, as_buffer=False)
+
+    default = vectorizer.embed(words, as_buffer=True)
+    assert buffer_to_array(default, dtype="float32") == raw
+
+    float16 = vectorizer.embed(words, as_buffer=True, dtype="float16")
+    # assert buffer_to_array(float16, dtype="float16") == raw # fails
```
justin-cechmanek: For `float16` and `bfloat16` converting to a from this byte array results in rounding errors. Unclear how much of an issue this is, but it will fail on exact match checks.

tylerhutcherson: Why is this happening? Is it avoidable?

redisvl/utils/vectorize/text/azureopenai.py
```diff
@@ -190,7 +190,7 @@ def embed_many(
         if len(texts) > 0 and not isinstance(texts[0], str):
             raise TypeError("Must pass in a list of str values to embed.")
 
-        dtype = kwargs.pop("dtype", None)
```
tylerhutcherson, line 193: I think we might want to approach this in the base class. We can have an optional attribute `self.dtype` that by default is `float32`. But if you set it in the class init, it will be used by all vectorization methods. However, if you pass in at the method level, will override the class attribute. Sort of the classic global var approach. Thoughts?

tests/integration/test_vectorizers.py
justin-cechmanek: Python's native dtype is float64, so when we convert to (b)float16 it's lossy. Converting back from bytes to Python float won't match exactly. It's not a bug, just something that we can keep in mind.

redisvl/utils/vectorize/text/azureopenai.py
justin-cechmanek: We can go with that approach. I'll make the changes. If we go this route we should point it out in the notebooks also.

tests/integration/test_vectorizers.py
tylerhutcherson: Ahhh right. 👍🏼 makes complete sense. I misunderstood initially.

redisvl/utils/vectorize/base.py
```diff
@@ -19,11 +20,22 @@ class Vectorizers(Enum):
 class BaseVectorizer(BaseModel, ABC):
     model: str
     dims: int
+    dtype: str
 
     @property
     def type(self) -> str:
         return "base"
 
+    @validator("dtype")
```
rbs333, line 29: <3

docs/user_guide/vectorizers_04.ipynb
rbs333, line 1: Two small things on this

1. this does have the warning which doesn't really matter but a heads up
image
2. is there a reason mistral is commented out? is there something we need to fix with that

rbs333 (approved):
Few notes but nothing blocking

docs/user_guide/vectorizers_04.ipynb
tylerhutcherson: Yeah let's get rid of this warning. Good catch. 2 -- I think it had to do with stale API keys. @justin-cechmanek are those active again?

justin-cechmanek: Looks like the Mistral client is deprecated. I'll look into updating our vectorizer class for it.

tylerhutcherson: Nice. I see this one here: https://github.com/redis/redis-vl-python/pull/255

After we get this warning removed from the notebook, I am happy to merge!

tylerhutcherson (approved):

PR #254, opened 2025-01-06
fix protobuf broken version for google (optional) dependencies

tylerhutcherson:
The optional dependencies for google AI platform (vertex) bring in protobuf. We need to pin to the correct version that doesn't have a deadly bug (documented [here](https://github.com/protocolbuffers/protobuf/issues/19430)).


tylerhutcherson:
Working on fixing something in the CI

justin-cechmanek (approved):
👍🏼

tylerhutcherson:
@justin-cechmanek @rbs333 @bsbodden one massive limitation we had in our CICD is we didn't pin the poetry version. Turns out, poetry released 2.0.0 just yesterday. So this was failing for yet another reason. I set version to `1.8.3` so it's consistent across all github actions. Use this in your local environments too for now.

PR #255, opened 2025-01-07
migrates mistral vectorizer to new client

justin-cechmanek:


tylerhutcherson (dismissed):
@justin-cechmanek is there a version pinning requirement for the mistralai client to support this change (>= to some version)? Otherwise, looks good to me.

justin-cechmanek:
> @justin-cechmanek is there a version pinning requirement for the mistralai client to support this change (>= to some version)? Otherwise, looks good to me.

It's from 0.*.* to 1.0.0. thanks for reminding me to update the pyproject

rbs333 (approved):

PR #261, opened 2025-01-23
Add a default for the `dtype` field of `BaseVectorizer`

abrookins:
In 0.3.8, we introduced a new field to `BaseVectorizer`, `dtype`, but did not give it a default value. This broke downstream software written against previous versions of redisvl that lacked this field.

Fixes [langchain/langchain-redis#48](https://github.com/langchain-ai/langchain-redis/issues/48).

### DIFF
--- a/redisvl/schema/schema.py
+++ b/redisvl/schema/schema.py
@@ -170,12 +170,12 @@ class IndexSchema(BaseModel):
 
     @root_validator(pre=True)
     @classmethod
-    def validate_and_create_fields(cls, values):
+    def validate_and_create_ields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
         """
         Validate uniqueness of field names and create valid field instances.
         """
         # Ensure index is a dictionary for validation
-        index = values.get("index")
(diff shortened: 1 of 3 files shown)


redisvl/schema/schema.py
```diff
@@ -170,12 +170,12 @@ def _make_field(storage_type, **field_inputs) -> BaseField:
 
     @root_validator(pre=True)
     @classmethod
-    def validate_and_create_fields(cls, values):
+    def validate_and_create_ields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```
tylerhutcherson:
```suggestion
    def validate_and_create_fields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
```

tylerhutcherson: Oh sorry I think I might have jumped on this too soon since it's still draft ;) but just heads up

abrookins: Hah, good catch 😂

tylerhutcherson (approved):
Thank you!

PR #263, opened 2025-01-23
Properly handle asyncio.gather() call in SemanticCache.aheck

abrookins:
An `asyncio.gather()` call should be awaited. Failing to await it causes the tasks to run after the method exits, which can cause unexpected behavior, including failure if the event loop no longer exists.

Fixes RAE-564


tylerhutcherson (approved):
TY! Good catch

PR #267, opened 2025-01-29
Consider vectorizer the owner of dtype

abrookins:
The `dtype` argument to `SemanticCache`, `SemanticRouter`, and `SemanticSessionManager` is intended to specify the vectorizer's data type. However, the constructors for these classes allow passing in both a vectorizer instance (with dtype already set) and `dtype`, in which case, the vectorizer may have a different `dtype` than the `dtype` used in the index schema.

This PR works on eliminating this possibility by adding validation and treating the vectorizer as the true "owner" of `dtype`. We also begin the deprecation process for standalone `dtype` arguments, guiding users to pass in vectorizer instances instead if they require customizing the vectorizer's dtype. Finally, we make vectorizer validation consistent across these classes.

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -86,11 +86,27 @@ class SemanticCache(BaseLLMCache):
         else:
             prefix = name
 
-        # Set vectorizer default
-        if vectorizer is None:
+        # Validate a provided vectorizer or set the default
+        if vectorizer:
+            if not isinstance(vectorizer, BaseVectorizer):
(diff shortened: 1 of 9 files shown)


tylerhutcherson:
@abrookins looks like we'd still need to do a similar treatment to the `SemanticSessionManager` class :)

abrookins:
> @abrookins looks like we'd still need to do a similar treatment to the `SemanticSessionManager` class :)

Ah, yes! Thank you, I'll get on that.

tylerhutcherson (approved):

tylerhutcherson (approved):

PR #270, opened 2025-02-05
Improve CI test runtime with pytest-xdist

tylerhutcherson:
`pytest-xdist` runs workers to tackle the test suite concurrently (using more cpus):
https://pytest-xdist.readthedocs.io/en/stable/distribution.html#running-tests-across-multiple-cpus

The addition of this tool AND some small updates to testcontainer usage (for Redis instances) helps us achieve a ~3-4x speed up locally.


tylerhutcherson:
Some issue with the latest redis stack version???

scripts.py
```diff
@@ -29,12 +29,12 @@ def check_mypy():
...
 
 def test():
-    subprocess.run(["python", "-m", "pytest", "--log-level=CRITICAL"], check=True)
+    subprocess.run(["python", "-m", "pytest", "-n", "6", "--log-level=CRITICAL"], check=True)
 
 
 def test_verbose():
     subprocess.run(
-        ["python", "-m", "pytest", "-vv", "-s", "--log-level=CRITICAL"], check=True
+        ["python", "-m", "pytest", "-n", "6", "-vv", "-s", "--log-level=CRITICAL"], check=True
```
abrookins: I'd use "auto" here. As a data point, if I use "auto" and set Docker's resources to use all cores, pytest assigns 14 cores and I see a ~16s speedup locally (76.39s with 6 cores, 61.40s with "auto"/14 cores).

scripts.py
```diff
@@ -44,6 +44,8 @@ def test_cov():
             "python",
             "-m",
             "pytest",
+            "-n",
+            "6",
```
abrookins: And here.

scripts.py
tylerhutcherson: Nice!

conftest.py
```diff
@@ -8,27 +8,55 @@
...
+        compose_file_name="docker-compose.yml",
+        pull=True,
+    )
+
     compose.start()
 
+    # If you mapped the container port 6379:6379 in docker-compose.yml,
+    # you might have collisions across workers. If you rely on ephemeral
+    # host ports, remove the `ports:` block in docker-compose.yml and do:
     redis_host, redis_port = compose.get_service_host_and_port("redis", 6379)
```
abrookins:
```suggestion
    #redis_host, redis_port = compose.get_service_host_and_port("redis", 6379)
```

conftest.py
```diff
@@ -8,27 +8,55 @@
...
-    redis_url = f"redis://{redis_host}:{redis_port}"
-    os.environ["REDIS_URL"] = redis_url
```
abrookins, line 20: Should line 35 also be commented out? (I added a suggestion for it.)

This isn't a blocker for me, but ideally, the codebase should only support and suggest behavior that's configurable at runtime through environment variables or CLI flags, without requiring developers to comment or uncomment code.

abrookins (approved):
Nice work! That's a big savings in local testing time that should also save time and money in CI over the long run.

conftest.py
tylerhutcherson: I think this is just hangover code I still need to clean up. We don't need to extract the host/port here because DockerCompose via testcontainers is handling the ephemeral port mapping and the `redis_url` fixture below attaches to it when called upon to get the right host/port combination

abrookins (approved):
LGTM 👍

abrookins (approved):

PR #275, opened 2025-02-11
Deprecate global distance_threshold with semantic router

rbs333:
- change ownership of distance_threshold
- move filtering to happen as part of db call
- make code more DRY
- rename user_guide because it's really been bothering me


schemas/semantic_router.yaml
```diff
@@ -18,6 +18,6 @@ vectorizer:
   type: hf
   model: sentence-transformers/all-mpnet-base-v2
 routing_config:
-  distance_threshold: 0.3
```
tylerhutcherson, line 21: Will routingConfig still accept a distance threshold here?

redisvl/extensions/router/schema.py
```diff
@@ -63,13 +63,19 @@ class DistanceAggregationMethod(Enum):
 class RoutingConfig(BaseModel):
     """Configuration for routing behavior."""
 
-    distance_threshold: float = Field(default=0.5)
```
tylerhutcherson, line 66: hanging comment ;)

poetry.lock
tylerhutcherson, line 1: We should restore the OG lockfile as no dependencies changed in this environment due to the PR

redisvl/extensions/router/schema.py
rbs333: Get got as the gen z say

schemas/semantic_router.yaml
rbs333: yes but it'll be a depreciation warning so this is a good catch and should be removed

schemas/semantic_router.yaml
```diff
@@ -18,6 +18,6 @@ vectorizer:
...
   max_k: 2
   aggregation_method: avg
+  distance_threshold: 0.3
```
tylerhutcherson: @rbs333

tylerhutcherson (dismissed):
Nice work!

tylerhutcherson (approved):

poetry.lock
```diff
@@ -6253,4 +6253,4 @@ voyageai = ["voyageai"]
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.9,<3.14"
-content-hash = "e4a76982923cc6e4d9da48a4f924a1f6dad4df913db1195f78a22674eaf04eaa"
+content-hash = "e4a76982923cc6e4d9da48a4f924a1f6dad4df913db1195f78a22674eaf04eaa"
```
rbs333, line 6256: 🤷

tylerhutcherson (approved):

redisvl/extensions/router/schema.py
```diff
@@ -16,7 +16,7 @@ class Route(BaseModel):
     """List of reference phrases for the route."""
     metadata: Dict[str, str] = Field(default={})
     """Metadata associated with the route."""
-    distance_threshold: Optional[float] = Field(default=None)
+    distance_threshold: float = Field(default=0.5)
```
abrookins, line 19: I saw you deleted another validator, so maybe this isn't important, but with Pydantic 2, you can validate that a number is within a range using `Annotated`:

```
from typing_extensions import Annotated

from pydantic.v1 import BaseModel, Field, ValidationError

class Foo(BaseModel):
    bar: Annotated[float, Field(strict=True, gt=0, lte=1)]  # or gt, lt


for bad_number in (-1, 2):
    try:
        Foo(bar=bad_number) 
    except ValidationError:
        print("worked")
    else:
        raise RuntimeError("Didn't work!")

print(Foo(bar=0.5))
```

This should still work with a v1 BaseModel.

redisvl/extensions/router/schema.py
```diff
@@ -63,27 +63,28 @@ class DistanceAggregationMethod(Enum):
...
+    """Aggregation method to use to classify queries."""
     aggregation_method: DistanceAggregationMethod = Field(
         default=DistanceAggregationMethod.avg
     )
-    """Aggregation method to use to classify queries."""
+
+    """The maximum number of top matches to return."""
+    distance_threshold: float = Field(
+        default=0.5,
+        deprecated=True,
```
abrookins, line 78: On a v1 BaseModel, this only adds "deprecated" to the JSON schema for this field, since what's really happening is any extra keyword args pass through to the schema. That's not ideal, but I'm ok with it. In v2, this would log a deprecation warning if a user used the field. We can leave it in place like this and then if it's still around when we cut over to using v2 BaseModel, it'll work better.

redisvl/extensions/router/semantic.py
```diff
@@ -241,15 +252,26 @@ def _build_aggregate_request(
...
         vector: List[float],
-        distance_threshold: float,
         aggregation_method: DistanceAggregationMethod,
-    ) -> RouteMatch:
-        """Classify to a single route using a vector."""
+        max_k: int = 1,
+    ) -> List[RouteMatch]:
+        """Get the route matches for a given vector and aggregation method."""
+
+        thresholds: List[float] = [route.distance_threshold for route in self.routes]
```
abrookins: Why the variable type annotation here? Did mypy not pick up that `route.distance_threshold` is a float?

abrookins (approved):
Couple of comments, but this looks good overall!

redisvl/extensions/router/schema.py
rbs333: Oh that's nice

rbs333: Yeah I deleted a validator for the now depreciated field since it has no impact. I think we should update to the new style of validators but I think that should be included in the pr for pydantic versioning stuff because it fits more with that theme: https://redislabs.atlassian.net/browse/RAAE-607

redisvl/extensions/router/semantic.py
rbs333: Good catch it failed mypy originally because the pydantic model had it as optional so I update there but forgot to delete this. Pushed a fix

abrookins (approved):

PR #280, opened 2025-02-13
Refactor async client connection-handling

abrookins:
Refactor interfaces related to the Redis client in AsyncRedisIndex (and possibly others).

## Constraints
- We want to initialize the Redis connection lazily, not at object instantiation or module import.
- Python doesn't have async properties, so properties are limited to sync function calls
    - Therefore, we don't want properties to be part of the interface
- For functions we "remove," we'll leave them in place with a deprecation warning
    - Remaining backwards-compatible until the next major release

## The Design Today
- Can create a client and set it using `index.connect()`
    - Pass in a `redis_url`
    - Calls `index.set_client()`
- Can pass in a client and set it with `index.set_client()`
    - Pass in a client instance
- **Can't** pass in a client instance with `__init__()`
    - Or anything URL, etc.
- Can create a client with `index.from_existing()`
    - Pass it a `redis_url`
    - Calls `index.set_client()`
- **Can't** use index as an async context manager
    - Requires explicit resource handling, `atexit` handler
    - But this is broken
- The `setup_async_redis()` decorator creates a function wrapper that calls `validate_async_redis()` with every `set_client()` call
- The `RedisConnectionFactory.connect()` returns incompatible types (sync, async `Redis`)
    - Sync vs. async depends on a parameter

## The Design After Refactor
- **Can** use as an async context manager
    - Either disconnect the index manually with `disconnect()` or use it as a context manager
    - `async with Index()...` will `disconnect()` after you exit the context block
- Lazily instantiate client with `self._get_client()` method ("private")
- Remove `set_client()` public interface if possible
    - Lazily instantiate client
- Remove `connect()` public interface if possible
    - Lazily instantiate client
- Leave `from_existing()`
- Leave `client()` property, now just returns `._redis_client` and can be None
- Call `validate_async_redis()`  when setting a new client instance for the first time
    - But make this an internal check rather than attached to public interface
- Allow `redis_url`, `redis_kwargs`, or `redis_client` in `__init__()`

### Examples Post-Refactor
```python
#1: New instance with redis URL
index = AsyncRedisIndex(redis_url="...")
#2: New instance with existing client
index = AsyncRedisIndex(redis_client=client)
#3: New instance with `from_existing`
index = AsyncRedisIndex.from_existing(..., redis_url="...")

#4 Passing both a client and a URL is isallowed
try:
    index = AsyncRedisIndex(redis_client=client, redis_url="...")
except:
    pass
else:
    raise RuntimeError("Should have raised!")

# The client is lazily connected here, and we send telemetry.
await index.something_using_redis()
await index.something_else_using_redis()
# Close the client.
await index.close()

async with AsyncRedisIndex(redis_url="...") as index:
    # Same story: client is lazily created now
    await index.something_using_redis()
    await index.something_else_using_redis()
    # Client is automatically closed

# __enter__ is reentrant
async with index:
    # Lazily opens a new client connection.
    await index.a_third_thing_using_redis()
    # Client is automatically closed
```


docs/user_guide/02_hybrid_queries.ipynb
```diff
@@ -142,9 +139,16 @@
...
    "cell_type": "code",
-   "execution_count": 26,
+   "execution_count": 6,
    "metadata": {},
    "outputs": [
+    {
+     "name": "stdout",
+     "output_type": "stream",
+     "text": [
+      "@credit_score:{high}=>[KNN 10 @user_embedding $vector AS vector_distance] RETURN 6 user credit_score age job office_location vector_distance SORTBY vector_distance ASC DIALECT 2 LIMIT 0 10\n"
```
tylerhutcherson: Looks like something is logging out the query string by default? I do like showing this sometimes in our notebooks. But if this happens on every `.query()` might be too noisy.

redisvl/redis/connection.py
```diff
@@ -255,19 +309,22 @@ def get_async_redis_connection(url: Optional[str] = None, **kwargs) -> AsyncRedi
             ValueError: If url is not provided and REDIS_URL environment
                 variable is not set.
         """
-        if url:
-            return AsyncRedis.from_url(url, **kwargs)
-        # fallback to env var REDIS_URL
-        return AsyncRedis.from_url(get_address_from_env(), **kwargs)
+        warn(
```
tylerhutcherson, line 312: Let's go ahead and create a ticket for this so we don't forget

tests/integration/test_async_search_index.py
```diff
@@ -313,10 +312,87 @@ async def test_check_index_exists_before_search(async_client, async_index):
...
+    assert async_index._redis_client == client
+
+
+@pytest.mark.asyncio
+async def test_search_index_that_owns_client_context_manager(index_schema, redis_url):
+    async_index = AsyncSearchIndex(schema=index_schema, redis_url=redis_url)
+    async with async_index:
+        await async_index.create(overwrite=True, drop=True)
+        assert async_index._redis_client
+    assert async_index._redis_client is None
```
tylerhutcherson, line 368: So if providing a connection URL to let redisvl manage your connection, it is not maintaining that connection when using the context manager. That makes sense, but does also mean the usage should ideally be well documented

tests/integration/test_connection.py
```diff
@@ -19,6 +18,9 @@
 
 EXPECTED_LIB_NAME = f"redis-py(redisvl_v{__version__})"
 
+# Remove after we remove connect() method from RedisConnectionFactory
+pytestmark = pytest.mark.filterwarnings("ignore::DeprecationWarning")
```
tylerhutcherson, line 22: very cool

redisvl/index/index.py
```diff
@@ -368,26 +376,27 @@ def connect(self, redis_url: Optional[str] = None, **kwargs):
...
         .. code-block:: python
 
             index.connect(redis_url="redis://localhost:6379")
 
         """
-        client = RedisConnectionFactory.connect(
-            redis_url=redis_url, use_async=False, **kwargs
+        # TODO: Intentionally not including required modules to match existing
+        # behavior, but we need to review.
+        self.__redis_client = RedisConnectionFactory.get_redis_connection(
```
tylerhutcherson, line 390: I believe the existing behavior still validated modules -- we used the `@setup_redis` decorator on `set_client`
https://github.com/redis/redis-vl-python/pull/280/files#diff-ebc75d7c32f45e2d6c0c7ea45a8ab92fddb96197967b64ab1572dec4efbcad4aL388

redisvl/index/index.py
```diff
@@ -255,8 +229,7 @@ class SearchIndex(BaseSearchIndex):
         from redisvl.index import SearchIndex
 
         # initialize the index object with schema from file
-        index = SearchIndex.from_yaml("schemas/schema.yaml")
-        index.connect(redis_url="redis://localhost:6379")
+        index = SearchIndex.from_yaml("schemas/schema.yaml", redis_url="redis://localhost:6379")
```
tylerhutcherson, line 234: Can we add a few more examples of init to the docstring here since this has changed? Ditto for the async class

redisvl/index/index.py
```diff
@@ -864,106 +894,101 @@ async def from_existing(
...
-
-        Args:
-            redis_client (aredis.Redis): An Async Redis
-                client instance to be used for the connection.
-
-        Raises:
-            TypeError: If the provided client is not valid.
-
-        .. code-block:: python
+        # TODO: Intentionally not including required modules to match existing
```
tylerhutcherson: Ditto to above -- this did validate modules by running through the decorator used on `set_client`

tylerhutcherson (approved):
Huge effort; great work! Left a few comments, but once addressed merge away

tylerhutcherson:
Ahh one more thing, we need to make sure to update the `README`

tests/integration/test_async_search_index.py
abrookins: So it's actually the opposite if I understood you: if you give it a client object that means you are managing the object, but if you give it connection details and it creates the client object, then it manages the object and will disconnect and null it out when the context exits.

abrookins:
Thanks, @tylerhutcherson! I saw a typo in a comment, too. I'll get those changes in and send it back to you this morning.

tests/integration/test_async_search_index.py
tylerhutcherson: We're on the same page. cool

PR #281, opened 2025-02-13
Upgrade to Pydantic V2 models

tylerhutcherson:
This PR introduces `pydantic ^2.0.0` models, validators, and techniques. The implementation here improves performance of model construction and validation. It also introduces best practices when possible related to validators and serialization. Because RedisVL was already on pydantic v2, this change should not impact any end users as we already required this version and had been relying on the `v1` shim.


abrookins:
Beast PR 🔥

redisvl/extensions/session_manager/semantic_session.py
```diff
@@ -110,7 +110,7 @@ def __init__(
             existing_index = SearchIndex.from_existing(
                 name, redis_client=self._index.client
             )
-            if existing_index.schema != self._index.schema:
+            if existing_index.schema.to_dict() != self._index.schema.to_dict():
```
abrookins, line 113: Glad this worked, this is what I was going to suggest!

PR #282, opened 2025-02-13
Support Redis 8

tylerhutcherson:
This PR adds testing support for Redis 8 (currently `redis:8.0-M03` from Docker) to our GitHub Actions pipeline. Overall, our entire test suite passed against Redis 8 with one notable exception:

**Breaking Change in Redis 8, Dialect 2 Only**
When a search query references a field that is not present in the index schema, Redis 8 now raises:
```
redis.exceptions.ResponseError: Unknown field at offset NN near <unknown-field-name>
```
Previously (and in other dialects), Redis would allow such queries but return empty results. This new behavior is arguably more correct, but it breaks a test that expects an empty result when searching on a non-indexed field.

## Context and Discussion
A single test failed where our code intentionally queries a field not in the index, expecting empty results. With Redis 8, Dialect 2, this now raises an error.
The team discussed whether to:
- Adjust the test to skip or otherwise handle this new behavior (since querying a non-indexed field arguably should be an error).
- Catch and raise a more user-friendly message in our library. The Redis error text can be cryptic, so providing a clearer explanation to users might be beneficial.
## Decision
- Testing: We will proceed with adding Redis 8 to our CI matrix.
- The Failing Test: We will update this test to either skip on Redis 8 or handle the error more gracefully.
- Error Messaging: We may consider adding a better exception message so that developers get a more understandable error when they inadvertently query non-indexed fields. But this will go in a different PR.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,10 +60,11 @@ jobs:
 
       - name: Set Redis image name
         run: |
-          if [[ ${{ matrix.redis-stack-version }} == '8.0-M03']]; then
+          if [[ "${{ matrix.redis-stack-version }}" == "8.0-M03" ]]; then
             echo "REDIS_IMAGE=redis:${{ matrix.redis-stack-version }}" >> $GITHUB_ENV
           else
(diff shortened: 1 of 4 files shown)


rbs333 (approved):
👍

PR #283, opened 2025-02-13
Avoid looking at local variables for deprecation warnings

abrookins:
Deprecation warnings should not be looking at local function variables. 😅


justin-cechmanek (dismissed):
Looks good 👍🏼

tylerhutcherson:
LGTM! Thanks for the fix and wow look at those tests 🥇 👍🏼

tylerhutcherson (approved):

PR #289, opened 2025-02-21
fix url pass

rbs333:
There was a bug in from_existing where it was not passing the redis_url correctly because the var is named `url`
image

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -358,7 +358,7 @@ class SearchIndex(BaseSearchIndex):
             with self._lock:
                 if self.__redis_client is None:
                     self.__redis_client = RedisConnectionFactory.get_redis_connection(
-                        url=self._redis_url,
+                        redis_url=self._redis_url,
                         **self._connection_kwargs,
                     )
         return self.__redis_client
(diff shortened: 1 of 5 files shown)


tests/integration/test_search_index.py
```diff
@@ -84,6 +84,17 @@ def test_search_index_from_existing(client, index):
     assert index2.schema == index.schema
 
 
+def test_search_index_from_existing_url(redis_url, index):
```
abrookins, line 87: Most excellent

abrookins (approved):

redisvl/index/index.py
```diff
@@ -358,7 +358,7 @@ def _redis_client(self) -> Optional[redis.Redis]:
             with self._lock:
                 if self.__redis_client is None:
                     self.__redis_client = RedisConnectionFactory.get_redis_connection(
-                        url=self._redis_url,
+                        redis_url=self._redis_url,
```
rbs333, line 361: personally I think we should always use redis_url vs url because it's more specific and that's what we tend to say when we do it

abrookins: Agreed

tylerhutcherson: yup

PR #292, opened 2025-02-27
Add distance threshold optimizer classes

rbs333:
- add base, router, and cache threshold optimizer
- uses ranx to support eval metrics
- class is designed to receive a optimizer class in the case where someone wants to supply their own without us having to make a release to support something custom


redisvl/extensions/router/semantic.py
```diff
@@ -293,6 +301,9 @@ def _get_route_matches(
                 )
             raise e
 
+        for match in aggregation_result.rows:
+            print(f"\n {match=}")
```
tylerhutcherson: Assume this is a debugging section?

redisvl/utils/threshold_optimizer/base.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+class EvalMetric(Enum):
+    """Evaluation metrics for threshold optimization."""
+
+    F1 = "f1"
+    PRECISION = "precision"
+    RECALL = "recall"
+
+    def __str__(self) -> str:
```
tylerhutcherson: I think `EvalMetric(str, Enum)` would do the trick? See how we handle other enums in the lib

redisvl/utils/threshold_optimizer/cache.py
```diff
@@ -0,0 +1,85 @@
...
+
+def _get_best_threshold(metrics: dict) -> float:
+    """
+    Returns the threshold with the highest F1 score.
+    If multiple thresholds have the same F1 score, returns the lowest threshold.
+    """
+    return max(metrics.items(), key=lambda x: (x[1]["score"], -x[0]))[0]
+
+
+def _grid_search_opt_cache(
```
tylerhutcherson: any particular reason these aren't members of the class?

redisvl/utils/threshold_optimizer/schema.py
```diff
@@ -0,0 +1,11 @@
+from typing import List, Optional
+
+from pydantic import BaseModel, Field
+from ulid import ULID
+
+
+class TestData(BaseModel):
+    q_id: str = Field(default_factory=lambda: str(ULID()))
```
tylerhutcherson: `_id` feels maybe cleaner than knowing what the `q` is for ;)

redisvl/utils/threshold_optimizer/__init__.py
tylerhutcherson, line 1: I would add imports here similar to vectorizers so we can do `from redisvl.utils.threshold_optimizer import CacheThresholdOptimizer`

Might also consider just naming it `redisvl.utils.optimize` to follow the same grammar as `redisvl.utils.vectorize`

tylerhutcherson (approved):
Very nice! Left some of my feedback but I am pumped about this.

redisvl/utils/threshold_optimizer/cache.py
rbs333: My thought process is this shows that you can pass in any function that meets this definition not dependant on any class state if you wanted to go a more custom route. It's a little atypical though since it is still an internal method that doesn't really make sense to expose outside of this context.

redisvl/utils/threshold_optimizer/schema.py
rbs333: used q_id since that's ranx convention for this but yeah _id works

PR #295, opened 2025-03-18
Add datetime-enabled filters

rbs333:
Goal:
1. Create a new DatetimeFilter or TimestampFilter
      -  Allow querying for a specific date without time
      -  Allow querying for a specific date with time
      -  Allow querying for a date range
      -  Allow querying for a time range
      -  Allow querying with or without a timezone
      -  Default to timezone-aware UTC datetimes

2. Alternatively, create a new Timestamp field type that allows specifying via YAML or dictionary that a numeric field is actually a timestamp, with or without a timezone.


redisvl/query/filter.py
```diff
@@ -562,3 +565,394 @@ def __str__(self) -> str:
...
+    }
+
+    OPERATOR_MAP = {
+        FilterOperator.EQ: "@%s:[%s %s]",  # For exact timestamp match (converted to range for dates)
+        FilterOperator.NE: "(-@%s:[%s %s])",
+        FilterOperator.GT: "@%s:[(%s +inf]",
+        FilterOperator.LT: "@%s:[-inf (%s]",
+        FilterOperator.GE: "@%s:[%s +inf]",
+        FilterOperator.LE: "@%s:[-inf %s]",
+        FilterOperator.BETWEEN: "@%s:[(%s (%s]",  # should between be inclusive?
```
rbs333: Right now "between" is defined as non-inclusive - not sure if that's the right decision could see it either way.

docs/user_guide/02_hybrid_queries.ipynb
abrookins, line 1: Updating the docs in a notebook? 🔥 🔥

redisvl/query/filter.py
abrookins: I suggest making this inclusive by default and either parameterize it or embed meaning into multiple operators. Since many people are familiar with Pandas, we might follow the Pandas interface, which uses an `inclusive` kwarg: https://pandas.pydata.org/docs/reference/api/pandas.date_range.html

tests/integration/test_query.py
```diff
@@ -310,6 +347,27 @@ def test_filters(index, query):
...
+    t = Timestamp("last_updated") < sample_datetimes["high"]
+    search(query, index, t, 5)
+
+    t = Timestamp("last_updated") <= sample_datetimes["mid"]
+    search(query, index, t, 5)
+
+    t = Timestamp("last_updated") == sample_datetimes["mid"]
+    search(query, index, t, 3)
+
+    t = Timestamp("last_updated").between(
```
abrookins: An `inclusive` kwarg matching the behavior in Pandas for `date_range` could work well.

redisvl/query/filter.py
```diff
@@ -8,6 +10,19 @@
 # mypy: disable-error-code="override"
 
 
+class Inclusive(str, Enum):
+    """Enumeration for distance aggregation methods."""
```
tylerhutcherson: Different docstring

redisvl/query/filter.py
```diff
@@ -373,10 +392,61 @@ def __le__(self, other: int) -> "FilterExpression":
...
+        self, start: int, end: int, inclusive: str = "both"
+    ) -> "FilterExpression":
+        """Create a Numeric equality filter expression.
+
+        Args:
+            other (int): The value to filter on.
+
+        .. code-block:: python
+
+            from redisvl.query.filter import Num
```
tylerhutcherson: Docstring

tests/unit/test_filter.py
```diff
@@ -292,3 +306,179 @@ def test_num_filter_zero():
...
+    assert str(ts) == "*"
+
+    ts = Timestamp("created_at") != None
+    assert str(ts) == "*"
+
+    ts = Timestamp("created_at") > None
+    assert str(ts) == "*"
+
+
+def test_timestamp_invalid_input():
```
tylerhutcherson, line 460: On a similar theme, what happens when the `Timestamp` field is called / created for a field that isn't numeric in the index? How does that behave? Given there is a requirement that `Timestamp` filters are tied to `num` fields, thinking through how we should document that too.

docs/user_guide/02_hybrid_queries.ipynb
tylerhutcherson, line 1: Small issue here:

```
11:03:03 [RedisVL] INFO   Indices:
11:03:03 [RedisVL] INFO   1. float64_session
11:03:03 [RedisVL] INFO   2. float64_cache
11:03:03 [RedisVL] INFO   3. float16_cache
11:03:03 [RedisVL] INFO   4. float32_session
11:03:03 [RedisVL] INFO   5. float16_session
11:03:03 [RedisVL] INFO   6. bfloat_session
11:03:03 [RedisVL] INFO   7. float32_cache
11:03:03 [RedisVL] INFO   8. bfloat_cache
11:03:03 [RedisVL] INFO   9. user_queries
```

This is collecting indices from your local instance that weren't generated by the notebook (flush the db before running the notebook and recommit)

tylerhutcherson (approved):
Very cool! Left a few notes on docstrings and stdouts. Good from me once addressed!

tylerhutcherson:
Ahh one more thing: new class so we need to add to our docs.
https://github.com/redis/redis-vl-python/blob/main/docs/api/filter.rst

tests/unit/test_filter.py
rbs333: This is a great question. Currently if you tried to do this with a text field, for example, it would let you create the filter but then fail on type when trying to perform the search itself. This would provide a hint to the user but is not perfect. 

I think it might be worth an enhancement to add a function that checks this ahead of time for all filter types against schema. Because right now I don't think we provide a meaningful error for a text field on a numeric either.

@abrookins thoughts?

PR #298, opened 2025-03-24
Add option to normalize vector distances on query

rbs333:
This pr accomplishes 2 goals:
1. Add an option for users to easily get back a similarity value between 0 and 1 that they might expect to compare against other vector dbs. 
2. Fix the current bug that `distance_threshold` is validated to be between 0 and 1 when in reality it can take values between 0 and 2. 

> Note: after much careful thought I believe it is best that for `0.5.0` we do **not** start enforcing all distance_thresholds between 0 and 1 and move to this option as default behavior. Ideally this metric would be consistent throughout our code and I don't love supporting this flag but I think it provides the value that is scoped for this ticket while inflicting the least amount of pain and confusion. 

Changes:

1. Adds the `normalize_vector_distance` flag to VectorQuery and VectorRangeQuery. 

Behavior:
- If set to `True` it normalizes values returned from redis to a value between 0 and 1. 
- For cosine similarity, it applies `(2 - value)/2`.
- For L2 distance, it applies normalization `(1/(1+value))`.
- For IP, it does nothing and throws a warning since normalized IP is cosine by definition.
- For VectorRangeQuery, if `normalize_vector_distance=True` the distance threshold is now validated to be between 0 and 1 and denormalized for execution against the database to make consistent. 

2. Relaxes validation for semantic caching and routing to be between 0 and 2 fixing the current bug and aligning with how the database actually functions.

### DIFF
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -238,9 +238,9 @@ class SemanticCache(BaseLLMCache):
         Raises:
             ValueError: If the threshold is not between 0 and 1.
         """
-        if not 0 <= float(distance_threshold) <= 1:
+        if not 0 <= float(distance_threshold) <= 2:
             raise ValueError(
-                f"Distance must be between 0 and 1, got {distance_threshold}"
+                f"Distance must be between 0 and 2, got {distance_threshold}"
             )
         self._distance_threshold = float(distance_threshold)
 
--- a/redisvl/extensions/llmcache/semantic.py
+++ b/redisvl/extensions/llmcache/semantic.py
@@ -22,7 +22,7 @@ from redisvl.extensions.llmcache.schema import (
     SemanticCacheIndexSchema,
 )
 from redisvl.index import AsyncSearchIndex, SearchIndex
-from redisvl.query import RangeQuery
+from redisvl.query import VectorRangeQuery
 from redisvl.query.filter import FilterExpression
 from redisvl.query.query import BaseQuery
 from redisvl.redis.connection import RedisConnectionFactory
@@ -390,7 +390,7 @@ class SemanticCache(BaseLLMCache):
         vector = vector or self._vectorize_prompt(prompt)
         self._check_vector_dims(vector)
 
-        query = RangeQuery(
+        query = VectorRangeQuery(
             vector=vector,
             vector_field_name=CACHE_VECTOR_FIELD_NAME,
             return_fields=self.return_fields,
@@ -473,7 +473,7 @@ class SemanticCache(BaseLLMCache):
         vector = vector or await self._avectorize_prompt(prompt)
         self._check_vector_dims(vector)
 
-        query = RangeQuery(
+        query = VectorRangeQuery(
             vector=vector,
             vector_field_name=CACHE_VECTOR_FIELD_NAME,
             return_fields=self.return_fields,
@@ -481,6 +481,7 @@ class SemanticCache(BaseLLMCache):
             num_results=num_results,
             return_score=True,
             filter_expression=filter_expression,
+            normalize_vector_distance=True,
         )
 
         # Search the cache!
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -53,7 +53,7 @@ from redisvl.redis.connection import (
 )
 from redisvl.redis.utils import convert_bytes
 from redisvl.schema import IndexSchema, StorageType
-from redisvl.schema.fields import VectorDistanceMetric
+from redisvl.schema.fields import VECTOR_NORM_MAP, VectorDistanceMetric
 from redisvl.utils.log import get_logger
 
 logger = get_logger(__name__)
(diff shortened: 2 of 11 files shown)


redisvl/index/index.py
```diff
@@ -75,11 +87,18 @@ def process_results(
...
-        (storage_type == StorageType.JSON)
+        (schema.index.storage_type == StorageType.JSON)
         and isinstance(query, FilterQuery)
         and not query._return_fields  # type: ignore
     )
 
+    normalize_cosine_distance = (
+        (isinstance(query, VectorQuery) or isinstance(query, VectorRangeQuery))
+        and query._normalize_cosine_distance
+        and schema.fields[query._vector_field_name].attrs.distance_metric  # type: ignore
```
rbs333: Pass in full schema such that can search for specific vector field in the case that there are multiple vector fields in the index.

redisvl/query/query.py
```diff
@@ -188,6 +188,7 @@ def __init__(
         dialect: int = 2,
         sort_by: Optional[str] = None,
         in_order: bool = False,
+        normalize_cosine_distance: bool = False,
```
rbs333: Added to VectorQuery and RangeVectorQuery class because vector distance does not apply to filter and count queries and therefore would not be part of base.

redisvl/utils/utils.py
```diff
@@ -191,3 +191,10 @@ def wrapper():
             return
 
     return wrapper
+
+
+def norm_cosine_distance(value: float) -> float:
```
rbs333, line 196: Could add additional check logic to this function kept it simple stupid to start

redisvl/query/query.py
rbs333: I also named the field `normalize_cosine_distance` since it only applies to that distance metric not to all vector calculations.

tylerhutcherson:
Nice @rbs333 good start. We do want to enable normalization regardless of distance metric.

1) So imagine a map between chosen/indexed distance metric and the normalization function by distance metric. So in that vein, we can rename the param to a generic `normalize` naming!
2) Redis maps cosine distance to 0-2 both in input and output. and I think our goal is to normalize it to 0-1 instead (which is how most users think of and use our implementation anyways because 0-2 is not well documented on redis docs.
3) there are some places where we currently (wrongly) enforce the distance threshold between 0-1, even though it's technically 0-2. (see semantic cache base class). this will be "correct" once you make this change as we can internally control that normalization param on the semantic cache query definitions (internals)

tylerhutcherson:
> > Nice @rbs333 good start. We do want to enable normalization regardless of distance metric.
> > 
> > 1. So imagine a map between chosen/indexed distance metric and the normalization function by distance metric. So in that vein, we can rename the param to a generic `normalize` naming!
> > 2. Redis maps cosine distance to 0-2 both in input and output. and I think our goal is to normalize it to 0-1 instead (which is how most users think of and use our implementation anyways because 0-2 is not well documented on redis docs.
> > 3. there are some places where we currently (wrongly) enforce the distance threshold between 0-1, even though it's technically 0-2. (see semantic cache base class). this will be "correct" once you make this change as we can internally control that normalization param on the semantic cache query definitions (internals)
> 
> 1. Heard and can do.
> 
> Is inner product also bounded between 0-2 returned from redis? It is not documented well and if that is the case this formula is expressed wrong in the docs.
> 
> For euclidean, what normalization were you thinking? It's not obvious to me.
> 
> image
> 2. As implemented currently the output of the normalization is between 0 and 1.
> 3. Heard and will update.

For euclidean we should be able to do `1/(1-raw_dist)` since the raw distance value is unbounded on the upper limit.

redisvl/extensions/llmcache/semantic.py
```diff
@@ -22,7 +22,7 @@
     SemanticCacheIndexSchema,
 )
 from redisvl.index import AsyncSearchIndex, SearchIndex
-from redisvl.query import RangeQuery
```
tylerhutcherson, line 25: excellent, thank you!

redisvl/index/index.py
```diff
@@ -75,11 +87,26 @@ def process_results(
...
     # Determine if unpacking JSON is needed
     unpack_json = (
-        (storage_type == StorageType.JSON)
+        (schema.index.storage_type == StorageType.JSON)
         and isinstance(query, FilterQuery)
         and not query._return_fields  # type: ignore
     )
 
+    if (
+        isinstance(query, VectorQuery) or isinstance(query, VectorRangeQuery)
```
tylerhutcherson: Should be able to check `BaseVectorQuery` given they both inherit from here

tylerhutcherson (commented):
Ok looking solid, left a few comments. But unfortunately, we might need one more discussion on this as I am reading this code and thinking about 2 things:

- Given L2 is unbounded by definition, is there a scenario here where having this value normalized makes any sense in a use case? We might want to actually take a peek at 2-3 other vector db solutions to see if they document how they normalize these kinds of metrics.
- Secondly, and more importantly, if a user sets `distance_threshold` as a value between [x, y], wouldn't they expect the vector distance they get on the output to match this? Tricky

rbs333:
> Ok looking solid, left a few comments. But unfortunately, we might need one more discussion on this as I am reading this code and thinking about 2 things:
> 
> * Given L2 is unbounded by definition, is there a scenario here where having this value normalized makes any sense in a use case? We might want to actually take a peek at 2-3 other vector db solutions to see if they document how they normalize these kinds of metrics.
> * Secondly, and more importantly, if a user sets `distance_threshold` as a value between [x, y], wouldn't they expect the vector distance they get on the output to match this? Tricky

@tylerhutcherson personally I think we should limit scope of this ticket to just normalize for cosine distance. The customer needs we are addressing, that I have actually witnessed people get confused by, is having an option for a vector similarity type output for cosine. I have not seen a single demo from a competitor or use case from a customer where they used euclidean distance and especially not normalized euclidean distance. I think we risk spinning our wheels on something nobody really cares about. 

On the second point, yeah this was something I debated as well. Ideally I think we get to a point that for RangeQueries, at very least, normalization is the default option such that a `distance_threshold` between 0 and 1 always makes sense. The extra troubling thing is that this would work except for the cases involving the `semantic_router` right now we use aggregations to handle that with distance_threshold where we can't normalize ahead of time which makes things complicated.

redisvl/query/query.py
```diff
@@ -493,6 +519,14 @@ def set_distance_threshold(self, distance_threshold: float):
             raise TypeError("distance_threshold must be of type float or int")
         if distance_threshold < 0:
             raise ValueError("distance_threshold must be non-negative")
+        if self._normalize_vector_distance:
+            if distance_threshold > 1:
```
tylerhutcherson, line 523: nice!

tylerhutcherson (approved):
NICE

PR #299, opened 2025-03-24
Adds support for HYBRID_POLICY on KNN queries with filters

bsbodden:
# Add support for hybrid policy and epsilon parameters in vector searches

This PR adds support for configuring hybrid search policy for vector searches and epsilon for vector range queries in RedisVL, matching the capabilities available in Redis Vector Search.

### Added features:
1. **Hybrid Policy Control for VectorQuery**
   - Added `hybrid_policy` parameter with options `BATCHES` or `ADHOC_BF`
   - Added `batch_size` parameter for controlling batch size in `BATCHES` mode
   - Implemented methods to get/set these parameters

2. **EPSILON Support for VectorRangeQuery**
   - Added `epsilon` parameter to control range query boundaries
   - Implemented methods to get/set this parameter
   - Properly adds epsilon to query attributes, not as direct parameter

### Tests:
- Added unit tests for new parameters and methods
- Added integration tests that verify query construction but avoid execution with unsupported parameters

## Description

Redis Vector Search allows fine-tuning of vector queries through hybrid policy selection and epsilon configuration. This PR exposes these parameters in RedisVL, giving users better control over performance and accuracy trade-offs:

- **Hybrid Policy**: Controls how filters are applied during vector search:
  - `BATCHES`: Paginates through small batches of nearest neighbors
  - `ADHOC_BF`: Computes scores for all vectors passing the filter

- **Epsilon**: For range queries, controls boundary expansion through `radius * (1 + epsilon)`, enabling deeper search at the expense of performance

### DIFF
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -188,6 +188,8 @@ class VectorQuery(BaseVectorQuery, BaseQuery):
         dialect: int = 2,
         sort_by: Optional[str] = None,
         in_order: bool = False,
+        hybrid_policy: Optional[str] = None,
+        batch_size: Optional[int] = None,
     ):
         """A query for running a vector search along with an optional filter
         expression.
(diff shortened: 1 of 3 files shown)


tylerhutcherson (approved):
Looks great -- only question -- are the hybrid policy and batches params only for KNN queries or vector range queries too? Can you also add the docstring for those similar to how you did for the epsilon param?

bsbodden:
> Looks great -- only question -- are the hybrid policy and batches params only for KNN queries or vector range queries too? Can you also add the docstring for those similar to how you did for the epsilon param?

Did the same treatment for the VectorRangeQuery - added more tests and fix the docstrings.

PR #301, opened 2025-03-25
Fix: RedisVL breaks existing logging configurations

### LINKED ISSUE #293: RedisVL breaks existing logging configurations
Hi,

Thank you for this very helpful repo. Redis has always been and is still total 🔥.

Simply importing some structures from RedisVL breaks existing logging configurations, because they get overridden by the following function call: https://github.com/redis/redis-vl-python/blob/ad9bb211c37732ebbd28299cb3cd527eb1f44835/redisvl/utils/log.py#L18

Here is a simple reproducer:
```python
import logging
import sys

handler = logging.StreamHandler(sys.stderr)
handler.setFormatter(
    logging.Formatter(
        "%(asctime)s %(levelname)s [%(name)s] [%(filename)s:%(lineno)s] %(message)s"
    )
)

root_logger = logging.getLogger()
root_logger.handlers = [handler]
root_logger.setLevel(logging.INFO)

my_logger = logging.getLogger("app")
my_logger.info("This is correctly formatted")

from redisvl.query.filter import Text

my_logger.info("This is NOT correctly formatted and has been overridden by RedisVL")
```

Output:
```bash
➜  redisvl-issue-log poetry run python redisvl_issue_log/repro.py
2025-02-28 18:18:24,036 INFO [app] [repro.py:16] This is correctly formatted
18:18:24 app INFO   This is NOT correctly formatted and has been overridden by RedisVL
```

This is a very frustrating issue, since it is happening silently at the first RedisVL import. I spent a few hours pinning to this repository. I believe many could have encountered such issues trying to configure general logging, and gave up figuring it out.

Could we disable `coloredlogs` or at least make sure importing objects outside of CLI does not trigger its [installation on the root logger](https://coloredlogs.readthedocs.io/en/latest/readme.html#id6)?

Many thanks 🙏

bsbodden:
Fix: Prevent RedisVL from overriding existing logging configurations

## Issue
When importing any RedisVL component, the library currently overrides existing user-defined logging configurations silently. This happens because `coloredlogs.install()` is called during module import in `redisvl/utils/log.py`, which affects the root logger without the user's knowledge or consent.

## Changes
- Modified the `get_logger` function in `log.py` to respect existing logging configurations
- Added `reconfigure=False` to prevent modifying existing loggers
- Added `isatty=True` to only use colors in supported terminals
- Added a check to only configure loggers that don't already have handlers

## Testing
Replicates the before RedisVL is loaded scenario by using an external (to the test) py file that's run on it's own process

Fixes #293

### DIFF
--- a/redisvl/utils/log.py
+++ b/redisvl/utils/log.py
@@ -3,7 +3,6 @@ import sys
 
 import coloredlogs
 
-# constants for logging
 coloredlogs.DEFAULT_DATE_FORMAT = "%H:%M:%S"
 coloredlogs.DEFAULT_LOG_FORMAT = "%(asctime)s %(name)s %(levelname)s   %(message)s"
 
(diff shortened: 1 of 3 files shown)


rbs333 (approved):
LGTM

PR #302, opened 2025-03-25
Rename for test data class for pytest conflict

rbs333:


tests/integration/test_threshold_optimizer.py
```diff
@@ -111,7 +111,7 @@ def test_routes_different_distance_thresholds_optimizer_default(
 
     # now run optimizer
     router_optimizer = RouterThresholdOptimizer(router, test_data_optimization)
-    router_optimizer.optimize(max_iterations=10)
+    router_optimizer.optimize(max_iterations=10, search_step=0.5)
```
rbs333, line 114: Stop gap makes intermittent error damn near impossible. Also discovered helpful kwarg gap.

tylerhutcherson (approved):
LGTM -- only question is where else would we need to update TestData (examples / docs?). But likely something we can do after release. Maybe create a ticket for that now though so we don't forget!

bsbodden (commented):
LGTM!

PR #303, opened 2025-03-26
Add support for full text queries and hybrid search queries

justin-cechmanek:


redisvl/query/query.py
```diff
@@ -399,3 +403,123 @@ def params(self) -> Dict[str, Any]:
...
     # keep for backwards compatibility
     pass
+
+
+class TextQuery(FilterQuery):
+    def __init__(
+        self,
+        text: str,
+        text_field: str,
+        text_scorer: str = "BM25",
```
rbs333: default should be BM25STD

redisvl/query/query.py
```diff
@@ -399,3 +403,123 @@ def params(self) -> Dict[str, Any]:
...
+            in_order (bool): Requires the terms in the field to have
+                the same order as the terms in the query filter, regardless of
+                the offsets between them. Defaults to False.
+            params (Optional[Dict[str, Any]], optional): The parameters for the query.
+                Defaults to None.
+        """
+        import nltk
+        from nltk.corpus import stopwords
+
+        nltk.download("stopwords")
```
rbs333: Would this do the download on every instantiation of the class? Didn't think about before but might be worth caching somewhere. Also realizing now that stopwords should be a variable that a user can optionally supply. This would be necessary if working in a different language or more specific situation.

rbs333:
@justin-cechmanek you should rebase your branch and change this pr to be against `0.5.0`

redisvl/query/query.py
justin-cechmanek: nltk caches it locally, and yeah, making the language selectable, or uploading your own sounds like a good add.

tylerhutcherson: I think we should look again at Redis stopwords implementation. This is handled at the index level instead

https://redis.io/docs/latest/develop/interact/search-and-query/advanced-concepts/stopwords/#:~:text=Redis%20Stack%20has%20a%20default,are%20discarded%20and%20not%20indexed.

redisvl/query/query.py
```diff
@@ -647,3 +651,123 @@ def params(self) -> Dict[str, Any]:
...
+        params: Optional[Dict[str, Any]] = None,
+    ):
+        """A query for running a full text and vector search, along with an optional
+        filter expression.
+
+        Args:
+            text (str): The text string to perform the text search with.
+            text_field (str): The name of the document field to perform text search on.
+            text_scorer (str, optional): The text scoring algorithm to use.
+                Defaults to BM25. Options are {TFIDF, BM25, DOCNORM, DISMAX, DOCSCORE}.
```
tylerhutcherson: What about `BM25STD`

redisvl/query/query.py
tylerhutcherson: But maybe for a later optimization

redisvl/query/query.py
```diff
@@ -647,3 +651,123 @@ def params(self) -> Dict[str, Any]:
...
+
+
+class HybridQuery(AggregateRequest):
+    def __init__(
+        self, text_query: TextQuery, vector_query: VectorQuery, alpha: float = 0.7
+    ):
+        """An aggregate query for running a hybrid full text and vector search.
+
+        Args:
+            text_query (TextQuery): The text query to run text search with.
```
tylerhutcherson: I know we discussed this at some point, but is it too much complexity to ask a user to essentially define 3 different queries to do a simple hybrid search? there are also several params on both query objects that are not handled or used/considered by the downstream hybrid search implementation

redisvl/query/query.py
```diff
@@ -647,3 +651,123 @@ def params(self) -> Dict[str, Any]:
 class RangeQuery(VectorRangeQuery):
     # keep for backwards compatibility
     pass
+
+
+class TextQuery(FilterQuery):
```
tylerhutcherson: What's the benefit of wrapping filter query here? Also might advocate for `FullTextQuery` for clarity

redisvl/query/query.py
```diff
@@ -647,3 +651,123 @@ def params(self) -> Dict[str, Any]:
...
+        # Handle query modifiers
+        self.scorer(self._text_scorer)
+        self.paging(0, self._num_results).dialect(dialect)
+
+        if return_score:
+            self.with_scores()
+
+    def tokenize_and_escape_query(self, user_query: str) -> str:
+        """Convert a raw user query to a redis full text query joined by ORs"""
+        from redisvl.utils.token_escaper import TokenEscaper
```
tylerhutcherson: import at module level to avoid re import on query

tylerhutcherson (commented):
Left some comments nice start

redisvl/query/query.py
justin-cechmanek: If we wrap `BaseQuery` we don't inherit as much, like `return_fields`. It feel like `FullTextQuery` should come with all these options that `FilterQuery` has.

redisvl/query/query.py
justin-cechmanek: Since Hybrid search requires calling ft.aggregate with an AggregateRequest object we moved it into a separate class hierarchy and it won't build from a TextQuery now.

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,221 @@
...
+        vector: Union[bytes, List[float]],
+        vector_field: str,
+        text_scorer: str = "BM25STD",
+        filter_expression: Optional[Union[str, FilterExpression]] = None,
+        alpha: float = 0.7,
+        dtype: str = "float32",
+        num_results: int = 10,
+        return_fields: Optional[List[str]] = None,
+        stopwords: Optional[Union[str, Set[str]]] = "english",
+        dialect: int = 4,
```
rbs333: Default dialect should be 2 I believe.

tests/unit/test_aggregation_types.py
```diff
@@ -0,0 +1,115 @@
...
+from redis.commands.search.query import Query
+from redis.commands.search.result import Result
+
+from redisvl.index.index import process_results
+from redisvl.query.aggregate import HybridAggregationQuery
+from redisvl.query.filter import Tag
+
+# Sample data for testing
+sample_vector = [0.1, 0.2, 0.3, 0.4]
+sample_text = "the toon squad play basketball against a gang of aliens"
```
rbs333, line 12: space jam ❤️

redisvl/index/index.py
```diff
@@ -650,6 +686,44 @@ def fetch(self, id: str) -> Optional[Dict[str, Any]]:
             return convert_bytes(obj[0])
         return None
 
+    def aggregate_query(
```
tylerhutcherson: Nice! I would propose we consider consolidating this under `.query()` and have it check the type of the query and call the right internal method for search/aggregate as opposed to asking the user to call the search index searching method based on the type of query it is. cc @abrookins

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,220 @@
...
+        text_field_name: str,
+        vector: Union[bytes, List[float]],
+        vector_field_name: str,
+        text_scorer: str = "BM25STD",
+        filter_expression: Optional[Union[str, FilterExpression]] = None,
+        alpha: float = 0.7,
+        dtype: str = "float32",
+        num_results: int = 10,
+        return_fields: Optional[List[str]] = None,
+        stopwords: Optional[Union[str, Set[str]]] = "english",
```
tylerhutcherson, line 43: I think if we list this `Optional` the default would be `None` generally speaking. So if we want it to have a default, then we can treat it like other args that have a default like `num_results` or `alpha`!

tylerhutcherson: Maybe we want default behavior here to be `None` and we make the user declare if they want us to handle the stopwords client-side via redisvl

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,220 @@
...
+        super().__init__(query_string)
+
+        self.scorer(text_scorer)  # type: ignore[attr-defined]
+        self.add_scores()  # type: ignore[attr-defined]
+        self.apply(
+            vector_similarity=f"(2 - @{self.DISTANCE_ID})/2", text_score="@__score"
+        )
+        self.apply(hybrid_score=f"{1-alpha}*@text_score + {alpha}*@vector_similarity")
+        self.sort_by(Desc("@hybrid_score"), max=num_results)
+        self.dialect(dialect)  # type: ignore[attr-defined]
```
tylerhutcherson, line 123: AWESOME

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,220 @@
...
+            except Exception as e:
+                raise ValueError(f"Error trying to load {stopwords} from nltk. {e}")
+        elif isinstance(stopwords, (Set, List, Tuple)) and all(  # type: ignore
+            isinstance(word, str) for word in stopwords
+        ):
+            self._stopwords = set(stopwords)
+        else:
+            raise TypeError("stopwords must be a set, list, or tuple of strings")
+
+    def tokenize_and_escape_query(self, user_query: str) -> str:
```
tylerhutcherson: should probably be a private method

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,220 @@
...
+
+    def _build_query_string(self) -> str:
+        """Build the full query string for text search with optional filtering."""
+        if isinstance(self._filter_expression, FilterExpression):
+            filter_expression = str(self._filter_expression)
+        else:
+            filter_expression = ""
+
+        # base KNN query
+        knn_query = f"KNN {self._num_results} @{self._vector_field} ${self.VECTOR_PARAM} AS {self.DISTANCE_ID}"
```
tylerhutcherson, line 211: eventually we should probably create some testable helper utils that make query strings for different parts

redisvl/query/query.py
```diff
@@ -681,3 +685,138 @@ def params(self) -> Dict[str, Any]:
...
+                their own stop words by providing a List or Set of words. if set to None,
+                then no words will be removed. Defaults to 'english'.
+
+        Raises:
+            ValueError: if stopwords language string cannot be loaded.
+            TypeError: If stopwords is not a valid iterable set of strings.
+        """
+        self._text = text
+        self._text_field = text_field_name
+        self._text_scorer = text_scorer
```
tylerhutcherson: Do we need to set this scorer anywhere on the query??

redisvl/index/index.py
```diff
@@ -846,7 +819,7 @@ def _query(self, query: BaseQuery) -> List[Dict[str, Any]]:
         results = self.search(query.query, query_params=query.params)
         return process_results(results, query=query, schema=self.schema)
 
-    def query(self, query: BaseQuery) -> List[Dict[str, Any]]:
+    def query(self, query: Union[BaseQuery, AggregationQuery]) -> List[Dict[str, Any]]:
```
tylerhutcherson, line 822: Update docstring too

redisvl/query/aggregate.py
```diff
@@ -45,7 +45,7 @@ def __init__(
         dialect: int = 2,
     ):
         """
-        Instantiages a HybridAggregationQuery object.
+        Instantiages a HybridQuery object.
```
tylerhutcherson: Typo

tylerhutcherson (approved):

redisvl/query/aggregate.py
```diff
@@ -0,0 +1,219 @@
...
+import nltk
+from nltk.corpus import stopwords as nltk_stopwords
+from redis.commands.search.aggregation import AggregateRequest, Desc
+
+from redisvl.query.filter import FilterExpression
+from redisvl.redis.utils import array_to_buffer
+from redisvl.utils.token_escaper import TokenEscaper
+
+
+# base class
```
abrookins:
```suggestion

```

redisvl/query/query.py
```diff
@@ -681,3 +687,170 @@ def params(self) -> Dict[str, Any]:
...
+        return_fields: Optional[List[str]] = None,
+        num_results: int = 10,
+        return_score: bool = True,
+        dialect: int = 2,
+        sort_by: Optional[str] = None,
+        in_order: bool = False,
+        params: Optional[Dict[str, Any]] = None,
+        stopwords: Optional[Union[str, Set[str]]] = "english",
+    ):
+        """A query for running a full text and vector search, along with an optional
```
abrookins: Does this really do vector search? 🤔

abrookins (approved):
Looks good! Couple of small comments, but 👍

PR #304, opened 2025-03-26
Support client-side schema validation using Pydantic

tylerhutcherson:
This PR implements a layered architecture for managing and validating searchable data in Redis, with clear separation of concerns between schema definition, data validation, and storage operations.

### 1. Schema Definition Layer
- `IndexSchema` provides the blueprint for data structure and constraints
    - Defines fields with specific types (TEXT, TAG, NUMERIC, GEO, VECTOR)
    - Supports different storage types (HASH, JSON) with appropriate configuration

### 2. Validation Layer
- `SchemaModelGenerator` dynamically creates Pydantic models from schema definitions
- Implements a caching mechanism to avoid redundant model generation
- Maps Redis field types to appropriate Python/Pydantic types
- Provides type-specific validators:
    - VECTOR: validates dimensions and value ranges (e.g., INT8 range checks)
    - GEO: validates geographic coordinate format
    - NUMERIC: prevents boolean values

### 3. Storage Layer
- `BaseStorage` is the abstract class provides the foundation for Redis operations
- Specialized implementations (HashStorage, JsonStorage) for different Redis data types
- Enforces schema validation during write operations when set to True
- Implements optimized batch operations using Redis pipelines
- Supports both synchronous and asynchronous interfaces
- Handles key generation, preprocessing, and error handling

### 4. Index Layer
The `SearchIndex` contains the setting `validate_on_load`, which defaults on `False`.

### Write Flow:
Objects are preprocessed and validated against the schema
Objects are prepared with appropriate keys
Batch writing occurs using Redis pipelines for efficiency
TTL (expiration) can be applied if specified
### Read Flow:
Keys are fetched in batches using pipelines
Data is converted from Redis format to Python objects
Bytes are automatically converted to appropriate types


redisvl/schema/validation.py
```diff
@@ -0,0 +1,275 @@
...
+                    VectorDataType.INT8,
+                    VectorDataType.UINT8,
+                ):
+                    return List[int]
+                return List[float]
+            else:
+                return bytes
+
+        # If we get here, the field type is not supported
+        raise ValueError(f"Unsupported field type: {field.type}")
```
abrookins, line 91: Alternatively, we could log a warning and give the field an Any type to skip it during validation, or else log then ignore it because we're using extra=True later.

redisvl/schema/validation.py
```diff
@@ -0,0 +1,275 @@
...
+        Get or create a Pydantic model for a schema.
+
+        Args:
+            schema: The IndexSchema to convert to a Pydantic model
+
+        Returns:
+            A Pydantic model class that can validate data against the schema
+        """
+        # Use schema identifier as cache key
+        cache_key = schema.index.name
```
abrookins: Slightly worried about this because, while it seems unlikely, two indexes could share the same name in some circumstances (different redis instances, etc.). Would it work equally well to take the hash() of the schema class? There should only be one of those in the process and it will certainly match.

redisvl/schema/validation.py
```diff
@@ -0,0 +1,275 @@
...
+                "model_config": {"arbitrary_types_allowed": True, "extra": "allow"},
+            }
+        )
+
+        # Create the model class using type()
+        model_name = f"{schema.index.name}__PydanticModel"
+        return type(model_name, (BaseModel,), class_dict)
+
+
+def extract_from_json_path(obj: Dict[str, Any], path: str) -> Any:
```
abrookins, line 213: Hmmmm. I'd use a formal JSONPath parser here.

abrookins (commented):
This is EPIC! Looks great, too. Love the validation test coverage. I had a couple of comments, and probably the only thing I felt strongly about was using an actual JSONPath library for traversing objects.

redisvl/schema/validation.py
tylerhutcherson: Totally could! By this point, only fields logged in the schema are being validated. So if it's not an approved redis type for search then it wouldn't even get to this point

redisvl/schema/validation.py
tylerhutcherson: Makes sense

redisvl/schema/validation.py
tylerhutcherson: Yeah fair, I've not used the lib much. If you have cycles today to timebox this, I would welcome that contribution!

abrookins (approved):
Looks great! 👍

PR #305, opened 2025-03-28
Add batch_search to sync Index

abrookins:
Add a `batch_search` method to `SearchIndex` and `AsyncSearchIndex`. These methods run search commands in a pipeline and return correctly-parsed and post-processed results.

TODO:
- [x] Async version
- [x] `batch_query`

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -1283,8 +1283,8 @@ class AsyncSearchIndex(BaseSearchIndex):
             raise RedisSearchError(f"Error while aggregating: {str(e)}") from e
 
     async def batch_search(
-        self, queries: List[str], batch_size: int = 100, **query_params
-    ) -> List[List[Dict[str, Any]]]:
+        self, queries: List[BaseQuery], batch_size: int = 100, **query_params
+    ) -> List["Result"]:
         """Perform a search against the index for multiple queries.
 
         This method takes a list of queries and returns a list of search results.
@@ -1298,9 +1298,9 @@ class AsyncSearchIndex(BaseSearchIndex):
                 for each query.
 
         Returns:
-            List[List[Dict[str, Any]]]: The search results.
+            List[Result]: The search results.
         """
-        all_parsed = []
+        all_results = []
         client = await self._get_client()
         search = client.ft(self.schema.index.name)
         options = {}
@@ -1340,16 +1340,8 @@ class AsyncSearchIndex(BaseSearchIndex):
                         query=_built_query,
                         duration=duration,
                     )
-                    parsed = process_results(
-                        parsed_raw,
-                        query=_built_query,
-                        storage_type=self.schema.index.storage_type,
-                    )
-                    # Create separate lists of parsed results for each query
-                    # passed in to the batch_search method, so that callers can
-                    # access the results for each query individually
-                    all_parsed.append(parsed)
-        return all_parsed
+                    all_results.append(parsed_raw)
+        return all_results
 
     async def search(self, *args, **kwargs) -> "Result":
         """Perform a search on this index.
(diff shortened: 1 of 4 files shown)


redisvl/index/index.py
```diff
@@ -652,6 +655,74 @@ def aggregate(self, *args, **kwargs) -> "AggregateResult":
...
+    ) -> List[List[Dict[str, Any]]]:
+        """Perform a search against the index for multiple queries.
+
+        This method takes a list of queries and returns a list of search results.
+        The results are returned in the same order as the queries.
+
+        Args:
+            queries (List[str]): The queries to search for.
+            batch_size (int, optional): The number of queries to search for at a time.
+                Defaults to 100.
```
tylerhutcherson: My gut tells me we should lower this to like 10. I am thinking about RAM available on shards and the payload size that will come back from searches. Don't have any hard data on this but for queries I think we might want to go a bit smaller with the ability for users to try larger at their own risk

redisvl/index/index.py
```diff
@@ -652,6 +655,74 @@ def aggregate(self, *args, **kwargs) -> "AggregateResult":
...
+            # so we need to manually execute each command in a pipeline
+            # and parse the results
+            with self._redis_client.pipeline(transaction=False) as pipe:
+                batch_built_queries = []
+                for query in batch_queries:
+                    query_args, q = search._mk_query_args(  # type: ignore
+                        query, query_params=query_params
+                    )
+                    batch_built_queries.append(q)
+                    pipe.execute_command(
```
tylerhutcherson, line 710: oh what, for real? I thought we just do `pipe.ft(<index-name>).search(...)` I swear I have code somewhere that does this

redisvl/index/index.py
```diff
@@ -652,6 +655,74 @@ def aggregate(self, *args, **kwargs) -> "AggregateResult":
...
+                st = time.time()
+                results = pipe.execute()
+
+                # We don't know how long each query took, so we'll use the total time
+                # for all queries in the batch as the duration for each query
+                duration = (time.time() - st) * 1000.0
+
+                for i, query_results in enumerate(results):
+                    _built_query = batch_built_queries[i]
+                    parsed_raw = search._parse_search(  # type: ignore
```
tylerhutcherson: Smart to tap into the search parsing capabilities baked into redis py :)

tylerhutcherson (approved):
This is awesome. Once you get the sync version of `batch_query` in, let's merge. One question about batch size defaults and pipeline, but otherwise amazing!

redisvl/index/index.py
abrookins: Totally agree 👍

redisvl/index/index.py
abrookins: Uhhhh maybe I did it wrong 😂

abrookins: Well look at that, `pipe.ft().search()` works just fine. This is a much better outcome.

abrookins: I spoke too soon. `search` appears to be broken for pipelines. Oh well!
```

async_index = <redisvl.index.index.AsyncSearchIndex object at 0x10ba26c50>

    @pytest.mark.asyncio
    async def test_batch_search_with_multiple_batches(async_index):
        await async_index.create(overwrite=True, drop=True)
        data = [{"id": "1", "test": "foo"}, {"id": "2", "test": "bar"}]
        await async_index.load(data, id_field="id")
    
>       results = await async_index.batch_search(
            [FilterQuery("@test:{foo}"), FilterQuery("@test:{bar}")],
            batch_size=2,

redisvl/index/index.py:1311: in batch_search
    await pipe.ft(self.schema.index.name).search(  # type: ignore
env/lib/python3.11/site-packages/redis/commands/search/commands.py:945: in search
    return self._parse_results(
env/lib/python3.11/site-packages/redis/commands/search/commands.py:72: in _parse_results
    return self._RESP2_MODULE_CALLBACKS[cmd](res, **kwargs)
env/lib/python3.11/site-packages/redis/commands/search/commands.py:79: in _parse_search
    return Result(
env/lib/python3.11/site-packages/redis/commands/search/result.py:29: in __init__
    self.total = res[0]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <redis.asyncio.client.Pipeline(<redis.asyncio.connection.ConnectionPool(<redis.asyncio.connection.Connection(host=0.0.0.0,port=52824,db=0)>)>)>
name = 0

    def __getitem__(self, name: KeyT):
>       raise TypeError("Async Redis client does not support class retrieval")
E       TypeError: Async Redis client does not support class retrieval

env/lib/python3.11/site-packages/redis/commands/core.py:2551: TypeError
```

PR #307, opened 2025-04-01
update search step on recall and precision

rbs333:
We were experiencing flakiness with a similar test before and haven't had an issue (as far as I know) with that test since updating the search_step. 

Error occurs for this step when after 10 tries the random search hasn't made the threshold large enough to show improvement which is why it failed sometimes.


abrookins (approved):
LGTM 👍

PR #314, opened 2025-04-07
Fix HybridQuery class description

### LINKED ISSUE #313: Incosistency in HybridQuery class description
We're generating markdown docs using a Sphinx extension.

I noticed that the `.. code-block:: python` bit has a leading space. This results in an incorrectly formatted markdown file. 

https://github.com/redis/redis-vl-python/blob/4154aae400d029f236c880a2e5ea2d77885956c2/redisvl/query/aggregate.py#L19-L25

compared to the correct:

https://github.com/redis/redis-vl-python/blob/4154aae400d029f236c880a2e5ea2d77885956c2/redisvl/query/query.py#L690-L694

paololazzari:
Fixes #313


rbs333 (approved):
thanks!

PR #316, opened 2025-04-09
Add a drop_documents() method to search index classes

abrookins:
Wrap key creation so users don't have to call private methods.


tylerhutcherson (approved):
Lovely!

PR #317, opened 2025-04-09
Add support for EF_RUNTIME param

abrookins:
Add support for EF_RUNTIME param at query time for HNSW vector queries.


redisvl/query/query.py
```diff
@@ -308,6 +317,10 @@ def _build_query_string(self) -> str:
             if self._hybrid_policy == HybridPolicy.BATCHES and self._batch_size:
                 knn_query += f" BATCH_SIZE {self._batch_size}"
 
+        # Add EF_RUNTIME parameter if specified
+        if self._ef_runtime:
+            knn_query += f" EF_RUNTIME {self._ef_runtime}"
```
tylerhutcherson: You can use the `self.EF_RUNTIME_PARAM` var here right?

tylerhutcherson (approved):
Solid. Question about two things:

1) looks like the way our query constructors are moving, each special param we add calls `_build_query_string` again. I understand why in the current setup, but surely on init we don't want to call that many many times. Will be slower than we want. Might be a bit more refactor there we could do here or another PR if we want.

2) Can you clarify the approach taken between using query params and direct query string manipulation with values?

abrookins:
This is currently stalled because integration tests showed the EF_RUNTIME option doesn't seem to work like the docs say, or else I've interpreted them wrong. (Syntax errors from Redis.)

tests/integration/test_query.py
```diff
@@ -165,7 +165,7 @@ def index(sample_data, redis_url):
                     "attrs": {
                         "dims": 3,
                         "distance_metric": "cosine",
-                        "algorithm": "flat",
```
rbs333, line 168: @abrookins, catching up to this, didn't see a test for it (but may have missed it) what happens if algorithm is `flat` and you set ef_runtime? Does it just do nothing

PR #318, opened 2025-04-11
Add embeddings cache

tylerhutcherson:
Initial design and refactor here. 
- Includes some clean up for the internals of the OG `SemanticCache`.
- Refactors the core of the cache classes to use `BaseCache` for both `EmbeddingsCache` and `SemanticCache`.
- Updates documentation, working tests, and user guides.

NOTE: Will hold off on integrating with our vectorizers for a separate PR. This one is already long.


redisvl/extensions/cache/base.py
```diff
@@ -0,0 +1,222 @@
...
+            self._owns_redis_client = True
+            self._redis_client = None  # type: ignore
+
+    def _get_prefix(self) -> str:
+        """Get the key prefix for Redis keys.
+
+        Returns:
+            str: The prefix to use for Redis keys.
+        """
+        return f"{self.name}:"
```
justin-cechmanek, line 69: Do we need this as a method instead of just including `self.name` where it's needed elsewhere?

justin-cechmanek (approved):
Looks good! Just have the one question, but it's not blocking.

abrookins (approved):
This makes sense to me! 👍

redisvl/extensions/cache/base.py
tylerhutcherson: Encapsulating the prefix definition like this allows me to unit test it

PR #319, opened 2025-04-15
Validate EF_RUNTIME query parameter

abrookins:
The EF_RUNTIME parameter is only valid in queries that target vector fields using the HNSW algorithm. If you run the same query against a 'flat' vector field, Redis will return an error. Because we have all the information to prevent you from running the bad query, we can introduce validation that raises a helpful error instead of relying on Redis error handling.


.cursor/rules/redisvl.mdc
```diff
@@ -0,0 +1,11 @@
+---
+description: 
+globs: 
+alwaysApply: true
```
tylerhutcherson, line 4: This file meant to be in this PR?

redisvl/index/index.py
```diff
@@ -814,8 +837,21 @@ def batch_query(
             all_parsed.append(parsed)
         return all_parsed
 
+    def _validate_query(self, query: BaseQuery) -> None:
```
tylerhutcherson: Could probably be moved to the base search index class since it's shared?

redisvl/index/index.py
```diff
@@ -814,8 +837,21 @@ def batch_query(
...
+        """Validate a query."""
+        if isinstance(query, VectorQuery):
+            field = self.schema.fields[query._vector_field_name]
+            if query.ef_runtime and field.attrs.algorithm != "hnsw":  # type: ignore
```
tylerhutcherson: I think there's an enum for this

tylerhutcherson (approved):

tylerhutcherson:
Looks great left two no big deal suggestions

redisvl/index/index.py
abrookins: I'll find that. Also, this error message must have been an autocomplete.

redisvl/index/index.py
abrookins: Agreed!

.cursor/rules/redisvl.mdc
abrookins: Actually, yeah, I lumped it in because Cursor was driving me crazy "fixing" that line over and over.

PR #322, opened 2025-04-17
Enable semantic router reference updates

rbs333:
This pr add methods so that you can easily get/add/delete route references.

### DIFF
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -101,6 +101,8 @@ class SemanticRouter(BaseModel):
             routes=routes,
             vectorizer=vectorizer,
             routing_config=routing_config,
+            redis_url=redis_url,
+            redis_client=redis_client,
         )
 
         self._initialize_index(redis_client, redis_url, overwrite, **connection_kwargs)
@@ -130,7 +132,9 @@ class SemanticRouter(BaseModel):
             )
 
         router_dict = redis_client.json().get(f"{name}:route_config")  # type: ignore
-        return cls.from_dict(router_dict)
+        return cls.from_dict(
+            router_dict, redis_url=redis_url, redis_client=redis_client
+        )
 
     @deprecated_argument("dtype")
     def _initialize_index(
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -8,6 +8,7 @@ from redis import Redis
 from redis.commands.search.aggregation import AggregateRequest, AggregateResult, Reducer
 from redis.exceptions import ResponseError
 
+from redisvl.exceptions import RedisModuleVersionError
 from redisvl.extensions.constants import ROUTE_VECTOR_FIELD_NAME
 from redisvl.extensions.router.schema import (
     DistanceAggregationMethod,
@@ -19,6 +20,7 @@ from redisvl.extensions.router.schema import (
 from redisvl.index import SearchIndex
 from redisvl.query import FilterQuery, VectorRangeQuery
 from redisvl.query.filter import Tag
+from redisvl.redis.connection import RedisConnectionFactory
 from redisvl.redis.utils import convert_bytes, hashify, make_dict
 from redisvl.utils.log import get_logger
 from redisvl.utils.utils import deprecated_argument, model_to_dict
@@ -41,9 +43,7 @@ class SemanticRouter(BaseModel):
     """Configuration for routing behavior."""
 
     _index: SearchIndex = PrivateAttr()
-    _vectorizer: ClassVar[Union[BaseVectorizer, None]] = None
-    _name: ClassVar[str] = ""
-    _cls_index: ClassVar[Union[SearchIndex, None]] = None
+    _persist_config: bool = PrivateAttr()
 
     model_config = ConfigDict(arbitrary_types_allowed=True)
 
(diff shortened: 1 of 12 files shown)


redisvl/extensions/router/semantic.py
```diff
@@ -141,6 +158,26 @@ def _initialize_index(
...
+        schema = SemanticRouterIndexSchema.from_params(
+            router_name, vectorizer.dims, vectorizer.dtype  # type: ignore
+        )
+
+        return SearchIndex(
+            schema=schema,
+            redis_client=redis_client,
+            redis_url=redis_url,
+            **connection_kwargs,
+        )
```
justin-cechmanek: Why do we need to pull this out as a separate method when we only execute this code within _initialize_index()?

redisvl/extensions/router/semantic.py
```diff
@@ -597,3 +638,196 @@ def to_yaml(self, file_path: str, overwrite: bool = True) -> None:
...
+    @classmethod
+    def delete_route_references(
+        cls,
+        route_name: str,
+        router_name: str = "",
+        reference_ids: List[str] = [],
+        redis_client: Optional[Redis] = None,
+        redis_url: str = "",
+    ) -> int:
+        """Get references for an existing route route.
```
justin-cechmanek:
```suggestion
        """Get references for an existing router route.
```

justin-cechmanek (commented):
Looks like some of the cells in the 08_semantic_router notebook are returning no routes. Can we double check this

redisvl/extensions/router/semantic.py
rbs333: Good catch. Initially I was going to make broader use of this function but then changed course. Will revert.

redisvl/extensions/router/semantic.py
```diff
@@ -596,3 +620,200 @@ def to_yaml(self, file_path: str, overwrite: bool = True) -> None:
...
+                    "If class not instantiated, must provide redis_client or redis_url and router_name"
+                )
+        else:
+            index = cls._cls_index
+            router_name = cls._name
+
+        if reference_ids:
+            queries = cls._make_filter_queries(reference_ids)
+        else:
+            _, keys = index.client.scan(
```
rbs333: Open to suggestions if there is a better way to get all keys within a prefix.

tests/integration/test_semantic_router.py
```diff
@@ -41,13 +41,14 @@ def routes():
 @pytest.fixture
 def semantic_router(client, routes):
     router = SemanticRouter(
-        name="test-router",
+        name=f"test-router-{str(ULID())}",
```
rbs333, line 44: Adding this here to help prevent collisions within test execution. 

Generally to improve our test consistency I think it would be good practice to make all function level fixtures (the default) as idempotent as possible. Given we execute tests with multiple workers in an async fashion that might help.

redisvl/extensions/router/semantic.py
```diff
@@ -596,3 +620,200 @@ def to_yaml(self, file_path: str, overwrite: bool = True) -> None:
...
+            router_name = cls._name
+
+        route_references: List[Dict[str, Any]] = []
+        keys: List[str] = []
+
+        # embed route references as a single batch
+        reference_vectors = vectorizer.embed_many(references, as_buffer=True)
+
+        # set route references
+        for i, reference in enumerate(references):
```
tylerhutcherson, line 667: isn't this implementing the same or similar logic as `_add_routes(...)`?

redisvl/extensions/router/semantic.py
```diff
@@ -596,3 +620,200 @@ def to_yaml(self, file_path: str, overwrite: bool = True) -> None:
         with open(fp, "w") as f:
             yaml_data = self.to_dict()
             yaml.dump(yaml_data, f, sort_keys=False)
+
+    # reference methods
+
+    @classmethod
+    def add_route_references(
```
tylerhutcherson, line 642: Only question is an ergonomic one.. what would be the benefit of using `add_route_references(...)` on a non-instantiated router class, wherein you will still end up passing through the args here to essentially init the class? Isn't it effectively the same as calling `router = SemanticRouter(...); router.add_route_references(...)`? 

I know this is/was one of the tricky and sticky points, so just curious to hear what you're thinking! :)

tylerhutcherson (approved):
Nice! Left a question for ya

redisvl/extensions/router/semantic.py
rbs333: The key difference here is that to you don't need to pass a list of routes when using the class method. You would have to provide the router name and redis_url but theoretically if you were unaware of the initial routes it wouldn't matter. I think there is some value in not having to know the previous routes.

That said, I'm open to removing because it makes the code more direct. A better way to handle could be to relax the requirement on passing routes in the init.

redisvl/extensions/router/semantic.py
rbs333: Yes. The gotcha here is the route reference one takes advantage of `self.` and this one is based on `cls.` I decided against "functionizing" since there wasn't a 3rd case. This is an example of where the code becomes cleaner if we drop the class method idea (like we had talked about at one point).

redisvl/extensions/router/semantic.py
tylerhutcherson: Ahh yeah, I see what you mean! 

Core issue is that class init is pretty rigid and expects certain inputs. It was probably a gap in the initial interface design for sure on my part.

PR #324, opened 2025-04-23
Fix ranx numba type warnings

rbs333:


tylerhutcherson:
@rbs333 looks like a rebase is needed with your latest commits (for this branch) replayed on top of `0.6.0`

PR #325, opened 2025-04-23
Add cache option to vectorizer from dict

rbs333:


redisvl/utils/vectorize/__init__.py
```diff
@@ -23,22 +27,33 @@
...
+    cache_folder=os.getenv("SENTENCE_TRANSFORMERS_HOME"),
+) -> BaseVectorizer:
     vectorizer_type = Vectorizers(vectorizer["type"])
     model = vectorizer["model"]
+
+    args = {"model": model}
+    if cache:
+        emb_cache = EmbeddingsCache(**cache)
+        args["cache"] = emb_cache
+        args["cache_folder"] = cache_folder
```
tylerhutcherson: I don't think we need to explicitly call out the cache folder as that is just a `kwarg` for the `HFTextVectorizer` and would break other vectorizers. So I would just look at supporting arbitrary `**kwargs` here so that anything beyond defined arguments are passed as kwargs to the vectorizer init.

PR #326, opened 2025-04-23
Changes default cache embedding model

justin-cechmanek:
Changes the default semantic cache embedding model to our fine tuned model.

### DIFF
--- a/redisvl/extensions/cache/llm/semantic.py
+++ b/redisvl/extensions/cache/llm/semantic.py
@@ -111,8 +111,19 @@ class SemanticCache(BaseLLMCache):
             if dtype:
                 vectorizer_kwargs.update(dtype=dtype)
 
+            # raise a warning to inform users we changed the default model
+            # remove this warning in future releases
+            logger.warning(
+                "The default vectorizer has changed from `sentence-transformers/all-mpnet-base-v2` "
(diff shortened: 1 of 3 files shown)


redisvl/extensions/cache/llm/semantic.py
```diff
@@ -111,8 +111,19 @@ def __init__(
             if dtype:
                 vectorizer_kwargs.update(dtype=dtype)
 
+            # raise a warning to inform users we changed the default model
+            # remove this warning in future releases
+            logger.warning(
```
tylerhutcherson, line 116: With the `warning` below in the case of pre-existing index AND overwrite=True, I don't think we actually need this one here right?

tylerhutcherson (approved):
LGTM! one comment pending

redisvl/extensions/cache/llm/semantic.py
justin-cechmanek: You're right that a mismatch won't happen in that case, but there still is a behavior change when we switch default models. It's why I had to update the tests because the embedding distances changed.

justin-cechmanek: If someone goes from 0.5 to 0.6 and never specifies a vectorizer, and runs the same script start to finish - writes/reads/clears - they may get different results.

PR #327, opened 2025-04-24
Semantic router testing name generation

rbs333:


tylerhutcherson (approved):

tests/integration/test_threshold_optimizer.py
```diff
@@ -35,13 +37,13 @@ def routes():
 
 
 @pytest.fixture
-def semantic_router(client, routes, hf_vectorizer):
```
rbs333, line 38: test was failing in conftest before evening making it to check the db version

PR #328, opened 2025-04-26
Flaky test fixes

abrookins:
- Centralize HG vectorizers
- Use worker ID in search index names and keys


tylerhutcherson:
Looks like some index naming checks fail now? Otherwise LGtM!

PR #333, opened 2025-04-30
Reruns notebooks with new default semantic cache embedding model

justin-cechmanek:


abrookins (approved):
LGTM! Some of the output suggests we need to revisit how we're building tables now.

PR #338, opened 2025-05-21
Fix various commands to work with Redis Cluster

abrookins:
## Description

This PR introduces **Redis Cluster support** throughout the library by updating internal client abstractions and connection logic. It ensures compatibility with both standalone and clustered Redis deployments.

### Key Changes

- **Redis client compatibility**
  - Upgraded `redis` from `5.2.1` → `6.1.0`, **keeping backwards-compatible support** for 5.x series
  - Replaced explicit `redis.Redis`/`redis.asyncio.Redis` types with union types (`SyncRedisClient`, `AsyncRedisClient`, etc.) in `redisvl.types`
  - Updated all cache, index, and router components to support both standalone and cluster Redis clients (sync and async)

- **Indexing improvements**
  - Cluster-safe implementations for `FT.CREATE`, `FT.SEARCH`, and `FT.DROPINDEX` (supposed to be supported, but redis-py 6.1 still isn't sending these to the default node...)
  - Introduced `cluster_create_index` and `cluster_search` utility functions to help fill these client gaps

- **Connection handling**
  - Refactored `RedisConnectionFactory` to detect and create cluster clients automatically
  - Improved sync-to-async conversion with type checks -- converting a sync Cluster client to async isn't possible because the connection pool works differently

- **Test infrastructure**
  - Added Docker Compose-based Redis Cluster setup for integration tests (`tests/cluster-compose.yml`)
  - New fixtures: `redis_cluster_url`, `cluster_client`
  - New tests validating cluster behavior (`test_redis_cluster_support.py`)
  - Updated tests that were creating new vectorizers unnecessarily to use the existing session-scoped fixtures (cut down API traffic to HF)

- **Other updates**
  - Added `tests/data` to `.gitignore`
  - Removed `types-redis` because its outdated type annotations cause erroneous lint errors
  - Improved TTL and async client cleanup logic
  - Cleaned up types, error handling, and inline documentation

## Gotchas
- I could not get Redis Cluster to come up in CI, only locally, and couldn't determine the problem from stdout/stderr, logs, etc.
- To unblock other work, I've flagged the cluster tests so they run separately, with `--run-cluster-tests`.
- This means Redis Cluster tests are currently turned off in CI

## Why

Supporting Redis Cluster enables RedisVL to operate in **high-availability, horizontally-scalable environments**, making it production-ready for distributed Redis deployments.

This also unblocks us from allowing downstream libraries, such as langgraph-redis, to support Redis Cluster.

## Breaking Changes

- **Dropped**: `types-redis` (not compatible with updated `redis` package)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -90,7 +90,7 @@ jobs:
         python-version: ["3.9", "3.10", 3.12, 3.13] 
         connection: ["hiredis", "plain"]
         redis-py-version: ["5.x", "6.x"]
-        redis-version: ["6.2.6-v9", "latest", "8.0-M03"]
+        redis-version: ["6.2.6-v9", "latest", "8.0.1"]
 
     steps:
       - name: Check out repository
@@ -132,7 +132,7 @@ jobs:
 
       - name: Set Redis image name
         run: |
-          if [[ "${{ matrix.redis-version }}" == "8.0-M03" ]]; then
+          if [[ "${{ matrix.redis-version }}" == "8.0.1" ]]; then
             echo "REDIS_IMAGE=redis:${{ matrix.redis-version }}" >> $GITHUB_ENV
           else
             echo "REDIS_IMAGE=redis/redis-stack-server:${{ matrix.redis-version }}" >> $GITHUB_ENV
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -78,7 +78,7 @@ jobs:
           make test-all
 
   test:
-    name: Python ${{ matrix.python-version }} - ${{ matrix.connection }} [redis ${{ matrix.redis-version }}]
+    name: Python ${{ matrix.python-version }} - ${{ matrix.connection }} - redis-py ${{ matrix.redis-py-version }} [redis ${{ matrix.redis-version }}]
     runs-on: ubuntu-latest
     needs: service-tests
     env:
@@ -89,6 +89,7 @@ jobs:
         # 3.11 tests are run in the service-tests job
         python-version: ["3.9", "3.10", 3.12, 3.13] 
         connection: ["hiredis", "plain"]
+        redis-py-version: ["5.x", "6.x"]
         redis-version: ["6.2.6-v9", "latest", "8.0-M03"]
 
     steps:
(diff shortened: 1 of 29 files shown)


tests/cluster-compose.yml
```diff
@@ -0,0 +1,199 @@
...
+      - redis-node-2
+      - redis-node-3
+      - redis-node-4
+      - redis-node-5
+      - redis-node-6
+    entrypoint:
+      - sh
+      - -c
+      - |
+        echo "Waiting for Redis nodes to be available..."
```
abrookins, line 129: This is a little much, but it was the result of many iterations on cluster startup failures. I may simplify!

tests/conftest.py
```diff
@@ -52,6 +54,75 @@ def redis_container(worker_id):
...
+                project_name,
+                "up",
+                "--wait",  # Wait for healthchecks
+                "-d",  # Detach
+            ],
+            check=True,  # check=True can be problematic if successful commands return non-zero, but `up --wait -d` should be 0 on success.
+        )
+        yield
+    except subprocess.CalledProcessError as e:
+        # Attempt to get logs if setup failed
```
abrookins: Again, multiple iterations on repeated startup failure 😬

tests/conftest.py
```diff
@@ -62,6 +133,12 @@ def redis_url(redis_container):
     return f"redis://{host}:{port}"
 
 
+@pytest.fixture(scope="session")
+def redis_cluster_url(redis_cluster_container):
+    # Hard-coded due to Docker issues
+    return "redis://localhost:7001"
```
abrookins, line 175: Currently hard-coded due to issues obtaining the IP at runtime.

tests/conftest.py
```diff
@@ -52,6 +54,75 @@ def redis_container(worker_id):
...
+                "-f",
+                compose_file,
+                "-p",
+                project_name,
+                "down",
+                "-v",  # Remove volumes
+            ]
+        )
+
+
```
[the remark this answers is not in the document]
abrookins: `DockerCompose` didn't work here because commands with multiple services can exit with non-zero status codes despite succeeding.

.github/workflows/test.yml
```diff
@@ -89,6 +89,7 @@ jobs:
         # 3.11 tests are run in the service-tests job
         python-version: ["3.9", "3.10", 3.12, 3.13] 
         connection: ["hiredis", "plain"]
+        redis-py-version: ["5.x", "6.x"]
         redis-version: ["6.2.6-v9", "latest", "8.0-M03"]
```
tylerhutcherson: Can we also update to the latest Redis docker container while we're here?

tylerhutcherson (approved):
Huge effort!!

PR #341, opened 2025-05-27
Add index empty and missing support

tylerhutcherson:
PR implements support to index missing fields as well as empty fields.


abrookins:
@claude where you at

PR #355, opened 2025-06-19
Migrate from poetry to uv

tylerhutcherson:
Beyond poetry dependency resolution being the bane of my existence... `uv` is just better. **Faster, stronger..**

This PR introduces `uv` as our new package manager and greatly improves the economics of our testing CI workflow on a few fronts:
- Updated versions of common actions
- DRY(er) environment var setup
- Isolated dependency installs during CI tests
- Caching enabled for all of the things


redisvl/exceptions.py
```diff
@@ -27,7 +27,8 @@ class SchemaValidationError(RedisVLError):
     """Error when validating data against a schema."""
 
     def __init__(self, message, index=None):
-        if index is not None:
+        # Only add index prefix if the message doesn't already contain detailed validation info
```
rbs333, line 30: I'm not sure how this is related to this pr

CONTRIBUTING.md
rbs333, line 1: Good improvements here

Makefile
rbs333, line 1: Same some nice adds

rbs333 (approved):
I like it

- Pulled local uv sync worked for me
- Make additions worked
- Pre-commit also working for local commit
- Generally faster seems like an improvement

redisvl/exceptions.py
tylerhutcherson: Ahh yeah when I reran the notebooks, the messages for schema validation warnings were driving me insane. Meant to pull that into a separate commit/branch. I can remove.

rbs333: It can stay in just wasn't sure if you wanted it as part of this one or not :)

.github/workflows/test.yml
```diff
@@ -38,20 +37,24 @@ jobs:
...
+      - name: Install Python
+        uses: actions/setup-python@v5
         with:
           python-version: ${{ env.PYTHON_VERSION }}
-          cache: pip
 
-      - name: Install Poetry
-        uses: snok/install-poetry@v1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
abrookins, line 46: We don't have to do this right now, but Astral publishes Docker images with uv already installed in the image, and they're Python version specific, so we could use those as a base image for every test we've tested in the matrix, assuming they have a base image for that version. And that would probably cut down a lot of install time.

tylerhutcherson: Will take a look and just give it a try :)

tylerhutcherson: Ran into a few issues. Going to create a ticket.

PR #367, opened 2025-08-26
Filter expression as native string

rbs333:
Bug fix:

Solves problem if you pass a native string as a filter_expression not to set to blank string.

### DIFF
--- a/docs/overview/cli.ipynb
+++ b/docs/overview/cli.ipynb
@@ -24,7 +24,7 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "12:42:18 [RedisVL] INFO   RedisVL version 0.8.0\n"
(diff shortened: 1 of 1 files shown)


tests/unit/test_query_types.py
```diff
@@ -280,6 +280,74 @@ def test_text_query():
...
+    )
+
+    # Check that filter is stored correctly
+    assert text_query.filter == string_filter
+
+    # Check that the generated query string includes both text search and filter
+    query_string = str(text_query)
+    assert f"@{text_field_name}:(search | document | 12345)" in query_string
+    assert f"AND {string_filter}" in query_string
+    assert string_filter in query_string
```
justin-cechmanek, line 308: this second assertion seems redundant.

justin-cechmanek (commented):
👍🏼

PR #369, opened 2025-09-07
 feat(cache): add warnings when using sync methods with async-only Redis client

Shades-en:
### What is happening?
When below conditions are met -

1. I Initialise `EmbeddingCache` with `async_redis_client` only.
2. I then use `get` method of `EmbeddingCache` - `cache.get(text=text, model_name=model_name)`

It defaults to using local redis client instead of the supplied `async_redis_client` without any warnings. Which in turn results to empty cache always.

**It gets worse when we use `SemanticCache` and use `EmbeddingCache` with `async_redis_client` in its `vectorizer`**

Since `SemanticCache` only accepts synchronous `redis_client` we may tend to use `check` method or `store` method instead of `astore` or `acheck` method, which in turn again calls `cache.get(text=text, model_name=model_name)` and it return **EMPTY CACHE**

Hence warning logs become neccessary.

### What is done?
I have added some warning logs in case if developer supplies `async_redis_client` but uses `get`/ `mget`/`set`/`mset` methods.

### DIFF
--- a/redisvl/extensions/cache/embeddings/embeddings.py
+++ b/redisvl/extensions/cache/embeddings/embeddings.py
@@ -167,6 +167,13 @@ class EmbeddingsCache(BaseCache):
 
             embedding_data = cache.get_by_key("embedcache:1234567890abcdef")
         """
+        if self._owns_redis_client is False and self._redis_client is None:
+            logger.warning(
+                "You are using redis_client but you had initialised the EmbeddingCache with an async_redis_client only. "
+                "If you intended to use async_redis_client, please use async aget(text, model_name) method of EmbeddingCache instead "
+                "of get(text, model_name)."
+            )
+
         client = self._get_redis_client()
 
         # Get all fields
@@ -201,6 +208,13 @@ class EmbeddingsCache(BaseCache):
         """
         if not keys:
             return []
+
+        if self._owns_redis_client is False and self._redis_client is None:
+            logger.warning(
+                "You are using redis_client but you had initialised the EmbeddingCache with an async_redis_client only. "
+                "If you intended to use async_redis_client, please use async amget(texts, model_name) method of EmbeddingCache instead "
+                "of mget(texts, model_name)."
+            )
 
         client = self._get_redis_client()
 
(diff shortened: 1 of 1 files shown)


bsbodden:
## Suggested Improvements

I've reviewed this PR and have some suggestions to improve the warning messages and prevent log spam:

### 1. Add a class-level warning flag

Add this after line 14 (class definition):
```python
class EmbeddingsCache(BaseCache):
    """Embeddings Cache for storing embedding vectors with exact key matching."""

    _warning_shown: bool = False  # Class-level flag to prevent warning spam
```

### 2. Improve warning messages

Replace the warning blocks in all four methods with cleaner, spam-prevented versions:

**In `get_by_key` method (around line 170):**
```python
if self._owns_redis_client is False and self._redis_client is None:
    if not EmbeddingsCache._warning_shown:
        logger.warning(
            "EmbeddingsCache initialized with async_redis_client only. "
            "Use async methods (aget_by_key) instead of sync methods (get_by_key)."
        )
        EmbeddingsCache._warning_shown = True
```

**In `mget_by_keys` method (around line 205):**
```python
if self._owns_redis_client is False and self._redis_client is None:
    if not EmbeddingsCache._warning_shown:
        logger.warning(
            "EmbeddingsCache initialized with async_redis_client only. "
            "Use async methods (amget_by_keys) instead of sync methods (mget_by_keys)."
        )
        EmbeddingsCache._warning_shown = True
```

**In `set` method (around line 286):**
```python
if self._owns_redis_client is False and self._redis_client is None:
    if not EmbeddingsCache._warning_shown:
        logger.warning(
            "EmbeddingsCache initialized with async_redis_client only. "
            "Use async methods (aset) instead of sync methods (set)."
        )
        EmbeddingsCache._warning_shown = True
```

**In `mset` method (around line 336):**
```python
if self._owns_redis_client is False and self._redis_client is None:
    if not EmbeddingsCache._warning_shown:
        logger.warning(
            "EmbeddingsCache initialized with async_redis_client only. "
            "Use async methods (amset) instead of sync methods (mset)."
        )
        EmbeddingsCache._warning_shown = True
```

### Benefits of these changes:
1. **Clearer messages** - Directly tell users which async method to use
2. **No log spam** - Warning only shows once per session
3. **Better DX** - More actionable guidance for developers

### Test Coverage

Consider adding tests to verify:
- Warnings appear when using sync methods with async-only client  
- Warning only shows once (no spam)
- No warnings when using correct client types

I have a complete test file ready if you'd like to add test coverage for this feature.

bsbodden:
@Shades-en I'm closing this one and replacing with a rebased version and an extra test https://github.com/redis/redis-vl-python/pull/390 - you'll be acknowledge in the release notes, thank you for the contribution!

Shades-en:
> @Shades-en I'm closing this one and replacing with a rebased version and an extra test #390 - you'll be acknowledge in the release notes, thank you for the contribution!

Thank you @bsbodden sure that would be great! You can acknowledge this account itself!

PR #375, opened 2025-09-18
Fix filter expression being overridden with blank string

justin-cechmanek:
passing a string as filter expression no longer sets the filter to a blank string


justin-cechmanek:
This applies the same fix we needed for TextQuery to our HybridQuery class. For reference: https://github.com/redis/redis-vl-python/pull/367/files

PR #379, opened 2025-09-23
fix: handle ClusterPipeline missing nodes_manager attribute (#365)

### LINKED ISSUE #365: `AttributeError: 'ClusterPipeline' object has no attribute 'nodes_manager'` when using `index.load()` with Redis Cluster
**Description**

When using `redisvl` with a Redis Cluster client, calling the `index.load()` method results in an `AttributeError`. The error originates from within the `redis-py` library, which `redisvl` calls during its batch writing process.

The traceback indicates that `redisvl`'s storage layer uses a `ClusterPipeline` object for its `awrite` operation. An internal helper function in `redis-py`, `get_protocol_version`, is then called, which attempts to access `client.nodes_manager`. However, the `ClusterPipeline` object does not have this attribute, leading to the crash.

This issue occurs when `redisvl` is used as a dependency, for example, within `langgraph-checkpoint-redis` when it attempts to save a checkpoint to a Redis Cluster.

**Library Versions**

*   **redisvl**: `0.8.0`
*   **redis**: `6.4.0` (also observed with other `5.x` versions including 5.3.1)
*   **langgraph-checkpoint-redis**: `0.1.1`
*   **python**: `3.12`

**Traceback**

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/redisvl/index/index.py", line 1566, in load
    return await self._storage.awrite(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/redisvl/index/storage.py", line 477, in awrite
    await self._aset(pipe, key, obj)
  File "/usr/local/lib/python3.12/site-packages/redisvl/index/storage.py", line 679, in _aset
    client.json().set(key, "$", obj)  # type: ignore[return-value, misc]
    ^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/redis/commands/redismodules.py", line 24, in json
    jj = JSON(client=self, encoder=encoder, decoder=decoder)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/redis/commands/json/__init__.py", line 71, in __init__
    if get_protocol_version(self.client) in ["3", 3]:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/redis/commands/helpers.py", line 118, in get_protocol_version
    return client.nodes_manager.connection_kwargs.get("protocol")
           ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'ClusterPipeline' object has no attribute 'nodes_manager'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/langgraph/pregel/loop.py", line 1123, in _checkpointer_put_after_previous
    await prev
  File "/usr/local/lib/python3.12/site-packages/langgraph/pregel/loop.py", line 1125, in _checkpointer_put_after_previous
    await cast(BaseCheckpointSaver, self.checkpointer).aput(
  File "/usr/local/lib/python3.12/site-packages/langgraph/checkpoint/redis/aio.py", line 1079, in aput
    raise e
  File "/usr/local/lib/python3.12/site-packages/langgraph/checkpoint/redis/aio.py", line 998, in aput
    await self.checkpoints_index.load(
  File "/usr/local/lib/python3.12/site-packages/redisvl/index/index.py", line 1583, in load
    raise RedisVLError(f"Failed to load data: {str(e)}") from e
redisvl.exceptions.RedisVLError: Failed to load data: 'ClusterPipeline' object has no attribute 'nodes_manager'
```

bsbodden:
- Add get_protocol_version wrapper to safely handle ClusterPipeline objects
  - ClusterPipeline objects lack nodes_manager attribute, causing AttributeError
  - Fallback to protocol "3" when unable to detect version from cluster pipelines
  - Handle both sync and async ClusterPipeline cases

  Fixes #365

### DIFF
--- a/redisvl/utils/redis_protocol.py
+++ b/redisvl/utils/redis_protocol.py
@@ -1,0 +1,95 @@
+"""
+Utilities for handling Redis protocol version detection safely across different client types.
+
+This module provides safe wrappers around redis-py's get_protocol_version function
+to handle edge cases with Redis Cluster pipelines.
+"""
+
+from typing import Union
+
+from redis.asyncio.cluster import ClusterPipeline as AsyncClusterPipeline
+from redis.cluster import ClusterPipeline
+from redis.commands.helpers import get_protocol_version as redis_get_protocol_version
+
+from redisvl.utils.log import get_logger
+
+logger = get_logger(__name__)
+
+
+def get_protocol_version(client) -> str:
+    """
+    Wrapper for redis-py's get_protocol_version that handles ClusterPipeline.
+
+    ClusterPipeline doesn't have nodes_manager attribute, so we need to
+    handle this case specially to avoid AttributeError.
+
+    Args:
+        client: Redis client, pipeline, or cluster pipeline object
+
+    Returns:
+        str: Protocol version ("2" or "3")
+
+    Note:
+        This function addresses issue #365 where get_protocol_version() fails
+        with ClusterPipeline objects due to missing nodes_manager attribute.
+    """
+    # Handle sync ClusterPipeline
+    if isinstance(client, ClusterPipeline):
+        try:
+            # Try to get protocol from the underlying cluster client
+            if hasattr(client, "_redis_cluster") and client._redis_cluster:
+                try:
+                    result = redis_get_protocol_version(client._redis_cluster)
+                    if result is not None:
+                        return result
+                except (AttributeError, Exception):
+                    # If anything fails, fall back to default
+                    pass
+
+            logger.debug(
+                "ClusterPipeline without valid _redis_cluster, defaulting to protocol 3"
+            )
+            return "3"
+        except AttributeError as e:
+            logger.debug(
+                f"Failed to get protocol version from ClusterPipeline: {e}, defaulting to protocol 3"
+            )
+            return "3"
+
+    # Handle async ClusterPipeline
+    if isinstance(client, AsyncClusterPipeline):
+        try:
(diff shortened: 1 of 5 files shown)


redisvl/utils/redis_protocol.py
```diff
@@ -0,0 +1,95 @@
...
+                    if result is not None:
+                        return result
+                except (AttributeError, Exception):
+                    # If anything fails, fall back to default
+                    pass
+
+            logger.debug(
+                "AsyncClusterPipeline without valid _redis_cluster, defaulting to protocol 3"
+            )
+            return "3"
```
rbs333, line 75: @bsbodden could you help me unpack this fix a bit? 

So we were failing before because the objects was trying to use an attribute `nodes_manager` that didn't exist. The fix implemented here checks the protocol version and defaults to 3 in the case of not knowing. 

Was the fail happening because it was previously defaulting to "2" when in cluster mode vs with other connections types? 

I haven't done a ton with protocols in redis so trying to understand some of those key differences.

bsbodden: Well shit, it should be 2, not 3 - and I was just able to get a real integration test working with my cluster setup and this won't cut it. Good catch.

Issue #361, opened 2025-08-11
IndexSchema constructor and validation have conflicting type expectations


drcrallen:
https://github.com/redis/redis-vl-python/blob/04736f138a8326b27c73d6f0acad51ca1a025ba8/redisvl/schema/schema.py#L146

expects `fields: Dict[str, BaseField] = {}`

but https://github.com/redis/redis-vl-python/blob/04736f138a8326b27c73d6f0acad51ca1a025ba8/redisvl/schema/schema.py#L182-L186 will throw an error if `fields` is a dict.

PR #381, opened 2025-09-23
fix: resolve IndexSchema fields type annotation inconsistency (#361)

bsbodden:
Fix conflicting type expectations in IndexSchema constructor and validation.
  The type annotation declared `fields: Dict[str, BaseField]` but the validator
  raised an error when fields was actually provided as a dict.

  Changes:
  - Update model validator to accept both list and dict formats for fields
  - Support dict with field definitions: {"field_name": {"name": "field_name", "type": "text"}}
  - Support dict with BaseField instances: {"field_name": BaseFieldInstance}
  - Maintain backward compatibility with existing list format

### DIFF
--- a/redisvl/schema/schema.py
+++ b/redisvl/schema/schema.py
@@ -1,10 +1,11 @@
 import re
+from collections.abc import Mapping, Sequence
 from enum import Enum
 from pathlib import Path
-from typing import Any, Dict, List, Literal
+from typing import Any, Dict, List, Literal, Union
 
 import yaml
-from pydantic import BaseModel, model_validator
+from pydantic import BaseModel, Field, model_validator
 from redis.commands.search.field import Field as RedisField
 
 from redisvl.schema.fields import BaseField, FieldFactory
@@ -143,8 +144,11 @@ class IndexSchema(BaseModel):
 
     index: IndexInfo
     """Details of the basic index configurations."""
-    fields: Dict[str, BaseField] = {}
-    """Fields associated with the search index and their properties"""
+    fields: Dict[str, BaseField] = Field(default_factory=dict)
+    """Fields associated with the search index and their properties.
+
+    Note: When creating from dict/YAML, provide fields as a list of field definitions.
+    The validator will convert them to a Dict[str, BaseField] internally."""
     version: Literal["0.1.0"] = "0.1.0"
     """Version of the underlying index schema."""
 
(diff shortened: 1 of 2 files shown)


redisvl/schema/schema.py
```diff
@@ -144,7 +144,10 @@ class IndexSchema(BaseModel):
     index: IndexInfo
     """Details of the basic index configurations."""
     fields: Dict[str, BaseField] = {}
```
abrookins: You didn't add this code, but I believe this is the same problem as a default list or dictionary in a function signature, which is everything shares the same global default object. So instead, we would want:
```python
fields: Dict[str, BaseField] = Field(default_factory=dict)
```

abrookins: Also, I would use Mapping instead of Dict.

redisvl/schema/schema.py
```diff
@@ -144,7 +144,10 @@ class IndexSchema(BaseModel):
...
-    """Fields associated with the search index and their properties"""
+    """Fields associated with the search index and their properties.
+
+    Note: When creating from dict/YAML, provide fields as a list of field definitions.
```
abrookins, line 150: Does it support both input types now even if it normalizes to dict?

redisvl/schema/schema.py
```diff
@@ -181,17 +184,36 @@ def validate_and_create_fields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
-            field = cls._make_field(index.storage_type, **field_input)
-            if field.name in prepared_fields:
-                raise ValueError(
-                    f"Duplicate field name: {field.name}. Field names must be unique across all fields."
-                )
-            prepared_fields[field.name] = field
+            # If fields is already a dict of BaseField instances, use it directly
+            for name, field in input_fields.items():
+                if isinstance(field, BaseField):
+                    prepared_fields[name] = field
```
abrookins, line 198: Would it be a problem here if field.name and name did not match?

redisvl/schema/schema.py
```diff
@@ -181,17 +184,36 @@ def validate_and_create_fields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                elif isinstance(field, dict):
```
abrookins, line 199: I would do collections.abc.Mapping for this isinstance check (just to be clear, not the Mapping from typing)

redisvl/schema/schema.py
```diff
@@ -181,17 +184,36 @@ def validate_and_create_fields(cls, values: Dict[str, Any]) -> Dict[str, Any]:
...
+                    if field_obj.name != name:
+                        raise ValueError(
+                            f"Field name mismatch: key '{name}' vs field name '{field_obj.name}'"
+                        )
+                    prepared_fields[name] = field_obj
+                else:
+                    raise ValueError(
+                        f"Invalid field type for '{name}': expected BaseField or dict"
+                    )
+        elif isinstance(input_fields, list):
```
abrookins: If you want to be less strict about the types, you could do `elif isinstance(input_fields, Sequence) and not isinstance(input_fields, (str, bytes))` (from collections.abc).

redisvl/schema/schema.py
bsbodden: Done

redisvl/schema/schema.py
bsbodden: Done

redisvl/schema/schema.py
bsbodden: Done

redisvl/schema/schema.py
bsbodden: Done

Issue #368, opened 2025-08-27
Ideally we handle key_prefix separators better


tylerhutcherson:
https://github.com/langchain-ai/langchain-redis/issues/78

This can be addressed within the core of RedisVL buy checking keys for additional `:` separators between segments of a full key.

So for example, this means we would use the index schema property `key_separator` to check if it's also been applied in the index schema property `key_prefix` if so, handle it properly.

PR #382, opened 2025-09-24
fix: improve key separator handling to avoid double separators (#368)

bsbodden:
Normalize key prefixes by removing trailing separators when constructing
  Redis keys to prevent double separator issues (e.g., "user::123" becomes
  "user:123").

  Changes:
  - Fix BaseStorage._key() to strip trailing separators from prefix
  - Update SemanticRouter to use index's key_separator instead of hardcoded ':'
  - Extract route pattern generation into reusable _route_pattern() method
  - Fix scan pattern generation to respect custom separators

### DIFF
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -221,8 +221,25 @@ class SemanticRouter(BaseModel):
 
     @staticmethod
     def _route_ref_key(index: SearchIndex, route_name: str, reference_hash: str) -> str:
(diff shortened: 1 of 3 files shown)


rbs333 (approved):
Looks good! For backwards compatibility, would we have an issue if someone updates and index that had a weird prefix before and then it wouldn't reindex the data? I don't think this is a major concern but we might want to include a callout in the release notes and/or commit message to document.

Issue #349, opened 2025-06-16
how to get system prompt with get_recent


linyubupa:
how to get system prompt with get_recent

tylerhutcherson:
@justin-cechmanek can you look into this and offer a workaround for the time being? We might want to scope a small effort to improve the UX of the message history class.

PR #387, opened 2025-09-26
feat: add role filtering to message history get_recent method (#349)

bsbodden:
Add role parameter to get_recent() and get_relevant() methods in both MessageHistory and SemanticMessageHistory classes to enable filtering messages by role type.

Features:
- Support single role filtering: role="system"
- Support multiple role filtering: role=["system", "user"]
- Valid roles: "system", "user", "llm", "tool"
- Backward compatible: role=None returns all messages
- Works with existing parameters (top_k, session_tag, raw, etc.)
- Comprehensive validation with clear error messages

The implementation maintains full backward compatibility while enabling users to retrieve only specific message types like system prompts.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -48,6 +48,12 @@ index = SearchIndex(schema, redis_url="redis://localhost:6379")
   token.strip().strip(",").replace(""", "").replace(""", "").lower()
   ```
 
+### Git Operations
+**CRITICAL**: NEVER use `git push` or attempt to push to remote repositories. The user will handle all git push operations.
+
+### Code Quality
+**IMPORTANT**: Always run `make format` before committing code to ensure proper formatting and linting compliance.
+
 ### README.md Maintenance
 **IMPORTANT**: DO NOT modify README.md unless explicitly requested.
 
--- a/redisvl/extensions/message_history/message_history.py
+++ b/redisvl/extensions/message_history/message_history.py
@@ -119,6 +119,7 @@ class MessageHistory(BaseMessageHistory):
         as_text: bool = False,
         raw: bool = False,
         session_tag: Optional[str] = None,
+        role: Optional[Union[str, List[str]]] = None,
     ) -> Union[List[str], List[Dict[str, str]]]:
         """Retrieve the recent message history in sequential order.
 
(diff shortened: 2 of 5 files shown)


tests/unit/test_role_filter_get_recent.py
```diff
@@ -0,0 +1,373 @@
...
+            },
+            {
+                "role": "user",
+                "content": "User message 2",
+                "session_tag": "test_session",
+                "entry_id": "entry2",
+                "timestamp": 123456790,
+            },
+        ]
+        history._index.query = Mock(return_value=mock_messages)
```
justin-cechmanek: Can we do these queries for real instead of using Mock?

redisvl/extensions/message_history/message_history.py
```diff
@@ -130,17 +131,45 @@ def get_recent(
...
+                    raise ValueError("roles cannot be empty")
+                for r in role:
+                    if r not in valid_roles:
+                        raise ValueError(
+                            f"Invalid role '{r}'. Valid roles are: {valid_roles}"
+                        )
+                roles_to_filter = role
+            else:
+                raise ValueError("role must be a string or list of strings")
+
```
justin-cechmanek: This role validation is repeated in get_recent and get_relevant  for both standard and semantic classes. Should be move it to a base class internal method?

justin-cechmanek (changes requested):
Looks good! Just the two questions/suggestions

CLAUDE.md
```diff
@@ -48,6 +48,12 @@ index = SearchIndex(schema, redis_url="redis://localhost:6379")
   token.strip().strip(",").replace(""", "").replace(""", "").lower()
   ```
 
+### Git Operations
+**CRITICAL**: NEVER use `git push` or attempt to push to remote repositories. The user will handle all git push operations.
```
justin-cechmanek, line 52: lol, best to get this in early

justin-cechmanek (approved):
👍🏼

CLAUDE.md
bsbodden: Claude is a rogue operator!

Issue #346, opened 2025-06-03
AsyncRedisCluster.from_url() raises TypeError when cluster=true is passed via URL or kwargs due to unsupported argument in redis.asyncio.RedisClient


mprostko:
### Description

When using `redis-vl` with `redis-py`, the AsyncRedisCluster.from_url() method raises a TypeError if the `cluster=true` setting is passed either via the URL query string or as a keyword argument.

```**kwargs: Any``` argument was removed in `redis.asyncio.RedisCluster.__init__()` function in the following PR: https://github.com/redis/redis-py/pull/2217

### Example:

```python
from redis import Redis, RedisCluster
from redis.asyncio import Redis as AsyncRedis
from redis.asyncio.cluster import RedisCluster as AsyncRedisCluster
from redisvl.redis.utils import is_cluster_url

# --- Passing cluster setting via url query
url = 'redis://<user>:<password>@<cluster_host>:6379/?cluster=true'
print(f"Is cluster url? {str(is_cluster_url(url))}") # True

r = Redis.from_url(url) 		# OK
rc = RedisCluster.from_url(url) 	# OK
ar = AsyncRedis.from_url(url) 		# OK
arc = AsyncRedisCluster.from_url(url) 	# TypeError: RedisCluster.__init__() got an unexpected keyword argument 'cluster'

# --- Passing cluster setting via kwargs
url = 'redis://<user>:<password>@<cluster_host>:6379'
kwargs = { "cluster": True }
print(f"Is cluster url? {str(is_cluster_url(url, **kwargs))}") # True

r = Redis.from_url(url, **kwargs) 		# OK
rc = RedisCluster.from_url(url, **kwargs) 	# OK
ar = AsyncRedis.from_url(url, **kwargs) 	# OK
arc = AsyncRedisCluster.from_url(url, **kwargs) # TypeError: RedisCluster.__init__() got an unexpected keyword argument 'cluster'
```

### Suggested Fix:

Before calling `AsyncRedisCluster.from_url()`, the cluster parameter should be stripped from both the URL and kwargs, as those are currently not supported (https://github.com/redis/redis-py/blob/master/redis/asyncio/cluster.py#L263-L310).

This change should be implemented here: https://github.com/redis/redis-vl-python/blob/cd74f16886516d42563a8f936902debba6a88b7a/redisvl/redis/connection.py#L295-L296
[...]

PR #388, opened 2025-09-27
fix: Handle cluster parameter in AsyncRedisCluster connections (#346)

bsbodden:
AsyncRedisCluster.from_url() doesn't accept 'cluster' parameter but it might be present in URLs or kwargs for compatibility. This fix strips the cluster parameter before creating async Redis connections.

Changes:
- Add _strip_cluster_from_url_and_kwargs helper function
- Update _get_aredis_connection to strip cluster parameter
- Update get_async_redis_cluster_connection to strip cluster parameter
- Update deprecated get_async_redis_connection to handle both cluster types
- Strip cluster parameter for both AsyncRedis and AsyncRedisCluster

### DIFF
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -1,5 +1,6 @@
 import os
-from typing import Any, Dict, List, Optional, Type
+from typing import Any, Dict, List, Optional, Tuple, Type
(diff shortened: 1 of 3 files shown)


abrookins (approved):
Looks good! 👍

PR #402, opened 2025-10-07
Adds multi vector query class

justin-cechmanek:


tylerhutcherson:
Very nice. I do think it could be interesting if this vector class also handled the conversion to bytes (and thus any compression based on choice of dtype) 

I think we handle this within the query class today but some of that logic could be pulled into here as a true data container for vectors -- might be a cleaner and more testable pattern? @bsbodden @abrookins

redisvl/query/aggregate.py
```diff
@@ -227,3 +252,152 @@ def _build_query_string(self) -> str:
...
+        self._num_results = num_results
+
+        if isinstance(vectors, Vector):
+            self._vectors = [vectors]
+        else:
+            self._vectors = vectors  # type: ignore
+
+        if not all([isinstance(v, Vector) for v in self._vectors]):
+            raise TypeError(
+                "vector arugment must be a Vector object or list of Vector objects."
```
nkanu17: nit: typo -> `arugment`

redisvl/query/aggregate.py
```diff
@@ -227,3 +252,152 @@ def _build_query_string(self) -> str:
...
+        Args:
+            vectors (Union[Vector, List[Vector]]): The Vectors to perform vector similarity search.
+            return_fields (Optional[List[str]], optional): The fields to return. Defaults to None.
+            filter_expression (Optional[Union[str, FilterExpression]]): The filter expression to use.
+                Defaults to None.
+            num_results (int, optional): The number of results to return. Defaults to 10.
+            return_score (bool): Whether to return the combined vector similarity score.
+                Defaults to False.
+            dialect (int, optional): The Redis dialect version. Defaults to 2.
+        """
```
nkanu17, line 324: @justin-cechmanek I noticed the `return_score` parameter isn't actually used, scores are always calculated regardless of its value. Since it doesn't exist in the parent class or HybridQuery either, I'd suggest removing it unless you had a specific plan for it?

justin-cechmanek: removed.

abrookins:
@tylerhutcherson Yes, I think it makes sense to have fields own converting themselves to `bytes`:
```python
from redisvl.redis.utils import array_to_buffer


class Vector(BaseModel):
    # ...
    def as_buffer(self) -> bytes:
        if isinstance(self.vector, (bytes, bytearray)):
            return bytes(self.vector)
        return array_to_buffer(self.vector, dtype=self.dtype)
```

We should be able to touch that up later though, if we want to get in a first pass.

Another thing I was wondering. This PR hard-codes the distance at 2.0 (all items?), but wouldn't it also work to expose that as a parameter on `Vector`? Similarly, we have room to change this later, but I wondered if that was intentional.

justin-cechmanek:
> @tylerhutcherson Yes, I think it makes sense to have fields own converting themselves to `bytes`:
> 
> ```python
> from redisvl.redis.utils import array_to_buffer
> 
> 
> class Vector(BaseModel):
>     # ...
>     def as_buffer(self) -> bytes:
>         if isinstance(self.vector, (bytes, bytearray)):
>             return bytes(self.vector)
>         return array_to_buffer(self.vector, dtype=self.dtype)
> ```
> 
> We should be able to touch that up later though, if we want to get in a first pass.
> 
> Another thing I was wondering. This PR hard-codes the distance at 2.0 (all items?), but wouldn't it also work to expose that as a parameter on `Vector`? Similarly, we have room to change this later, but I wondered if that was intentional.

We can expose the distance and tie it back to the Vector class. It's set to 2.0 now since we have to use cosine for multivector to run and this ensures all docs are eligible for selection. Exposing the distance would change this from a multi-vector query to a mulit-vector-range query, which I'll add as a follow up ticket.

PR #404, opened 2025-10-13
Feat/raae 1094 add svs vamana implementation

nkanu17:
Adds first-class SVS-VAMANA support to redis-vl-python, including compression options, schema validation, Redis capability checks, docs, and tests.

**Highlights**
- New SVSVectorField + VectorIndexAlgorithm.SVS_VAMANA.
- Compression enums: LVQ4, LVQ4x4, LVQ4x8, LVQ8, LeanVec4x8, LeanVec8x8.
- Validation: only float16/float32; reduce < dims; reduce allowed only with LeanVec.
- Capability checks (sync/async): block create if requirements not met; clear RedisModuleVersionError.
- CompressionAdvisor: recommend configs + estimate memory savings.
- Parsing: normalize SVS attrs (incl. basic compression name mapping).

**Requirements**
- Redis ≥ 8.2.0 and RediSearch ≥ 2.8.10 (or SearchLight ≥ 2.8.10) for SVS-Vamana
- Backwards compatible, just can't use SVS-Vamana.

**Compatibility**
- No breaking changes for flat/hnsw.
- Backwards compatible, just can't use SVS-Vamana.

Docs & Tests
- Docs: schema.rst adds SVS section + example.
- Tests: validation, capability detection, CompressionAdvisor, and basic integration.
- Notebook or python guide on usage

### DIFF
--- a/docs/api/schema.rst
+++ b/docs/api/schema.rst
@@ -60,36 +60,80 @@ Fields in the schema can be defined in YAML format or as a Python dictionary, sp
         }
     }
 
-Supported Field Types and Attributes
-====================================
+Basic Field Types
+=================
 
-Each field type supports specific attributes that customize its behavior. Below are the field types and their available attributes:
+RedisVL supports several basic field types for indexing different kinds of data. Each field type has specific attributes that customize its indexing and search behavior.
 
-**Text Field Attributes**:
+Text Fields
+-----------
+
+Text fields support full-text search with stemming, phonetic matching, and other text analysis features.
+
+.. currentmodule:: redisvl.schema.fields
+
+.. autoclass:: TextField
+   :members:
+   :show-inheritance:
 
 .. autoclass:: TextFieldAttributes
    :members:
    :undoc-members:
 
-**Tag Field Attributes**:
+Tag Fields
+----------
+
+Tag fields are optimized for exact-match filtering and faceted search on categorical data.
+
+.. autoclass:: TagField
+   :members:
+   :show-inheritance:
 
 .. autoclass:: TagFieldAttributes
    :members:
    :undoc-members:
 
-**Numeric Field Attributes**:
+Numeric Fields
+--------------
+
+Numeric fields support range queries and sorting on numeric data.
+
+.. autoclass:: NumericField
+   :members:
+   :show-inheritance:
 
 .. autoclass:: NumericFieldAttributes
    :members:
    :undoc-members:
 
-**Geo Field Attributes**:
+Geo Fields
+----------
+
+Geo fields enable location-based search with geographic coordinates.
+
+.. autoclass:: GeoField
+   :members:
+   :show-inheritance:
 
 .. autoclass:: GeoFieldAttributes
    :members:
    :undoc-members:
 
-**Common Vector Field Attributes**:
+Vector Field Types
+==================
+
+Vector fields enable semantic similarity search using various algorithms. All vector fields share common attributes but have algorithm-specific configurations.
+
+Common Vector Attributes
+------------------------
+
+All vector field types share these base attributes:
+
+.. autoclass:: BaseVectorFieldAttributes
+   :members:
+   :undoc-members:
+
+**Key Attributes:**
 
 - `dims`: Dimensionality of the vector (e.g., 768, 1536).
 - `algorithm`: Indexing algorithm for vector search:
@@ -106,7 +150,8 @@ Each field type supports specific attributes that customize its behavior. Below 
 - `initial_cap`: Initial capacity hint for memory allocation (optional).
 - `index_missing`: When True, allows searching for documents missing this field (optional).
 
-**HNSW Vector Field Specific Attributes**:
+HNSW Vector Fields
+------------------
 
 HNSW (Hierarchical Navigable Small World) - Graph-based approximate search with excellent recall. **Best for general-purpose vector search (10K-1M+ vectors).**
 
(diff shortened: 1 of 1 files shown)


redisvl/utils/migration.py
```diff
@@ -26,26 +26,26 @@
 
 class IndexMigrator:
```
rbs333: I really like this as a way of creating backwards compatibility but might recommend we turn this into a recipe vs a class. Since it would be more of a one time thing people would do I'd rather give an easy guide vs supporting a class in the library moving forward.

docs/api/schema.rst
```diff
@@ -97,16 +97,64 @@ Each field type supports specific attributes that customize its behavior. Below
 **Common Vector Field Attributes**:
 
 - `dims`: Dimensionality of the vector.
-- `algorithm`: Indexing algorithm (`flat` or `hnsw`).
-- `datatype`: Float datatype of the vector (`bfloat16`, `float16`, `float32`, `float64`).
+- `algorithm`: Indexing algorithm (`flat`, `hnsw`, or `svs-vamana`).
+- `datatype`: Float datatype of the vector (`bfloat16`, `float16`, `float32`, `float64`). Note: SVS-VAMANA only supports `float16` and `float32`.
 - `distance_metric`: Metric for measuring query relevance (`COSINE`, `L2`, `IP`).
+- `initial_cap`: Initial capacity for the index (optional).
+- `index_missing`: When True, allows searching for documents missing this field (optional).
```
rbs333, line 151: This docstring doesn't give me the full picture. As a dev I'd still be confused about what it does.

docs/api/schema.rst
```diff
@@ -97,16 +97,64 @@ Each field type supports specific attributes that customize its behavior. Below
...
-- `ef_runtime`: Max top candidates during search.
-- `epsilon`: Range search boundary factor.
+- `m`: Max outgoing edges per node in each layer (default: 16).
+- `ef_construction`: Max edge candidates during build time (default: 200).
+- `ef_runtime`: Max top candidates during search (default: 10).
+- `epsilon`: Range search boundary factor (default: 0.01).
+
+**SVS-VAMANA Vector Field Specific Attributes**:
+
+SVS-VAMANA (Scalable Vector Search with VAMANA graph algorithm) provides fast approximate nearest neighbor search with optional compression support. This algorithm is optimized for Intel hardware and offers reduced memory usage through vector compression.
```
rbs333: like this explanation

redisvl/redis/connection.py
```diff
@@ -101,6 +106,92 @@ def unpack_redis_modules(module_list: List[Dict[str, Any]]) -> Dict[str, Any]:
     return {module["name"]: module["ver"] for module in module_list}
 
 
+@dataclass
+class VectorSupport:
```
rbs333: Is this class only leveraged for SVS support?

abrookins:
Nice! I think we should talk about migrations, but agree that in this case, for migrating a specific aspect of an index rather than more general schema migrations support, or data migrations, a recipe may be best.

docs/api/schema.rst
```diff
@@ -67,46 +67,325 @@ Each field type supports specific attributes that customize its behavior. Below
...
 **Common Vector Field Attributes**:
 
-- `dims`: Dimensionality of the vector.
-- `algorithm`: Indexing algorithm (`flat` or `hnsw`).
-- `datatype`: Float datatype of the vector (`bfloat16`, `float16`, `float32`, `float64`).
-- `distance_metric`: Metric for measuring query relevance (`COSINE`, `L2`, `IP`).
+- `dims`: Dimensionality of the vector (e.g., 768, 1536).
+- `algorithm`: Indexing algorithm for vector search:
+
+  - `flat`: Brute-force exact search. 100% recall, slower for large datasets. Best for <10K vectors.
```
rbs333, line 141: I would drop "Brute-force" here and just say exact search. Brute force kind of implies it's a bad option when often it's the best option.

docs/api/schema.rst
```diff
@@ -67,46 +67,325 @@ Each field type supports specific attributes that customize its behavior. Below
...
-- `ef_runtime`: Max top candidates during search.
-- `epsilon`: Range search boundary factor.
+HNSW (Hierarchical Navigable Small World) - Graph-based approximate search with excellent recall. **Best for general-purpose vector search (10K-1M+ vectors).**
+
+.. dropdown:: When to use HNSW & Performance Details
+   :color: info
+
+   **Use HNSW when:**
+
+    - Medium to large datasets (10K-1M+ vectors) requiring high recall rates
```
rbs333, line 163: Typically we recommend this for more in the 100k-1M range

docs/api/schema.rst
```diff
@@ -67,46 +67,325 @@ Each field type supports specific attributes that customize its behavior. Below
...
\ No newline at end of file
+   **Compression selection guide:**
+
+    - **No compression**: Best performance, standard memory usage
+
+    - **LVQ4/LVQ8**: Good balance of compression (2x-4x) and performance
+
+    - **LeanVec4x8/LeanVec8x8**: Maximum compression (up to 8x) with dimensionality reduction
+
+   **Memory Savings Examples (1M vectors, 768 dims):**
```
rbs333, line 248: love this

docs/api/schema.rst
```diff
@@ -67,46 +67,325 @@ Each field type supports specific attributes that customize its behavior. Below
...
+   :header-rows: 1
+   :widths: 15 20 20 20 25
+
+   * - Algorithm
+     - Best For
+     - Performance
+     - Memory Usage
+     - Trade-offs
+   * - **FLAT**
+     - Small datasets (<10K vectors)
```
rbs333: again I think we 10x that number

redisvl/utils/compression.py
```diff
@@ -0,0 +1,216 @@
+"""SVS-VAMANA compression configuration utilities."""
+
+from typing import Literal, Optional, TypedDict, cast
+
+
+class SVSConfig(TypedDict, total=False):
```
rbs333: would just use pydantic here instead since that's our go to

Issue #407, opened 2025-10-17
SemanticCache threshold scale inconsistencies (async normalize_vector_distance and docstring)


abrookins:
Summary
There are inconsistencies around the distance threshold scale in SemanticCache between sync and async code paths, and the docstring, which can lead to surprising behavior.

Details
- Docstring in set_threshold says: "ValueError: If the threshold is not between 0 and 1.", but the validation actually enforces 0 <= distance_threshold <= 2 (Redis COSINE native units). This is confusing to users reading the docstring.
- Async check path constructs VectorRangeQuery with normalize_vector_distance=true, which treats the provided distance_threshold as normalized [0,1] and denormalizes it internally. The sync path (and the rest of SemanticCache) appears to treat thresholds as native Redis COSINE distances [0,2]. This causes the same numeric threshold to be interpreted differently between sync and async.

Why it matters
Users may set a threshold like 0.15 expecting a tight native COSINE distance, but the async path will interpret it as normalized (and denormalize to 1.85), returning far more results than expected.

Proposed options
1) Keep thresholds in native units (0-2) everywhere
   - Set normalize_vector_distance=false in the async VectorRangeQuery
   - Update docstrings to consistently say 0-2

2) Adopt normalized thresholds (0-1) everywhere
   - Keep normalize_vector_distance=true in async and set the same for sync
   - Normalize on input before constructing queries in all paths
   - Update docs accordingly

3) Make it configurable and explicit
   - Add a distance_scale flag ("redis" vs "normalized") on SemanticCache, default to current behavior, and convert as needed.

Recommendation
Option 1 for least surprise and parity with Redis search defaults: use native COSINE [0-2] everywhere and normalize only when explicitly requested. At minimum, align sync/async and fix the docstring.

Action items
- Decide desired global convention
- Align sync and async VectorRangeQuery usage accordingly
- Update docstrings and README/examples to clarify the scale

Context
Related changes were explored in LangCacheWrapper to make threshold scale explicit. Tracking this issue separately for SemanticCache to avoid unintentional behavior changes.

nkanu17:
The core sync/async inconsistency was fixed in this [commit](https://github.com/redis/redis-vl-python/commit/595c4509c78ed1e239c062ab5b93c459f930786a). The documentation needs to be [updated](https://github.com/redis/redis-vl-python/commit/8747a3ab174ee34ed9446855da66c720f18d2e5d) as well.

PR #408, opened 2025-10-17
Add LangCache wrapper (LLM cache extension) with configurable distance scale

abrookins:
## Summary
Add a LangCache wrapper to RedisVL LLM cache extensions so users can leverage LangCache for semantic caching via the familiar `BaseLLMCache` interface. Also make the wrapper’s distance threshold semantics explicit and configurable so users can swap it out for an existing `SemanticCache` instance easily.

## Motivation
- Use LangCache while keeping RedisVL’s developer experience
- Avoid surprises by making threshold scale explicit

## What’s included
- New `LangCacheWrapper` implementing `BaseLLMCache` (sync + async) that talks to the LangCache service:
  - Exact + semantic search toggles, TTL, and standard `store`/`check` APIs
  - Consistent result shape (including `vector_distance`) compatible with existing callers
  - Lazily imports LangCache and gives a friendly error if it's not installed
- Configurable distance threshold scale via `distance_scale`:
  - `"normalized"`: interpret `distance_threshold` as 0–1 semantic distance; converts to/from similarity (`1 - d`)
  - `"redis"`: interpret `distance_threshold` as native COSINE [0–2]; converts using `norm_cosine_distance` / `denorm_cosine_distance`

## Usage
```python
from redisvl.extensions.cache.llm.langcache import LangCacheWrapper

cache = LangCacheWrapper(
    name="langcache",
    cache_id="<id>",
    api_key="<key>",
    distance_scale="normalized"  # or "redis" for COSINE [0–2]
)

cache.store(prompt="Hello?", response="Hi\!")

hits = cache.check(
    prompt="Hello\!",
    distance_threshold=0.15,  # interpreted per distance_scale
)
```

## Docs
- Follow-up docs PR will add LangCache wrapper docs and clarify threshold scales

## Related
- Opened #407 to track threshold-scale inconsistencies (docstring vs sync/async) in `SemanticCache`

### DIFF
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -23,7 +23,7 @@ from redisvl.redis.constants import (
 from redisvl.redis.utils import convert_bytes, is_cluster_url
 from redisvl.types import AsyncRedisClient, RedisClient, SyncRedisClient
 from redisvl.utils.log import get_logger
-from redisvl.utils.utils import deprecated_function
+from redisvl.utils.utils import deprecated_argument, deprecated_function
 
 logger = get_logger(__name__)
 
@@ -500,6 +500,7 @@ class RedisConnectionFactory:
         return client
 
     @staticmethod
+    @deprecated_argument("url", "redis_url")
     async def _get_aredis_connection(
         redis_url: Optional[str] = None,
         **kwargs,
@@ -527,11 +528,6 @@ class RedisConnectionFactory:
         """
         _deprecated_url = kwargs.pop("url", None)
         url = _deprecated_url or redis_url or get_address_from_env()
-
-        if _deprecated_url is not None:
-            logger.warning(
-                "The `url` parameter is deprecated. Please use `redis_url` instead."
-            )
 
         client: AsyncRedisClient
         if url.startswith("redis+sentinel"):
@@ -563,6 +559,7 @@ class RedisConnectionFactory:
         return client
 
     @staticmethod
+    @deprecated_argument("url", "redis_url")
     def get_async_redis_connection(
         redis_url: Optional[str] = None,
         **kwargs,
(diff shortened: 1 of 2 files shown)


abrookins:
Addressed review feedback:
- Broadened attributes type in check/acheck to Optional[Dict[str, Any]]
- Clarified docstrings to document both conversion paths (normalized vs redis)
- Reduced duplication by factoring helpers: _similarity_threshold, _build_search_kwargs, _hits_from_response

redisvl/extensions/cache/llm/langcache.py
```diff
@@ -0,0 +1,533 @@
...
+
+from redisvl.extensions.cache.llm.base import BaseLLMCache
+from redisvl.extensions.cache.llm.schema import CacheHit
+from redisvl.query.filter import FilterExpression
+from redisvl.utils.log import get_logger
+from redisvl.utils.utils import denorm_cosine_distance, norm_cosine_distance
+
+logger = get_logger(__name__)
+
+try:
```
rbs333: This seem potentially odd to me that we catch an import error here. I would have imagined that if they want langcache they would `pip install redisvl[langcache]` and we'd maybe throw a warning here?

redisvl/index/index.py
```diff
@@ -1229,7 +1229,7 @@ async def connect(self, redis_url: Optional[str] = None, **kwargs):
             DeprecationWarning,
         )
         client = await RedisConnectionFactory._get_aredis_connection(
-            redis_url=redis_url, **kwargs
+            url=redis_url, **kwargs
```
rbs333: At some point last year we had an issue with a conflict between url and redis_url. My preference is for redis_url when that's what we mean

rbs333 (commented):
Overall looks good to me! Left a couple comments that I think we should sync on

abrookins:
Fully deprecated `url` as a parameter to the async connection functions per review.

PR #409, opened 2025-10-22
fix: inconsistent FLAT/HNSW algorithm recommendations

nkanu17:
Updated algorithm recommendations in docs/api/schema.rst to provide consistent, performance-based guidance:

   FLAT Algorithm:
     • Before: Mixed recommendations (<10K in some places, <100K in others)
     • After: Consistent <100K vectors recommendation across all sections

   HNSW Algorithm:
     • Before: 10K-1M+ vectors (creating a gap between FLAT and HNSW)
     • After: 100K-1M+ vectors (providing clear transition point)

### DIFF
--- a/docs/api/schema.rst
+++ b/docs/api/schema.rst
@@ -160,7 +160,7 @@ HNSW (Hierarchical Navigable Small World) - Graph-based approximate search with 
 
    **Use HNSW when:**
 
-    - Medium to large datasets (10K-1M+ vectors) requiring high recall rates
+    - Medium to large datasets (100K-1M+ vectors) requiring high recall rates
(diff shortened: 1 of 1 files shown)


docs/api/schema.rst
```diff
@@ -160,7 +160,7 @@ HNSW (Hierarchical Navigable Small World) - Graph-based approximate search with
 
    **Use HNSW when:**
 
-    - Medium to large datasets (10K-1M+ vectors) requiring high recall rates
+    - Medium to large datasets (100K-1M+ vectors) requiring high recall rates
```
nkanu17: It was updated every where else except the schema I believe

PR #411, opened 2025-10-23
Vector object automatically converts vectors to byte strings

justin-cechmanek:


redisvl/query/aggregate.py
```diff
@@ -1,10 +1,11 @@
 from typing import Any, Dict, List, Optional, Set, Tuple, Union
 
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, field_validator, model_validator
 from redis.commands.search.aggregation import AggregateRequest, Desc
+from typing_extensions import Self
 
 from redisvl.query.filter import FilterExpression
-from redisvl.redis.utils import array_to_buffer
+from redisvl.redis.utils import array_to_buffer, buffer_to_array
```
nkanu17: `buffer_to_array` is unused

nkanu17 (changes requested):
Looks good to me except for the unused import!

I am surprised that the linter tests are not catching this

PR #418, opened 2025-10-31
LangCache integration cleanup

abrookins:
Made a few changes/fixes to the LangCache integration:
- The name "LangCacheWrapper" just doesn't hit as deep. Also, it doesn't follow the same pattern as "SemanticCache," the Redis-based cache in this library. Seeing "Cache" twice feels better than "Wrapper."
- Adjust how we send attributes
- Use the correct import for the LangCache client


hanchoong-redis (approved):

bsbodden (approved):
LGTM! 🚀

PR #420, opened 2025-11-04
fix #417: fix setting ttl on async pipeline

### LINKED ISSUE #417: setting ttl delete the pipeline commands (cluster+async)
Here is my example to reproduce, using `redis==6.4.0` and `redisvl==0.10.0` (docker-compose.yml to reproduce [here](https://github.com/dfroger/issue/tree/master/redis-vl)):

```python
import asyncio
import struct

from redis.asyncio.cluster import ClusterNode, RedisCluster
from redisvl.index import AsyncSearchIndex
from redisvl.schema import IndexSchema


SCHEMA = IndexSchema.from_dict(
    {
        "index": {
            "name": "user-index",
            "prefix": "user:",
        },
        "fields": [
            {"name": "user", "type": "tag"},
            {"name": "credit_score", "type": "tag"},
            {
                "name": "embedding",
                "type": "vector",
                "attrs": {
                    "algorithm": "flat",
                    "dims": 3,
                    "distance_metric": "cosine",
                    "datatype": "float32",
                },
            },
        ],
    }
)


async def client():
    index = AsyncSearchIndex(
        SCHEMA,
        redis_client=RedisCluster(
            startup_nodes=[ClusterNode(host="redis-1", port=6379)],
            username=None,
            password=None,
            client_name="client-issue",
        ),
    )
    await index.client.initialize()
    await index.create()
    keys = await index.load(
        [
            {
                "user": "bob",
                "credit_score": "good",
                "embedding": struct.pack("3f", 0.1, 0.2, 0.3),
            }
        ],
        keys=["user:bob"],
        ttl=3600
    )
    print(f"Inserted: {keys}", flush=True)


asyncio.run(client())
```

Here, nothing is inserted because [`await pipe.expire()`](https://github.com/redis/redis-vl-python/blob/v0.10.0/redisvl/index/storage.py#L493) triggers [`ClusterPipeline.__await__`](https://github.com/redis/redis-py/blob/v6.4.0/redis/asyncio/cluster.py#L1587) which [empties the command queue](https://github.com/redis/redis-py/blob/v6.4.0/redis/asyncio/cluster.py#L1807).

What is wrong here? 

- If there is something wrong in my code example, would be nice to get an error during the initialization.
- If this is a bug in `redis-vl`, is it the `await` to remove in `await pipe.expire()`, or something else?
- By the way, what is the purpose of `ClusterPipeline.__await__` in `redis-py`?

dfroger:
BEFORE:
- `await pipe.expire()` triggers `ClusterPipeline.__await__`
- this triggers `.initialize()` which delete the pipeline commands

AFTER:
- `pipe.expire()` in case of pipeline
- `await pipe.expire()` in case of classic client

Here is a first try to fix https://github.com/redis/redis-vl-python/issues/417 .

Tests and lint are ok. However I failed to run `tests/integration/test_cluster_pipelining.py`, cluster doesn't seems to start ([test.log](https://github.com/user-attachments/files/23353596/test.log)).

### DIFF
--- a/redisvl/index/storage.py
+++ b/redisvl/index/storage.py
@@ -176,7 +176,7 @@ class BaseStorage(BaseModel):
         raise NotImplementedError
 
     @staticmethod
-    async def _aexpire(client: AsyncRedisClientOrPipeline, key: str, ttl: int):
+    async def _aexpire(client: AsyncRedisClientOrPipeline, key: str, ttl: int) -> None:
         """Asynchronously set TTL on a key using the provided client or pipeline
 
         Args:


abrookins:
Thanks @dfroger! We'll take a look ASAP.

redisvl/index/storage.py
```diff
@@ -175,6 +175,20 @@ async def _aget(
         """Asynchronously get data from Redis using the provided client or pipeline."""
         raise NotImplementedError
 
+    @staticmethod
+    async def _aexpire(client: AsyncRedisClientOrPipeline, key: str, ttl: int):
```
nkanu17:
```suggestion
    async def _aexpire(client: AsyncRedisClientOrPipeline, key: str, ttl: int) -> None:
```

Issue #431, opened 2025-11-20
Bug: In redisvl FT.CREATE fails when using INDEXMISSING, SORTABLE, and UNF together


nkanu17:
## Summary
`TextField` with `index_missing=True`, `sortable=True`, and `unf=True` fails at index creation due to incorrect field option ordering
When a RedisVL `TextField` is configured with all three of:

- `index_missing=True`
- `sortable=True`
- `unf=True`

the generated `FT.CREATE` field modifiers are ordered incorrectly:

- Generated: `TEXT ... SORTABLE UNF INDEXMISSING` → Redis rejects this.
- Required: `TEXT ... INDEXMISSING SORTABLE UNF` → Redis accepts this.

RediSearch’s parser currently requires `INDEXEMPTY` and `INDEXMISSING` to appear **before** `SORTABLE`/`UNF` in the field definition (see redis/redis#5177). RedisVL’s `TextField.as_redis_field()` does not enforce that ordering, causing `FT.CREATE` to fail with:

- `ResponseError: Field 'INDEXMISSING' does not have a type`.

A similar issue may affect other field types that support `index_missing` / `index_empty` and `sortable` / `unf` (e.g., `NumericField`, possibly `TagField` / `GeoField`).

---

## Current Behavior

- `TextField.as_redis_field()` (in `redisvl/schema/fields.py`) assembles kwargs for redis-py’s `TextField`, including:
  - `sortable` (bool)
  - `index_missing` (bool)
  - `index_empty` (bool)
  - `no_index` (bool)
- After constructing the redis-py `TextField`, it then mutates `field.args_suffix` to insert `UNF` relative to `SORTABLE`.
- With `index_missing=True`, `sortable=True`, `unf=True`, the resulting suffix is:
  - `["SORTABLE", "UNF", "INDEXMISSING"]`.
- This produces an `FT.CREATE` field definition equivalent to:
  - `... TEXT SORTABLE UNF INDEXMISSING`,
  which RediSearch parses as if `INDEXMISSING` were a field name, hence:
  - `ResponseError: Field 'INDEXMISSING' does not have a type`.
- There is a small script in the repo root (`test_unf_sortable.py`) that demonstrates this; its final test case fails with precisely this error and prints the problematic suffix ordering.

---

## Expected Behavior

- A `TextField` with `index_missing=True`, `sortable=True`, and `unf=True` should generate field modifiers in an order acceptable to RediSearch, e.g.:
[...]

PR #434, opened 2025-11-20
fix: (schema) ensure field modifiers follow canonical order for RediSearch parser

nkanu17:
Fix field modifier ordering to satisfy RediSearch parser requirements where
INDEXEMPTY and INDEXMISSING must appear BEFORE SORTABLE in field definitions.
This resolves index creation failures when using index_missing=True with
sortable=True.

Changes:
- Add _normalize_field_modifiers() helper function with set-based optimization
- Refactor TextField, TagField, NumericField, GeoField to use helper
- Implement canonical order: [INDEXEMPTY] [INDEXMISSING] [SORTABLE [UNF]] [NOINDEX]
- Add unit tests for helper and field classes
- Add integration tests against live Redis

Fixes: issue [#431](https://github.com/redis/redis-vl-python/issues/431#issue-3648415728)
Related: Field 'INDEXMISSING' does not have a type error

BREAKING CHANGE: None - backward compatible, only changes internal ordering

### DIFF
--- a/tests/integration/test_field_modifier_ordering_integration.py
+++ b/tests/integration/test_field_modifier_ordering_integration.py
@@ -297,7 +297,13 @@ class TestMultiFieldModifierOrderingIntegration:
         assert info is not None
 
         # Verify all fields were created
-        attrs_list = info[7]
+        # Find the 'attributes' key in the FT.INFO response (flat list format)
+        attrs_list = None
+        for i in range(0, len(info) - 1, 2):
+            if info[i] == b"attributes" or info[i] == "attributes":
+                attrs_list = info[i + 1]
+                break
+        assert attrs_list is not None, "'attributes' key not found in FT.INFO response"
         assert len(attrs_list) == 4
 
         # Cleanup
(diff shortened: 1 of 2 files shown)


tests/integration/test_field_modifier_ordering_integration.py
```diff
@@ -0,0 +1,473 @@
...
+the FT.CREATE command.
+"""
+
+import pytest
+
+from redisvl.index import SearchIndex
+from redisvl.redis.connection import RedisConnectionFactory
+from redisvl.schema import IndexSchema
+
+MIN_SEARCH_VERSION_FOR_INDEXMISSING = 21000  # RediSearch 2.10.0+
```
jeremyplichta: Please test how this performs with newer version of Redis, like Redis 8. Starting in Redis 8 the versioning of search is different. Here is output of `info modules` in a local redis 8.4 docker image

```
module:name=search,ver=80205,api=1,filters=0,usedby=[],using=[ReJSON],options=[handle-io-errors]
```

Cloud will start running 8+ more and more often now as well now that its available (pretty sure).

jeremyplichta: This is likely a larger issue across all or redisvl and I'm ok if we track this as a separate issue to as to not hold this up. The current customer we are doing this for is not on 8 yet.

PR #429, opened 2025-11-18
Add LangCache integration tests

abrookins:
This PR adds integration tests for our new integration with LangCache, the managed cache service.

PR #437, opened 2025-11-21
Decode LangCache attribute values on retrieval

abrookins:
In #429, we found and fixed an issue with LangCache attribute values, which is that some characters are not allowed (commas) and some are accepted but silently fail during retrieval (slashes). Our solution was to encode these characters when we persist them and at query time. This PR decodes the values in retrieved data as well.


bsbodden (approved):
LGTM! Send it!

PR #438, opened 2025-11-21
Fix more problematic attribute values for LangCache

abrookins:
Following on #437, we find and encode/decode more problematic characters used within attribute values in LangCache. This PR also adds a comprehensive punctuation test for other characters.


bsbodden (approved):
Ship it!

PR #445, opened 2025-12-02
Avoid configuring root logger in RedisVL

### LINKED ISSUE #443: Logging configuration affects application loggers
## Issue

The `redisvl/utils/log.py` module configures logging by adding a root logger, which can interfere with other loggers in an application. For example, importing `redisvl` can cause app logs to appear twice.

A previous PR (https://github.com/redis/redis-vl-python/pull/301) added a guard to avoid adding a root logger if handlers exist, but in my case I don’t want any logs from `redisvl` in production.

The Python documentation strongly advises that libraries don't log to the root logger or add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) (https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library):

> It is strongly advised that you do not log to the root logger in your library. Instead, use a logger with a unique and easily identifiable name, such as the __name__ for your library’s top-level package or module. Logging to the root logger will make it difficult or impossible for the application developer to configure the logging verbosity or handlers of your library as they wish.

> It is strongly advised that you do not add any handlers other than [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements.

This is the code that's adding a root logger:
https://github.com/redis/redis-vl-python/blob/32698e990e4baab553a9b379f4d175e6a9e48ed2/redisvl/utils/log.py#L13-L21

Reproducing the issue:
```python
import logging

app_logger = logging.getLogger("my_app")
app_logger.setLevel(logging.INFO)
app_logger.addHandler(logging.StreamHandler())

print("==== App logging works as expected ====")
app_logger.info("This is an INFO message from the app logger.")
print()

from redisvl.query.filter import Tag

print("==== Now app logging is affected ====")
app_logger.info("This message appears TWICE!")
```

Output:
```
==== App logging works as expected ====
This is an INFO message from the app logger.

==== Now app logging is affected ====
This message appears TWICE!
15:59:44 my_app INFO   This message appears TWICE!
```

## Proposed changes

Follow the recommendations from the Python documentation:
* Don't attach a root logger
* Only attach a [NullHandler](https://docs.python.org/3/library/logging.handlers.html#logging.NullHandler) to `redisvl` loggers

If this isn't possible for some reason, please let me know how I can work around this.

vishal-bala:
Closes #443 

This PR updates the logger configuration helper for RedisVL to avoid configuring the root logger via `logging.basicConfig`. That operation was interfering with user-configured loggers, causing app logs to be emitted multiple times and making it trickier to hide RedisVL logs when running in production.

In its place, we now configure a `NullHandler` (mainly to avoid issues with loggers not having any handlers configured). **This means that any RedisVL usage that expects logs to have a specific format must now configure that format for itself.**

### DIFF
--- a/redisvl/utils/log.py
+++ b/redisvl/utils/log.py
@@ -11,7 +11,5 @@ def get_logger(name, log_level="info", fmt=None):
 
     # Add a NullHandler to loggers to avoid "no handler found" warnings
     if not logger.handlers:
-        logger.addHandler(
-            logging.NullHandler()
-        )
+        logger.addHandler(logging.NullHandler())
     return logger


vishal-bala:
@tylerhutcherson Yup! There was an existing test for roughly the same case that I extended to test that we don't emit duplicated logs anymore (it's very similar to the reproducing steps the issuer mentioned - see changes to `logger_interference_checker.py` and the associated test)

PR #447, opened 2025-12-05
Implement interface for unified hybrid search in Redis 8.4.0+

vishal-bala:
This PR adds an interface for using the new [`FT.HYBRID`](https://redis.io/docs/latest/commands/ft.hybrid) search operation introduced in Redis Open Source 8.4.0.

### `redisvl.query.hybrid`
This new module defines query constructors for working with the new hybrid search operation. Attempting to import this module without having `redis-py>=7.1.0` will raise an `ImportError`. It defines a `HybridQuery` object which can be used to define a hybrid search operation in a format largely similar to what `AggregateHybridQuery` provides.

Under the hood, `redis-py` expects three separate configuration objects to run hybrid search - one for the text+vector searches, one for the score combination, and one for post-processing of the results (e.g. aggregations, loading, limiting, etc.). The `HybridQuery` class manages the definition of all three.

```python
# old
from redisvl.query.aggregate import AggregateHybridQuery

query = AggregateHybridQuery(
    text="medical professional",
    text_field_name="description",
    vector=[0.1, 0.1, 0.5, ...],
    vector_field_name="user_embedding",
    alpha=0.7,  # LINEAR only
    return_fields=["user", "job"]
)

results = index.query(query)

# new
from redisvl.query.hybrid import HybridQuery

query = HybridQuery(
    text="medical professional",
    text_field_name="description",
    vector=[0.1, 0.1, 0.5, ...],
    vector_field_name="user_embedding",
    combination_method="LINEAR",
    linear_alpha=0.3,  # NOTE: The linear combination is defined inconsistently between the two approaches
    return_fields=["user", "job"]
)

results = index.hybrid_search(query)
results = await async_index.hybrid_search(query)
```

### SearchIndex method
For `redis-py>=7.1.0`, the `SearchIndex` and `AsyncSearchIndex` classes define a `hybrid_search` method that takes a `HybridQuery` instance as an argument. For `redis-py<7.1.0` the method raises a `NotImplementedError`. The method returns a list of retrieved dictionaries.

### Additional Changes
- Fixed a typo in a test for `AggregateHybridQuery`
- `redis-py 7.0.0` introduced a change to the type of `Query._fields`, changing it from a tuple to a list - a test had to be updated to differentiate the expectation based on the redis-py version.

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -1030,6 +1030,18 @@ wheels = [
 ]
 
 [[package]]
+name = "ffmpeg-python"
+version = "0.2.0"
+source = { registry = "https://pypi.org/simple" }
+dependencies = [
+    { name = "future" },
+]
+sdist = { url = "https://files.pythonhosted.org/packages/dd/5e/d5f9105d59c1325759d838af4e973695081fbbc97182baf73afc78dec266/ffmpeg-python-0.2.0.tar.gz", hash = "sha256:65225db34627c578ef0e11c8b1eb528bb35e024752f6f10b78c011f6f64c4127", size = 21543, upload-time = "2019-07-06T00:19:08.989Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/d7/0c/56be52741f75bad4dc6555991fabd2e07b432d333da82c11ad701123888a/ffmpeg_python-0.2.0-py3-none-any.whl", hash = "sha256:ac441a0404e053f8b6a1113a77c0f452f1cfc62f6344a769475ffdc0f56c23c5", size = 25024, upload-time = "2019-07-06T00:19:07.215Z" },
+]
+
+[[package]]
 name = "filelock"
 version = "3.19.1"
 source = { registry = "https://pypi.org/simple" }
@@ -1168,6 +1180,15 @@ source = { registry = "https://pypi.org/simple" }
 sdist = { url = "https://files.pythonhosted.org/packages/de/e0/bab50af11c2d75c9c4a2a26a5254573c0bd97cea152254401510950486fa/fsspec-2025.9.0.tar.gz", hash = "sha256:19fd429483d25d28b65ec68f9f4adc16c17ea2c7c7bf54ec61360d478fb19c19", size = 304847, upload-time = "2025-09-02T19:10:49.215Z" }
 wheels = [
     { url = "https://files.pythonhosted.org/packages/47/71/70db47e4f6ce3e5c37a607355f80da8860a33226be640226ac52cb05ef2e/fsspec-2025.9.0-py3-none-any.whl", hash = "sha256:530dc2a2af60a414a832059574df4a6e10cce927f6f4a78209390fe38955cfb7", size = 199289, upload-time = "2025-09-02T19:10:47.708Z" },
+]
+
+[[package]]
+name = "future"
+version = "1.0.0"
+source = { registry = "https://pypi.org/simple" }
+sdist = { url = "https://files.pythonhosted.org/packages/a7/b2/4140c69c6a66432916b26158687e821ba631a4c9273c474343badf84d3ba/future-1.0.0.tar.gz", hash = "sha256:bd2968309307861edae1458a4f8a4f3598c03be43b97521076aebf5d94c07b05", size = 1228490, upload-time = "2024-02-21T11:52:38.461Z" }
+wheels = [
+    { url = "https://files.pythonhosted.org/packages/da/71/ae30dadffc90b9006d77af76b393cb9dfbfc9629f339fc1574a1c52e6806/future-1.0.0-py3-none-any.whl", hash = "sha256:929292d34f5872e70396626ef385ec22355a1fae8ad29e1a734c3e43f9fbc216", size = 491326, upload-time = "2024-02-21T11:52:35.956Z" },
 ]
 
 [[package]]
@@ -5809,11 +5830,12 @@ wheels = [
 
 [[package]]
 name = "voyageai"
-version = "0.3.5"
+version = "0.3.6"
 source = { registry = "https://pypi.org/simple" }
 dependencies = [
     { name = "aiohttp" },
     { name = "aiolimiter" },
+    { name = "ffmpeg-python" },
     { name = "langchain-text-splitters", version = "0.3.11", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.10'" },
     { name = "langchain-text-splitters", version = "1.0.0", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version >= '3.10'" },
     { name = "numpy", version = "2.0.2", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.10'" },
@@ -5826,9 +5848,9 @@ dependencies = [
     { name = "tenacity" },
     { name = "tokenizers" },
 ]
-sdist = { url = "https://files.pythonhosted.org/packages/51/9b/e40f90793c1d03610b6109852791f752fcb257989a96701258278f874e00/voyageai-0.3.5.tar.gz", hash = "sha256:963e0d71611af529fa0e496db232a4f660b5f73bce7af1ab288a7f59df7512da", size = 20414, upload-time = "2025-09-11T00:28:26.29Z" }
+sdist = { url = "https://files.pythonhosted.org/packages/f3/a6/5f93fcd9c8c1a05873d287f1600e04f6cbfb2d42fce15ed75b0d5bebb9fa/voyageai-0.3.6.tar.gz", hash = "sha256:411e7c11eae4917429f091553a9b6911860df626811fb9415f24197d0cc5e219", size = 26346, upload-time = "2025-12-09T01:32:52.278Z" }
 wheels = [
-    { url = "https://files.pythonhosted.org/packages/8a/9d/709f5c7fc80a7bf11952fbccfca2bc5525bd5d345521795358819bd01d02/voyageai-0.3.5-py3-none-any.whl", hash = "sha256:1f70fcf3532d7e0bbc4332b1831a6fc1f714f268eeddc8b2859b81bf06a82411", size = 28257, upload-time = "2025-09-11T00:28:24.62Z" },
+    { url = "https://files.pythonhosted.org/packages/bf/e9/e13785fb2a3c605ea924ce2e54d235e423f6d6e45ddb09574963655ec111/voyageai-0.3.6-py3-none-any.whl", hash = "sha256:e282f9cef87eb949e2dd30ffe911689f1068c50b8c3c6e90e97793f2a52c83dd", size = 34465, upload-time = "2025-12-09T01:32:51.32Z" },
 ]
 
 [[package]]


nkanu17 (commented):
You may have forgotten to export HybridQuery from redisvl.query.hybrid in the redisvl.query `__init__.py`

```
from redisvl.query.hybrid import HybridQuery  # this will work
from redisvl.query import HybridQuery  # this will fail
```

Basically because there are now two different HybridQuery classes
- redisvl.query.aggregate.HybridQuery (old wrapper around AggregateHybridQuery)
- redisvl.query.hybrid.HybridQuery (new one)

This might cause some confusion and we might have to either choose a different name or find a better solution and deprecate the older one. Might change the name of the older class

nkanu17 (commented):
Looks great overall!
Just need to verify how we need to handle the deprecated HybridQuery class
Don't forget to update the api/sphinx documentation and add a guide in ai-resources

.github/workflows/test.yml
```diff
@@ -96,8 +96,8 @@ jobs:
       matrix:
         # 3.11 tests are run in the service-tests job
         python-version: ["3.9", "3.10", "3.12", "3.13"] 
-        redis-py-version: ["5.x", "6.x"]
-        redis-version: ["6.2.6-v9", "latest", "8.0.2"]
+        redis-py-version: ["5.x", "6.x", "7.x"]
+        redis-version: ["6.2.6-v9", "latest", "8.0.2", "8.4.0"]
```
rbs333: can probably drop "8.0.2" since adding new version since there's nothing special about that one.

.github/workflows/test.yml
```diff
@@ -130,13 +130,15 @@ jobs:
...
+          elif [[ "${{ matrix.redis-py-version }}" == "6.x" ]]; then
             uv pip install "redis>=6,<7"
+          else
+            uv pip install "redis>=7,<8"
           fi
 
       - name: Set Redis image name
         run: |
-          if [[ "${{ matrix.redis-version }}" == "8.0.2" ]]; then
+          if [[ "${{ matrix.redis-version }}" == "8.0.2" || "${{ matrix.redis-version }}" == "8.4.0" ]]; then
```
rbs333: see prior comment

docs/api/query.rst
```diff
@@ -105,47 +105,73 @@ VectorRangeQuery
...
    :show-inheritance:
    :exclude-members: add_filter,get_args,highlight,return_field,summarize
 
 .. note::
-   The ``stopwords`` parameter in :class:`HybridQuery` (and :class:`AggregateHybridQuery`) controls query-time stopword filtering (client-side).
+   The ``stopwords`` parameter in :class:`AggregateHybridQuery` (and :class:`HybridQuery`) controls query-time stopword filtering (client-side).
    For index-level stopwords configuration (server-side), see :class:`redisvl.schema.IndexInfo.stopwords`.
    Using query-time stopwords with index-level ``STOPWORDS 0`` is counterproductive.
 
+.. note::
```
rbs333, line 126: I think we call this linear_alpha vs linear_beta in the actual below... I think we should opt for whatever aligns with the query syntax itself

redisvl/index/index.py
```diff
@@ -1003,6 +1023,56 @@ def search(self, *args, **kwargs) -> "Result":
         except Exception as e:
             raise RedisSearchError(f"Unexpected error while searching: {str(e)}") from e
 
+    def hybrid_search(self, query: Any, **kwargs) -> List[Dict[str, Any]]:
```
justin-cechmanek: We have the `query()` method that accepts all our different `query` objects types. Let's use that as the way to run all our queries. We can do a query type check in `query()` like we do now and change this to a private method to call from `query()`

redisvl/index/index.py
```diff
@@ -1824,6 +1894,53 @@ async def search(self, *args, **kwargs) -> "Result":
         except Exception as e:
             raise RedisSearchError(f"Unexpected error while searching: {str(e)}") from e
 
+    async def hybrid_search(self, query: Any, **kwargs) -> List[Dict[str, Any]]:
```
justin-cechmanek: Same change here. Let's instead expand our async `query()` method to something like 
```
async def query(self, query: Union[BaseQuery, AggregationQuery]) -> List[Dict[str, Any]]:
  if isinstance(query, AggregationQuery):
     return self._aggregate(query)
  elif isinstance(HybridQuery):
     return self._hybrid_search(query)
  else:
     return self._query(query)
```

redisvl/query/hybrid.py
```diff
@@ -0,0 +1,363 @@
...
+    )
+    from redis.commands.search.hybrid_query import HybridQuery as RedisHybridQuery
+    from redis.commands.search.hybrid_query import (
+        HybridSearchQuery,
+        HybridVsimQuery,
+        VectorSearchMethods,
+    )
+except ImportError:
+    raise ImportError(
+        "Hybrid queries require Redis Open Source >= 8.4.0 and redis-py>=7.1.0"
```
justin-cechmanek:
```suggestion
        "Hybrid queries require Redis >= 8.4.0 and redis-py>=7.1.0"
```
This works on enterprise Redis too :)

redisvl/query/hybrid.py
```diff
@@ -0,0 +1,363 @@
...
+        self,
+        text: str,
+        text_field_name: str,
+        vector: Union[bytes, List[float]],
+        vector_field_name: str,
+        text_scorer: str = "BM25STD",
+        text_filter_expression: Optional[Union[str, FilterExpression]] = None,
+        yield_text_score_as: Optional[str] = None,
+        vector_search_method: Optional[Literal["KNN", "RANGE"]] = None,
+        knn_k: Optional[int] = None,
```
justin-cechmanek: In our other query objects, including `RangeQuery` and `VectorRangeQuery` we have the `num_results` parameter. We can add consistency by using the same parameter here.

redisvl/query/hybrid.py
```diff
@@ -0,0 +1,363 @@
...
+        knn_ef_runtime: Optional[int] = None,
+        range_radius: Optional[int] = None,
+        range_epsilon: Optional[float] = None,
+        yield_vsim_score_as: Optional[str] = None,
+        vector_filter_expression: Optional[Union[str, FilterExpression]] = None,
```
justin-cechmanek:
```suggestion
        filter_expression: Optional[Union[str, FilterExpression]] = None,
```
for consistency with our other query objects

redisvl/index/index.py
rbs333: Working with this a bit more now I'm not sure that I love having a separate method for hybrid queries. I would kind of like to just pass a HybridQuery to the standard .query method.

tylerhutcherson: agreed. we should not need to break the searchindex interface

redisvl/query/hybrid.py
justin-cechmanek: having both a `text_filter_expression` and `vector_filter_expression` feels redundant. Can we change this to only use one `filter_expression` parameter like our other Query classes?

redisvl/query/hybrid.py
```diff
@@ -0,0 +1,363 @@
...
+        combination_method: Optional[Literal["RRF", "LINEAR"]] = None,
+        rrf_window: Optional[int] = None,
+        rrf_constant: Optional[float] = None,
+        linear_alpha: Optional[float] = None,
+        linear_beta: Optional[float] = None,
```
justin-cechmanek: having an alpha and beta parameter looks redundant, and from a user's point of view it's not clear how they would interact. Feels like a foot gun

redisvl/query/hybrid.py
```diff
@@ -0,0 +1,363 @@
...
+        vector_similarity_query=vsim_query,
+    )
+
+
+def build_combination_method(
+    combination_method: Literal["RRF", "LINEAR"],
+    rrf_window: Optional[int] = None,
+    rrf_constant: Optional[float] = None,
+    linear_alpha: Optional[float] = None,
+    linear_beta: Optional[float] = None,
```
justin-cechmanek: Similar to my previous comment, can we reduce this to one parameter only?

tests/integration/test_hybrid.py
```diff
@@ -0,0 +1,513 @@
...
+    filter_expression = (Text(text_field) == ("medical")) & (
+        (Text(text_field) != ("research"))
+    )
+    hybrid_query = HybridQuery(
+        text=text,
+        text_field_name=text_field,
+        vector=vector,
+        vector_field_name=vector_field,
+        text_filter_expression=filter_expression,
+        vector_filter_expression=filter_expression,
```
justin-cechmanek: Mentioned earlier, but it's not clear to a user how or why to pass two filter expressions.

tests/unit/test_hybrid_types.py
```diff
@@ -0,0 +1,1044 @@
...
+        "SEARCH",
+        "(~@description:(toon | squad | play | basketball | gang | aliens) AND @category:{movies})",
+        "SCORER",
+        "BM25STD",
+        "VSIM",
+        "@embedding",
+        bytes_vector,
+        "FILTER",
+        "@genre:{comedy}",
+    ]
```
justin-cechmanek: Same comments about having two filter_expressions

docs/api/query.rst
```diff
@@ -105,47 +105,73 @@ VectorRangeQuery
...
+      from redisvl.query import HybridQuery
 
-      query = VectorQuery(
+      query = HybridQuery(
+          text="query string",
+          text_field_name="description",
           vector=[0.1, 0.2, 0.3],
           vector_field_name="embedding",
+          vector_search_method="KNN",
+          knn_num_results=10,
```
justin-cechmanek: Looks like we just forgot one reference to knn_num_results in our doc string

redisvl/index/index.py
```diff
@@ -1824,6 +1893,55 @@ async def search(self, *args, **kwargs) -> "Result":
...
+
+        """
+        client = await self._get_client()
+        index = client.ft(self.schema.index.name)
+        self._validate_hybrid_query(query)
+
+        if not hasattr(index, "hybrid_search"):
+            raise ImportError(_HYBRID_SEARCH_ERROR_MESSAGE)
+
+        results = await index.hybrid_search(
```
justin-cechmanek, line 1933:
```suggestion
        results = await index._hybrid_search(
```
I think this was missed in the last change

redisvl/index/index.py
```diff
@@ -1003,6 +1018,56 @@ def search(self, *args, **kwargs) -> "Result":
...
+            results = index.query(hybrid_query)
+
+        """
+        index = self._redis_client.ft(self.schema.index.name)
+        self._validate_hybrid_query(query)
+
+        if not hasattr(index, "hybrid_search"):
+            raise ImportError(_HYBRID_SEARCH_ERROR_MESSAGE)
+
+        results = index.hybrid_search(
```
justin-cechmanek, line 1059:
```suggestion
        results = index._hybrid_search(
```
missing the underscore?

vishal-bala: This is actually calling a method defined on the `redis-py` search index, so it doesn't need to change 👍

redisvl/index/index.py
vishal-bala: This is also calling a method defined on the `redis-py` search index, so it doesn't need to change 👍

docs/api/query.rst
vishal-bala: Oops, good catch! 😅

Issue #449, opened 2025-12-12
Timestamp collision in ChatMessage


austinmyc:
In `message_history.schema.ChatMessage`, the id is being generated with the session tag and the current timestamp.
```python
@model_validator(mode="before")
    @classmethod
    def generate_id(cls, values):
        if TIMESTAMP_FIELD_NAME not in values:
            values[TIMESTAMP_FIELD_NAME] = current_timestamp()
        if ID_FIELD_NAME not in values:
            values[ID_FIELD_NAME] = (
                f"{values[SESSION_FIELD_NAME]}:{values[TIMESTAMP_FIELD_NAME]}"
            )
        return values
```
This means that during many use cases, like `add_messages(messages)` or `store(prompt, response)` that uses `add_messages` fail to insert the full conversation to the database, as messages might end up having the same id, and newest message would overwrite the older messages.

A simple fix might be adding a counter or a hash for the id:
```python
values[ID_FIELD_NAME] = f"{session_tag}:{timestamp}:{uuid4().hex[:8]}"
```

nkanu17:
@austinmyc Thank you for reporting this! [Here is the fix](https://github.com/redis/redis-vl-python/commit/caa962173dd6401b282b047e82ea022174a24109).

It will be part of the next release!

PR #450, opened 2025-12-15
fix: prevent timestamp collision in ChatMessage ID generation (issue #449)

nkanu17:
Issue[ #449 ](https://github.com/redis/redis-vl-python/issues/449)

In redisvl/extensions/message_history/schema.py, the ChatMessage class generates IDs using only the session tag and timestamp:
`values[ID_FIELD_NAME] = f"{values[SESSION_FIELD_NAME]}:{values[TIMESTAMP_FIELD_NAME]}"`

When multiple messages are added rapidly (e.g., via add_messages(messages) or store(prompt, response)), they can receive the same timestamp, resulting in identical IDs. Since Redis uses these IDs as keys, newer messages overwrite older ones with the same ID, causing message loss.

Fix:
- Add UUID suffix to ChatMessage entry_id to ensure uniqueness
- Update ID format from 'session:timestamp' to 'session:timestamp:uuid'
- Add test for rapid message creation with same timestamp


justin-cechmanek (approved):
👍🏼

abrookins (commented):
LGTM 👍

PR #454, opened 2025-12-16
RAAE-1309: add support for wildcard TAG filters

### LINKED ISSUE #453: RedisVL's Tag filter does not support wildcards
TAG fields support wildcard search with a star operator, but attempting to search this way using Tag from RedisVL produces an escaped operator, with no way to prevent that escaping. We should instead support either a flag set on init, or else a new modulo operator, so that a user can perform a wildcard TAG search (i.e. prefix search for tags).

fixes RAAE-1309

abrookins:
Allow using modulo queries to enable wildcard (prefix search) support in `Tag` filters, e.g.:

```python
wildcard_match = Tag("category") % "tech*"
```

closes #453 / RAAE-1309

### DIFF
--- a/redisvl/utils/token_escaper.py
+++ b/redisvl/utils/token_escaper.py
@@ -12,13 +12,30 @@ class TokenEscaper:
     # Source: https://redis.io/docs/stack/search/reference/escaping/#the-rules-of-text-field-tokenization
     DEFAULT_ESCAPED_CHARS = r"[,.<>{}\[\]\\\"\':;!@#$%^&*()\-+=~\/ ]"
 
+    # Same as above but excludes * to allow wildcard patterns
+    ESCAPED_CHARS_NO_WILDCARD = r"[,.<>{}\[\]\\\"\':;!@#$%^&()\-+=~\/ ]"
+
     def __init__(self, escape_chars_re: Optional[Pattern] = None):
         if escape_chars_re:
             self.escaped_chars_re = escape_chars_re
         else:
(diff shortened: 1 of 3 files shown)


redisvl/utils/token_escaper.py
```diff
@@ -12,13 +12,30 @@ class TokenEscaper:
     # Source: https://redis.io/docs/stack/search/reference/escaping/#the-rules-of-text-field-tokenization
     DEFAULT_ESCAPED_CHARS = r"[,.<>{}\[\]\\\"\':;!@#$%^&*()\-+=~\/ ]"
 
+    # Same as above but excludes * to allow wildcard patterns
+    ESCAPED_CHARS_NO_WILDCARD = r"[,.<>{}\[\]\\\"\':;!@#$%^&()\-+=~\/ ]"
```
vishal-bala, line 16: Is there a reason we'd prefer to define this in its entirety instead of as
```suggestion
    escaped_chars_no_wildcard = DEFAULT_ESCAPED_CHARS.replace("*", "")
```
?

Both make sense, just wondering about the thinking behind it!

vishal-bala (approved):
LGTM 👍

tylerhutcherson (approved):
looks good, thanks!

redisvl/utils/token_escaper.py
abrookins: I don't have a strong preference, but I prefer to see the pattern explicitly even if it's mostly a duplicate.

PR #455, opened 2025-12-17
fix: Use parameter substitution for vector data in hybrid search

vishal-bala:
Support for inlined blobs was dropped in FT.HYBRID due to some sequences being misinterpreted (see https://github.com/RediSearch/RediSearch/pull/7768). This PR updates the hybrid search implementation to pass vector data to hybrid searches via parameter substitution instead.


bsbodden (approved):
LGTM!

bsbodden:
@vishal-bala Let's stick to conventional commits for our repos https://www.conventionalcommits.org/en/v1.0.0/ (you can just give the PR name a conventional commit and we can "squash and merge"

PR #452, opened 2025-12-16
feat: Add support for multimodal embeddings in vectorizers

vishal-bala:
This PR generalizes the `BaseVectorizer` to be agnostic to any modality (since it previously exclusively supported text inputs). Building from the new base, this PR then extends the implementation for some vectorizers to support multimodal embeddings (renaming them away from being specifically for text).

## `BaseVectorizer`

The move away from having the `BaseVectorizer` explicitly expect text inputs means a change in the signature of the `embed` methods away from `vectorizer.embed(text="lorem ipsum...")` to `vectorizer.embed(content="lorem ipsum...")`. This is a breaking change for existing usages of the vectorizers that use the keyword argument, and the usages will need to be updated to align with the new schema.

Caching for multimodal embeddings is supported for all vectorizers introduced in this PR.

## Multimodal Implementations

The following vectorizers have been renamed to no longer be explicitly text vectorizers, and moved to no longer be defined in the `vectorize.text` module. Imports and usages for these vectorizers will need to be updated to avoid errors. The `CustomTextVectorizer` has also been renamed and moved to be `redisvl.utils.vectorize.custom.CustomVectorizer`.

### VoyageAI

Old: `redisvl.utils.vectorize.text.voyageai.VoyageAITextVectorizer`
New: `redisvl.utils.vectorize.voyageai.VoyageAIVectorizer`

```python
from redisvl.utils.vectorize import VoyageAIVectorizer

# --- Basic usage
vectorizer = VoyageAIVectorizer(
    model="voyage-3-large",
    api_config={"api_key": "your-voyageai-api-key"} # OR set VOYAGE_API_KEY in your env
)
query_embedding = vectorizer.embed(
    content="your input query text here",
    input_type="query"
)
doc_embeddings = vectorizer.embed_many(
    contents=["your document text", "more document text"],
    input_type="document"
)

# --- Multimodal usage - requires Pillow and voyageai>=0.3.6 (for video)
from PIL import Image
from voyageai.video_utils import Video

vectorizer = VoyageAIVectorizer(
    model="voyage-multimodal-3.5",
    api_config={"api_key": "your-voyageai-api-key"} # OR set VOYAGE_API_KEY in your env
)

# text
text_embedding = vectorizer.embed(
    content="your input query text here",
    input_type="query"
)

# image
image_embedding = vectorizer.embed_image(
    "path/to/your/image.jpg",
    input_type="query"
)
image_embedding = vectorizer.embed(
    Image.open("path/to/your/image.jpg"),
[...]

PR #457, opened 2025-12-19
fix: Lazy-import botocore exception to avoid unnecessary dependency for using `vectorize` module

vishal-bala:
PR #452 imports an exception from botocore outside of the function where it is needed, which unintentionally adds a dependency on botocore for the package. This PR fixes this issue to avoid importing from botocore outside of the function that needs the exception.


hanchoong-redis (approved):
Dang boi

PR #468, opened 2026-02-03
Add Python 3.14 support

abrookins:
## Summary

Add Python 3.14 support to RedisVL.

## Changes

- Update `requires-python` upper bound from `<3.14` to `<3.15`
- Add `Programming Language :: Python :: 3.14` classifier
- Add Python 3.14 to the CI test matrix

## Testing Strategy

**HuggingFace/sentence-transformers tests are skipped on Python 3.14** because the upstream `transformers` library doesn't fully support Python 3.14 yet.

Affected components (all optional):
- `HFTextVectorizer`
- `HFCrossEncoderReranker`
- `SemanticRouter` (when using default HF vectorizer)
- `SemanticMessageHistory` (when using HF vectorizer)

Users can still use these features on Python 3.9-3.13, or use alternative vectorizers (OpenAI, Cohere, etc.) on Python 3.14.

**Implementation:**
- Added `SKIP_HF` constant in `conftest.py` to detect Python 3.14+
- Skip HF-dependent test files entirely via `pytestmark`
- Skip individual HF tests via `@pytest.mark.requires_hf` marker
- Skip HF parameters in parametrized tests via `pytest.param(..., marks=...)`

Once `transformers` adds Python 3.14 support, these skips can be removed.

## Additional Fix

Added missing `@pytest.mark.requires_api_keys` marker to `test_default_dtype` (pre-existing bug unrelated to this PR).

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -4771,7 +4771,7 @@ wheels = [
 
 [[package]]
 name = "redisvl"
-version = "0.13.2"
+version = "0.14.0"
 source = { editable = "." }
 dependencies = [
     { name = "jsonpath-ng" },
@@ -4814,6 +4814,9 @@ pillow = [
 ]
 sentence-transformers = [
     { name = "sentence-transformers" },
+]
+sql-redis = [
+    { name = "sql-redis" },
 ]
 vertexai = [
     { name = "google-cloud-aiplatform" },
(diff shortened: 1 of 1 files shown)


tylerhutcherson:
@abrookins looks like we need to resolve `uv.lock` conflicts. But otherwise looks good. We can roll this along with some of the pending docs improvements to a clean `0.15.0` release.

vishal-bala (approved):
LGTM - seems a bit painful until sentence-transformers catches up 😄 

Also, maybe we should settle on a timeline for dropping Python 3.9 support, since it reached EoL back in October. No rush, but might let us simplify some things (and also reduce the matrix testing).

PR #471, opened 2026-02-11
Enable multiple prefixes per index

rbs333:
This PR let's you pass prefix as a list. It's not a recommend pattern but a blocker for a key customer. The main place this can cause issues in data loading.


tylerhutcherson:
@rbs333 Looks like this has leaked commits from the SQL work. Can you pare it back to the multi-prefix additions?

redisvl/index/index.py
vishal-bala, line 955: How involved are these to be run on every query? If they're non-negligible, do we want to cache/pre-load/lazy-load them?

vishal-bala (commented):
No major comments from me aside a question about whether building the registry/executor for SQL translation is expensive. There are a few Copilot comments that I agreed should be addressed (left a 👍 on them) - aside from those, it generally looks good! Super nice to be able to have SQL queries translated 🙌

redisvl/index/index.py
rbs333: Good callout. I've captured in this ticket https://redislabs.atlassian.net/browse/RAAE-1408 this is unrelated to this pr. I was working on this on the go so pushed before rebase which caused some muddy changes in.

rbs333: @vishal-bala note this pr was dirty with leftover from sql pr and is just to allow for multi-prefix.

rbs333: Sorry for the confusion

redisvl/index/index.py
```diff
@@ -302,12 +302,19 @@ def name(self) -> str:
...
-        first one."""
+        """The key prefix used in forming Redis keys.
+
+        For multi-prefix indexes, returns the first prefix.
+        """
         prefix = self.schema.index.prefix
         return prefix[0] if isinstance(prefix, list) else prefix
 
+    @property
+    def prefixes(self) -> List[str]:
```
vishal-bala, line 313: Is there a need for any extra validation here (e.g. if a prefix happens to be an empty string, etc.), or is that better left to raise an error somewhere else?

docs/user_guide/12_sql_to_redis_queries.ipynb
```diff
@@ -1,1142 +1,1142 @@
...
-     "metadata": {},
-     "output_type": "execute_result"
-    }
-   ],
-   "source": [
-    "# job is a tag field therefore this syntax works\n",
-    "sql_str = \"\"\"\n",
-    "    SELECT user, region, job, age\n",
-    "    FROM user_simple\n",
-    "    WHERE job IN ('software engineer', 'engineer', 'pancake tester')\n",
```
vishal-bala, line 543: A comment highlighting that there's a syntactic difference in the Redis translation of `region = 'a' OR region = 'b'` and `region IN ('a', 'b')` could be worthwhile (but not critical).

vishal-bala (commented):
LGTM!

vishal-bala (approved):

PR #477, opened 2026-02-17
fix: avoid NLTK download race condition in parallel test workers

bsbodden:
Try loading stopwords before downloading to prevent concurrent pytest-xdist workers from racing on nltk.download(), which raises FileExistsError when multiple processes write to the same path.

### DIFF
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -1460,8 +1460,14 @@ class TextQuery(BaseQuery):
             self._stopwords = set()
         elif isinstance(stopwords, str):
             try:
-                nltk.download("stopwords", quiet=True)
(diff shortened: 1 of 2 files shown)


vishal-bala (approved):
LGTM! I'd guess there's technically still a race condition in that
- two workers could try to start downloading at the same time, or
- a worker could try to load a partially downloaded dataset, fail, and then decide to start downloading it as well

But these are both fairly niche and unlikely, and probably not worth designing for.

rbs333 (commented):
Hey @vishal-bala @bsbodden Tyler and I had a conversation related to nltk at one point and had considered just dropping it from the project because A) it's bloated and B) we only use it for stop words really. 

We can probably still proceed with this pr just figured I'd call that out. CC: @tylerhutcherson

PR #478, opened 2026-02-19
Expose max_distance as an optional setting in multi vector queries

justin-cechmanek:


redisvl/query/aggregate.py
```diff
@@ -18,12 +18,20 @@
 class Vector(BaseModel):
     """
     Simple object containing the necessary arguments to perform a multi vector query.
+
+    Args:
+    vector: The vector values as a list of floats or bytes
+    field_name: The name of the vector field to search
+    dtype: The data type of the vector (default: "float32")
+    weight: The weight for this vector in the combined score (default: 1.0)
+    max_distance: The maximum distance for vector range search (default: 2.0, range: [0.0, 2.0])
```
nkanu17, line 27: Thanks for providing the range!

redisvl/query/aggregate.py
```diff
@@ -18,12 +18,20 @@
...
     """
 
     vector: Union[List[float], bytes]
     field_name: str
     dtype: str = "float32"
     weight: float = 1.0
+    max_distance: float = 2.0
```
vishal-bala: Instead of a custom validation method, you could validate this with Pydantic's built-in `Field` (you might have to `from pydantic import Field`)

```suggestion
    max_distance: float = Field(default=2.0, ge=0.0, le=2.0)
```

PR #479, opened 2026-02-20
Handle missing multivector distance

tomjohnhall:
When running a multi vector query, I noticed redis would sometimes return:

`Could not find the value for a parameter name, consider using EXISTS if applicable for distance_2`

This is running `redis-stack-server/7.4.0-v8` from brew.

I confirmed all documents contained the referenced embedding fields, but somehow the third distance was not always calculated and the query would error out. Maybe a match in other distances causes redis to skip the final distance calculation? Without debugging the specifics of query execution, this PR adds a case() in the query to set a default score of 0 for missing distances.

I also noticed that the vector radius is always hard coded to 2.0 - looks like there's a [separate PR](https://github.com/redis/redis-vl-python/pull/478) just opened to add custom ranges so I've left those changes out here. This fallback may solve issues downstream of that, and allow for safer handling of OR multi vectors of different ranges, rather than forcing AND.

### DIFF
--- a/redisvl/query/aggregate.py
+++ b/redisvl/query/aggregate.py
@@ -329,7 +329,11 @@ class MultiVectorQuery(AggregationQuery):
 
         # calculate the respective vector similarities
         for i in range(len(self._vectors)):
-            self.apply(**{f"score_{i}": f"(2 - @distance_{i})/2"})
+            self.apply(
+                **{
+                    f"score_{i}": f"case(exists(@distance_{i}), (2 - @distance_{i})/2, 0)"
+                }
+            )
 
         # construct the scoring string based on the vector similarity scores and weights
         combined_scores = []
--- a/tests/unit/test_aggregation_types.py
+++ b/tests/unit/test_aggregation_types.py
@@ -367,7 +367,7 @@ def test_multi_vector_query_string():
 
     assert (
         str(multi_vector_query)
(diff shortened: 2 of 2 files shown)


tomjohnhall:
I realise now that on a fresh instance, unstable features needs to be enabled to use case():

`FT.CONFIG SET ENABLE_UNSTABLE_FEATURES true`

So this change would be breaking until that feature becomes stable

vishal-bala:
Hey @tomjohnhall, thanks for jumping on that! You're right that these changes are breaking until the `case()` becomes part of the stable feature-set. The changes you have here are more of a workaround to deal with an unclear underlying issue - that some vectors are sometimes erroring out when you wouldn't expect them to.

If there isn't already one, could you raise an issue about that (along with a way to reproduce it, if you have one)? Once we have that, hopefully someone from our team, you, or someone else in the community can identify what's going on there and resolve it properly 🤞

tomjohnhall:
Thanks @vishal-bala I've narrowed it down to an issue of scale, not number of vectors / missing values. Raised issue [here](https://github.com/redis/redis-vl-python/issues/480)

justin-cechmanek:
hi @tomjohnhall we made some updates to handle multi vector distance that doesn't require the `case` statement [here](https://github.com/redis/redis-vl-python/commit/61f0f418b86c95aa5813a812a837ae92c65231db).
Do those changes meet your needs?

vishal-bala:
Hi @tomjohnhall, if there's nothing else to be done on this front, I'd like to close this PR - feel free to respond or reopen it if you have something in mind.

PR #532, opened 2026-03-12
feat(MCP): [1/5] Implement the base framework for a RedisVL MCP server

vishal-bala:
This PR implements the base framework for a RedisVL MCP server. In this iteration, the goal is to introduce the scaffolding for the MCP server to the extent that the server is runnable but does not have any meaningful functionality.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new MCP server layer that manages Redis connections, schema inspection/override logic, and request concurrency/timeouts; issues here could impact startup/shutdown reliability and Redis resource cleanup. Dependency changes also add a sizable optional stack (`fastmcp`), increasing integration surface area.
> 
> **Overview**
> Adds a new optional `redisvl.mcp` package that provides a runnable MCP server scaffold (`RedisVLMCPServer`) with explicit startup/shutdown lifecycle, guarded request execution (timeouts + max concurrency), and vectorizer initialization/cleanup.
> 
> Introduces YAML-based MCP configuration loading (`load_mcp_config`) with environment-variable substitution, strict validation (single index binding, runtime limits, field mappings), and schema inspection + safe override merging to patch incomplete `FT.INFO` output while preventing type/path identity changes.
> 
> Adds `MCPSettings` (env-backed via `pydantic-settings`), deterministic error mapping (`map_exception`), integration/unit tests for startup/teardown and guarded shutdown behavior, and a new `mcp` optional dependency group (plus import-sanity skips for `redisvl.mcp`).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit d29d3076249eed173271b791ff44b341b824c751. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/redisvl/mcp/server.py
+++ b/redisvl/mcp/server.py
@@ -1,4 +1,6 @@
 import asyncio
+from contextlib import asynccontextmanager
+from enum import Enum, auto
 from importlib import import_module
 from typing import Any, Awaitable, Optional, Type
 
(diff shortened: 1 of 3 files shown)


redisvl/mcp/config.py
```diff
@@ -0,0 +1,297 @@
...
+import yaml
+from pydantic import BaseModel, ConfigDict, Field, model_validator
+
+from redisvl.schema.fields import BaseField
+from redisvl.schema.schema import IndexSchema
+
+_ENV_PATTERN = re.compile(r"\$\{([^}:]+)(?::-([^}]*))?\}")
+
+
+class MCPRuntimeConfig(BaseModel):
```
rbs333, line 16: Is a runtime config 1:1 with an index? I'm trying to understand the taxonomy for multi-index multi vector fields potentially or am I thinking of this class wrong?

redisvl/mcp/server.py
```diff
@@ -0,0 +1,162 @@
...
+                close = getattr(vectorizer, "close", None)
+                if callable(aclose):
+                    await aclose()
+                elif callable(close):
+                    close()
+        finally:
+            if self._index is not None:
+                index = self._index
+                self._index = None
+                await index.disconnect()
```
nkanu17: 👍

redisvl/mcp/server.py
```diff
@@ -0,0 +1,162 @@
...
+
+    async def get_vectorizer(self) -> Any:
+        """Return the initialized vectorizer or fail if startup has not run."""
+        if self._vectorizer is None:
+            raise RuntimeError("MCP server has not been started")
+        return self._vectorizer
+
+    async def run_guarded(self, operation_name: str, awaitable: Awaitable[Any]) -> Any:
+        """Run a coroutine under the configured concurrency and timeout limits."""
+        del operation_name
```
nkanu17, line 106: `operation_name` may not be needed

nkanu17 (changes requested):
Apart from the one comment and @rbs333 's, the other comment I have is about `run_guarded`.

it seems the shutdown clears the index and vectorizer but run_guarded might still be allowing execution because the semaphore and config state are still available. 

I dont know the best way to solve this but I asked codex and it suggested something like using explicit lifecycle states. Worth mulling over.


What I mean is we should consider to makesure run_guarded checks if the server actually running or not instead of just checkin on the semaphore and config.

redisvl/mcp/config.py
vishal-bala: Yup! Each index has its own runtime config 👍

redisvl/mcp/server.py
vishal-bala: Yup! In this PR, the function is just a placeholder and the argument should be used by downstream code in the next PRs.

PR #574, opened 2026-04-04
Fix #480: Multi Vector > 10000 documents throws scoring error

### LINKED ISSUE #480: Multi Vector > 10000 documents throws scoring error
To reproduce:

```
from redisvl.query import MultiVectorQuery, Vector

from redisvl.schema import IndexSchema
import numpy as np
from redis import Redis

client = Redis(host="localhost", port=6379, decode_responses=True)

test_index = SearchIndex(
    IndexSchema.from_dict(
        {
            "index": {
                "name": "idx:test",
                "prefix": "test:",
                "storage_type": "json",
            },
            "fields": [
                {"name": "id", "type": "numeric"},
                {
                    "name": "embedding_1",
                    "type": "vector",
                    "attrs": {
                        "dims": 512,
                        "distance_metric": "cosine",
                        "algorithm": "hnsw",
                        "datatype": "float32",
                    },
                },
                {
                    "name": "embedding_2",
                    "type": "vector",
                    "attrs": {
                        "dims": 512,
                        "distance_metric": "cosine",
                        "algorithm": "hnsw",
                        "datatype": "float32",
                    },
                },
            ],
        },
    ),
    client,
    validate_on_load=True,
)

test_index.create(overwrite=True)

TEST_DOCS_COUNT = 10001

test_data = [
    {
        "id": i,
        "embedding_1": np.random.rand(512),
        "embedding_2": np.random.rand(512),
    }
    for i, _ in enumerate(range(TEST_DOCS_COUNT))
]

test_index.load(test_data)

embedding_1 = np.random.rand(512)
embedding_2 = np.random.rand(512)

query_vectors = [
    Vector(
        vector=embedding_1,
        field_name="embedding_1",
        dtype="float32",
        weight=0.7,
    ),
    Vector(
        vector=embedding_2,
        field_name="embedding_2",
        dtype="float32",
        weight=0.2,
    ),
]
query = MultiVectorQuery(
    vectors=query_vectors,
    num_results=10,
    return_fields=["id"],
)

results = test_index.query(query)
print(results)
```

With a TEST_DOCS_COUNT > 10,000 redis throws: 
`Error while aggregating: Could not find the value for a parameter name, consider using EXISTS if applicable for distance_1`

<=10,000 works fine.  I assume this is to do with the same logic that sets the max offset to 10,000 by default. 

The workaround I'm using for now is the unstable feature [case() to assign defaults](https://github.com/redis/redis-vl-python/pull/479) for missing scores.

Is RediSearch limited by some design constraint to this number? Seems low for many applications of vector search, or maybe I'm missing something.

JiwaniZakir:
Closes #480

When a `MultiVectorQuery` runs against an index with more than 10,000 documents, Redis does not yield a distance value for every document, causing the bare `(2 - @distance_i)/2` expression in `MultiVectorQuery.__init__` (redisvl/query/aggregate.py) to raise `"Could not find the value for a parameter name"`. The fix wraps each `apply()` call in an `if(exists(@distance_i), (2 - @distance_i)/2, 0)` guard so that documents missing a distance score default to 0 instead of erroring.

- `redisvl/query/aggregate.py` — `MultiVectorQuery.__init__`: replaced the two bare `apply()` expressions with `if(exists(...), ..., 0)` variants.
- `tests/unit/test_aggregation_types.py` — `test_multi_vector_query_string`: updated the expected query string to match the new `if(exists(...))` form.

The fix was verified by updating the unit test in `test_aggregation_types.py`, which asserts the full serialized aggregation string and now passes with the new expression.

---
*This PR was created with AI assistance (Claude). The changes were reviewed by quality gates and a critic model before submission.*

---

> [!NOTE]
> **Low Risk**
> Low risk: small query-string change that only affects how missing `@distance_i` values are handled, plus a matching unit-test update.
> 
> **Overview**
> Fixes `MultiVectorQuery` scoring failures when Redis does not return `@distance_i` for some documents by wrapping each per-vector score calculation in `if(exists(@distance_i), (2 - @distance_i)/2, 0)` so missing distances default to 0 instead of erroring.
> 
> Updates the unit test asserting the serialized aggregation query string to match the new guarded `APPLY` expressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1e69dcd237fa942ee8649104f432dbb409fc53d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/query/aggregate.py
+++ b/redisvl/query/aggregate.py
@@ -344,8 +344,13 @@ class MultiVectorQuery(AggregationQuery):
         super().__init__(query_string)
 
         # calculate the respective vector similarities
+        # use exists() guard so documents where Redis did not yield a distance
+        # (e.g. when the index contains >10 000 entries) get score 0 instead
+        # of raising "Could not find the value for a parameter name".
         for i in range(len(self._vectors)):
-            self.apply(**{f"score_{i}": f"(2 - @distance_{i})/2"})
+            self.apply(
+                **{f"score_{i}": f"if(exists(@distance_{i}), (2 - @distance_{i})/2, 0)"}
+            )
 
         # construct the scoring string based on the vector similarity scores and weights
         combined_scores = []
--- a/tests/unit/test_aggregation_types.py
+++ b/tests/unit/test_aggregation_types.py
@@ -391,7 +391,7 @@ def test_multi_vector_query_string():
 
     assert (
         str(multi_vector_query)
-        == f"@{field_1}:[VECTOR_RANGE {max_distance_1} $vector_0]=>{{$YIELD_DISTANCE_AS: distance_0}} AND @{field_2}:[VECTOR_RANGE {max_distance_2} $vector_1]=>{{$YIELD_DISTANCE_AS: distance_1}} SCORER TFIDF DIALECT 2 APPLY (2 - @distance_0)/2 AS score_0 APPLY (2 - @distance_1)/2 AS score_1 APPLY @score_0 * {weight_1} + @score_1 * {weight_2} AS combined_score SORTBY 2 @combined_score DESC MAX 10"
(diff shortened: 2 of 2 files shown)


justin-cechmanek:
Hi @JiwaniZakir. Redis Aggregations do not support an `if()` clause so this query syntax isn't correct. Our integration tests for multi-vector queries under `tests/integration/test_aggregation.py` catch this.

vishal-bala:
Hi @JiwaniZakir, since this PR has stalled for quite some time, I'd like to close it. Feel free to respond or reopen it if there's something else you'd like to add.

PR #576, opened 2026-04-08
Normalize Redis connection handling and CLI URL resolution

tylerhutcherson:
# Summary

This PR makes Redis connection behavior more consistent across the SDK and CLI. It fixes cases where different surfaces created Redis clients differently, dropped connection kwargs, or treated internally-created clients as externally owned.

On the SDK side, `from_existing()` paths for `SearchIndex`, `AsyncSearchIndex`, and `SemanticRouter` now consistently prefer a caller-provided client, preserve normalized connection kwargs, and correctly mark internally-created clients as owned so disconnect behavior works as expected. The shared kwargs splitting used by these entry points is now centralized in `redisvl/redis/connection.py` so internal _-prefixed kwargs do not leak into Redis
client construction.

This PR also routes more connection creation through `RedisConnectionFactory` instead of ad hoc client creation. That includes cache and SQL helper paths, and it fixes several extension surfaces that were passing connection kwargs incorrectly into SearchIndex.

On the CLI side, Redis URL resolution is simplified and made more predictable. The helper now cleanly resolves between explicit `--url`, explicit connection flags, `REDIS_URL`, and the local default, while also avoiding malformed URL assembly and accidental empty-auth behavior. The CLI entrypoints were simplified accordingly.

The net effect is a smaller and more uniform connection model: explicit clients are respected, factory-based connection behavior is reused across surfaces, reconnect/disconnect semantics are correct, and the CLI no longer has several connection-resolution edge cases that previously produced the wrong URL.

---

> [!NOTE]
> **Medium Risk**
> Touches core connection creation/ownership paths across `SearchIndex`, caches, routers, and CLI URL parsing; behavior changes could affect how clients are constructed and when connections are closed.
> 
> **Overview**
> **Normalizes Redis connection behavior across the SDK and CLI.** CLI commands now resolve Redis endpoints deterministically (explicit `--url` > explicit flags > `REDIS_URL` > localhost defaults), properly URL-encode auth, and only override env URLs with SSL via scheme rewriting.
> 
> **Unifies client creation and ownership semantics.** `SearchIndex.from_existing`/`AsyncSearchIndex.from_existing` and `SemanticRouter.from_existing` now prefer caller-provided clients, split/forward only connection kwargs to `RedisConnectionFactory`, track whether internally-created clients are owned (so `disconnect()` closes them), and ensure temporary clients are closed on load failures. Several extensions (`SemanticCache`, message histories, router init) now pass `connection_kwargs` explicitly (instead of splatting), and SQL query helper/client creation is routed through `RedisConnectionFactory`.
> 
> Adds unit tests covering CLI URL precedence/encoding and connection normalization (client preference, kwargs forwarding, and ownership/cleanup).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 517d281df676a432e008f9b1aa4e0aa328ee4700. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/cli/utils.py
+++ b/redisvl/cli/utils.py
@@ -1,6 +1,6 @@
 import os
 from argparse import ArgumentParser, Namespace
-from collections.abc import Sequence
+from typing import Optional
 
 from redisvl.redis.constants import REDIS_URL_ENV_VAR
 from redisvl.utils.log import get_logger
@@ -8,58 +8,22 @@ from redisvl.utils.log import get_logger
 logger = get_logger("[RedisVL]")
 DEFAULT_REDIS_HOST = "localhost"
 DEFAULT_REDIS_PORT = 6379
-_HOST_FLAGS = ("--host",)
-_PORT_FLAGS = ("-p", "--port")
-_USER_FLAGS = ("--user",)
-_PASSWORD_FLAGS = ("-a", "--password")
-_SSL_FLAGS = ("--ssl",)
-_ALL_CONNECTION_FLAGS = (
-    _HOST_FLAGS,
-    _PORT_FLAGS,
-    _USER_FLAGS,
-    _PASSWORD_FLAGS,
-    _SSL_FLAGS,
-)
-
-
-def _get_cli_argv(args: Namespace) -> tuple[str, ...]:
-    raw_argv = getattr(args, "_argv", ())
-    return tuple(raw_argv) if isinstance(raw_argv, Sequence) else ()
-
-
-def _argv_has_flag(argv: Sequence[str], *flags: str) -> bool:
-    return any(flag in argv for flag in flags)
 
 
 def _has_explicit_connection_options(args: Namespace) -> bool:
-    argv = _get_cli_argv(args)
-    if argv:
-        return any(_argv_has_flag(argv, *flags) for flags in _ALL_CONNECTION_FLAGS)
-
     return any(
         (
-            getattr(args, "host", None) not in (None, DEFAULT_REDIS_HOST),
-            getattr(args, "port", None) not in (None, DEFAULT_REDIS_PORT),
-            getattr(args, "user", None) not in (None, "", "default"),
+            getattr(args, "host", None) is not None,
+            getattr(args, "port", None) is not None,
+            getattr(args, "user", None) is not None,
             bool(getattr(args, "password", None)),
             bool(getattr(args, "ssl", False)),
         )
     )
 
 
-def _get_auth_credentials(args: Namespace) -> tuple[str | None, str | None]:
-    argv = _get_cli_argv(args)
-    if argv:
-        user = args.user if _argv_has_flag(argv, *_USER_FLAGS) else None
-        password = args.password if _argv_has_flag(argv, *_PASSWORD_FLAGS) else None
-        return user, password
-
-    user = getattr(args, "user", None)
-    if user in (None, "", "default"):
-        user = None
-
-    password = getattr(args, "password", None) or None
-    return user, password
+def _get_auth_credentials(args: Namespace) -> tuple[Optional[str], Optional[str]]:
+    return getattr(args, "user", None), getattr(args, "password", None)
 
 
 def _build_redis_url(args: Namespace) -> str:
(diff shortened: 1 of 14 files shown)


redisvl/cli/utils.py
```diff
@@ -1,29 +1,68 @@
...
+def _has_explicit_connection_options(args: Namespace) -> bool:
+    return any(
+        (
+            getattr(args, "host", None) is not None,
+            getattr(args, "port", None) is not None,
+            bool(getattr(args, "user", None)),
+            bool(getattr(args, "password", None)),
+            bool(getattr(args, "ssl", False)),
+        )
+    )
```
nkanu17, line 22: ## 🔴 [high] CLI Redis URL precedence change may break existing deployments relying on `REDIS_URL`

**Confidence:** 0.84 | **Review pass:** migration | **Priority score:** 0.85

The CLI Redis URL resolution order changed: if any connection-related flag is set (including just `--ssl`), `create_redis_url` now ignores `REDIS_URL`/`REDIS_URL_ENV_VAR` and builds a URL from defaults (`localhost:6379`). This is a **backward-incompatible precedence change** that can redirect connections away from the intended Redis instance.

### Why it matters
Many deployments set `REDIS_URL` in the environment and may also pass CLI flags like `--ssl` expecting the env var to remain authoritative. With the new logic, those invocations silently connect to localhost (or an unintended host/port).

### Failure scenario
A CI job exports `REDIS_URL=redis://prod-redis:6379` and runs a command with `--ssl` to enable TLS. Previously, the env var would be used. Now `_has_explicit_connection_options` returns `true` because `ssl` is `true`, so `create_redis_url` returns `_build_redis_url(args)` which defaults to `rediss://localhost:6379` — ignoring the env var entirely.

### Recommendation
Do not treat `--ssl` alone as an "explicit connection option" that bypasses `REDIS_URL_ENV_VAR`. Instead, apply `--ssl` as a modifier to the env URL, or only bypass env when `host`/`port`/`user`/`password` are explicitly set. Add unit tests covering:
1. env var + `--ssl` → uses env host/port
2. env var + explicit `--host`/`--port` → overrides env
3. no env var → falls back to localhost defaults

### Evidence
- `_has_explicit_connection_options` treats `bool(getattr(args, "ssl", False))` as an explicit option (line 20), so `--ssl` alone triggers bypass of env var.
- Parser defaults for `host`, `port`, `user`, `password` changed from concrete defaults to `None`, making the env-var path the intended default — but `--ssl` flips that path.
- `_build_redis_url` falls back to `DEFAULT_REDIS_HOST`/`DEFAULT_REDIS_PORT` (`localhost:6379`), producing `rediss://localhost:6379` when only `--ssl` is provided.

nkanu17 (commented):
automated review

redisvl/cli/utils.py
tylerhutcherson: @nkanu17 yeah that's a reasonable thing to fix. I can handle it

nkanu17 (commented):
security scan (confidence threshold: 0.65)

redisvl/cli/utils.py
```diff
@@ -1,29 +1,77 @@
...
+    auth = ""
+    if user:
+        auth = user
+        if password:
+            auth += f":{password}"
+        auth += "@"
+    elif password:
+        auth = f":{password}@"
+
+    return f"{scheme}://{auth}{host}:{port}"
```
nkanu17, line 49: ## Connection-string injection — credentials are not URL-encoded

**Confidence:** 0.82 | **Review pass:** injection-auth | **Priority score:** 0.74

`_build_redis_url()` constructs the Redis URL via raw string concatenation of `user`/`password`/`host`/`port` without percent-encoding. Reserved characters in credentials (`@`, `:`, `/`, `?`, `#`) will alter how the URL is parsed, potentially redirecting the connection to a different host.

### Failure scenario
A user or automation supplies a generated password containing `@`:
```
rvl index info --host myredis --password 'p@ssw0rd'
```
The constructed URL becomes `redis://:p@ssw0rd@myredis:6379`, which `urlparse` interprets as userinfo `:p` connecting to host `ssw0rd@myredis` — a completely wrong destination. More critically, a crafted password like `pw@evil.example:6379` can shift the authority so the client connects to `evil.example` instead.

### Recommendation
Percent-encode userinfo components before interpolation:
```python
from urllib.parse import quote

if user:
    auth = quote(user, safe="")
    if password:
        auth += f":{quote(password, safe='')}"
    auth += "@"
elif password:
    auth = f":{quote(password, safe='')}@"
```
Alternatively, avoid URL construction and pass `host`/`port`/`username`/`password` as separate kwargs to the Redis client. Add tests with reserved characters (`@`, `:`, `/`) in passwords.

nkanu17, line 49: ## Connection-string injection - credentials are not URL-encoded

**Confidence:** 0.82 | **Review pass:** injection-auth | **Priority score:** 0.74

`_build_redis_url()` constructs the Redis URL via raw string concatenation of `user`/`password`/`host`/`port` without percent-encoding. Reserved characters in credentials (`@`, `:`, `/`, `?`, `#`) will alter how the URL is parsed, potentially redirecting the connection to a different host.

### Failure scenario
A user or automation supplies a generated password containing `@`:
```
rvl index info --host myredis --password 'p@ssw0rd'
```
The constructed URL becomes `redis://:p@ssw0rd@myredis:6379`, which `urlparse` interprets as userinfo `:p` connecting to host `ssw0rd@myredis` — a completely wrong destination. More critically, a crafted password like `pw@evil.example:6379` can shift the authority so the client connects to `evil.example` instead.

### Recommendation
Percent-encode userinfo components before interpolation:
```python
from urllib.parse import quote

if user:
    auth = quote(user, safe="")
    if password:
        auth += f":{quote(password, safe='')}"
    auth += "@"
elif password:
    auth = f":{quote(password, safe='')}@"
```
Alternatively, avoid URL construction and pass `host`/`port`/`username`/`password` as separate kwargs to the Redis client. Add tests with reserved characters (`@`, `:`, `/`) in passwords.

nkanu17 (commented):
security scan (confidence threshold: 0.65)

nkanu17 (approved):
lgtm 👍

PR #590, opened 2026-04-24
feat(MCP): Give agents a hint about available fields for filtering and return

vishal-bala:
## Motivation

The RedisVL MCP `search-records` tool accepts both `filter` and `return_fields`, but the tool description did not tell MCP clients which fields were actually available on the configured index. In practice, that creates unnecessary trial and error for agents. They can infer that filtering is supported, but they cannot reliably discover which fields are valid without either failing requests or relying on out-of-band knowledge about the underlying schema.

That gap is especially painful because the MCP server already inspects the bound Redis index at startup and reconstructs its effective schema. The server had the information needed to guide clients, but it was not surfacing any of it in the tool metadata that agents actually read.

## Changes

This change makes the `search-records` description schema-aware. During startup, once the MCP server has loaded the effective inspected schema, it now passes that schema into search tool registration. The tool description is then built from the existing base description text plus two schema-derived hints.

The first hint advertises the fields that are currently supported by the MCP JSON filter DSL, formatted as `name(type)`. The second hint advertises every valid non-vector field name that can be requested through `return_fields`. Both lists preserve schema order so the hint stays stable and matches the configured index as closely as possible.

Custom `REDISVL_MCP_TOOL_SEARCH_DESCRIPTION` text is still supported, but it now acts as the leading description rather than replacing the server-owned guidance entirely. That means deployments can keep their own wording while still exposing accurate field hints derived from the configured index.

For an index with fields such as `content` (text), `category` (tag), `rating` (numeric), and `embedding` (vector), the `search-records` tool description now reads in the general shape below:

```text
Search records in the configured Redis index. Object filter fields: content(text), category(tag), rating(numeric). Allowed return_fields: content, category, rating.
```

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to MCP tool metadata construction and tool-registration plumbing, with no changes to query execution or data writes.
> 
> **Overview**
> Makes the MCP `search-records` tool description **schema-aware** by appending schema-derived hints for typed JSON-DSL filterable fields, `exists` support, and allowed non-vector `return_fields`.
> 
> Updates server startup to pass the inspected effective `IndexSchema` into `register_search_tool`, adjusts tests accordingly, and refreshes MCP docs/env-var wording to clarify that `REDISVL_MCP_TOOL_SEARCH_DESCRIPTION` now serves as a base description with schema hints still appended.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8300a977d481d8748f80d76a8dbd94f52b5f5c6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/concepts/mcp.md
+++ b/docs/concepts/mcp.md
@@ -91,6 +91,7 @@ These tools follow a stable contract:
 
 - request validation happens before query or write execution
 - filters support either raw strings or a RedisVL-backed JSON DSL
+- `search-records` describes the inspected schema by advertising JSON DSL filter fields and valid `return_fields`
 - error codes are mapped into a stable set of MCP-facing categories
 
 ## Why Use MCP Instead of Direct RedisVL Calls
--- a/docs/concepts/mcp.md
+++ b/docs/concepts/mcp.md
@@ -91,7 +91,7 @@ These tools follow a stable contract:
 
 - request validation happens before query or write execution
 - filters support either raw strings or a RedisVL-backed JSON DSL
-- `search-records` describes the inspected schema by advertising JSON DSL filter fields and valid `return_fields`
+- `search-records` describes the inspected schema by advertising typed JSON DSL filter fields, object-filter `exists` support, and valid `return_fields`
 - error codes are mapped into a stable set of MCP-facing categories
 
 ## Why Use MCP Instead of Direct RedisVL Calls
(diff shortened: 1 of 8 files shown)


rbs333 (approved):
LGTM

PR #592, opened 2026-04-24
Fix/cli: standardize exit codes, messages and streams

hillarytoh:
- Standardize non-MCP rvl CLI behavior: exit 0 for success/help, exit 2 for usage/argument and schema input issues, exit 1 for runtime failures; stderr for errors and stdout for normal output or help.

- Catch specific exception types in the CLI to map to the right exit code (e.g. invalid path vs Redis search)

- Version: print version to stdout (not only via logger).

---

> [!NOTE]
> **Medium Risk**
> Changes CLI exit codes/streams and error mapping, which can break existing scripts that depend on previous behavior. Logic changes are localized to CLI wrappers and do not modify core indexing/search functionality.
> 
> **Overview**
> Standardizes `rvl` CLI exit behavior and messaging: **success/help exit 0**, **usage/schema input problems exit 2**, and **runtime/Redis failures exit 1**, with errors consistently printed to *stderr*.
> 
> `index` and `stats` now explicitly catch schema-loading/validation errors (YAML/Pydantic/path issues) separately from `RedisSearchError`, and provide more actionable, index-scoped failure messages. Top-level command dispatch in `main.py` also reports unknown commands as usage errors (exit 2), and `version` now prints directly to stdout instead of logging.
> 
> Updates MCP argparse usage errors to exit 2 and adjusts the corresponding unit test expectation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7d20ffbe643ceb5e58462ba5cbe86a79b457e1e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/cli/index.py
+++ b/redisvl/cli/index.py
@@ -2,13 +2,10 @@ import argparse
 import sys
 from argparse import Namespace
 
-from redisvl.cli.utils import (
-    SCHEMA_INPUT_ERRORS,
-    add_index_parsing_options,
-    create_redis_url,
-    exit_redis_search_error,
-    exit_schema_input_error,
-)
+import yaml
+from pydantic import ValidationError
+
+from redisvl.cli.utils import add_index_parsing_options, create_redis_url
 from redisvl.exceptions import RedisSearchError
 from redisvl.index import SearchIndex
 from redisvl.redis.connection import RedisConnectionFactory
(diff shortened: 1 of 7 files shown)


redisvl/cli/index.py
```diff
@@ -33,26 +40,37 @@ def __init__(self):
...
+            sys.exit(2)
+
+        index: SearchIndex | None = None
+        try:
+            index = SearchIndex.from_yaml(args.schema, redis_url=create_redis_url(args))
+            index.create()
+        except SCHEMA_INPUT_ERRORS as e:
+            exit_schema_input_error(args, e)
+        except RedisSearchError as e:
+            exit_redis_search_error(args, index, e)
```
vishal-bala, line 101: I can see the reasoning to have `exit_schema_input_error` and `exit_redis_search_error` since those few lines get repeated often, but I think it might be better to just have those lines repeated than abstracted as a function.

The reasoning is that having the exact exit logic in-line with the try/except makes it totally clear what the behaviour is, rather than needing to peek into another file to know. Since the functions are just 3-4 lines each, the redundant code doesn't hurt _as_ bad 😅

(This is super subjective though, having the functions here is also a totally solid approach)

redisvl/cli/index.py
```diff
@@ -61,8 +79,14 @@ def info(self, args: Namespace):
...
-        index = self._connect_to_index(args)
-        _display_in_table(index.info())
+        index: SearchIndex | None = None
+        try:
+            index = self._connect_to_index(args)
+            _display_in_table(index.info())
+        except SCHEMA_INPUT_ERRORS as e:
+            exit_schema_input_error(args, e)
+        except RedisSearchError as e:
+            exit_redis_search_error(args, index, e)
```
vishal-bala: I think moving the try/except logic into the `self._connect_to_index` function might be a cleaner flow - that way you won't have to repeat the try/except multiple times, and you can catch the schema parsing errors directly on the schema parsing call (as opposed to catching on the entire connection procedure).

vishal-bala (approved):
Looks good to me! Just a couple of subjective comments about structure/flow (so feel free to take them or leave them), but otherwise I think we're all set 👍

PR #615, opened 2026-05-14
perf: use UNLINK instead of DEL in SearchIndex.drop_keys

### LINKED ISSUE #600: SearchIndex.drop_keys should use UNLINK instead of DEL
## Current behavior

`SearchIndex.drop_keys` (`redisvl/index/index.py:826`) issues a synchronous `DEL`:

```python
def drop_keys(self, keys: str | list[str]) -> int:
    if isinstance(keys, list):
    else:
        return self._redis_client.delete(keys)
```

`DEL` reclaims memory on the main thread. A single drop of tens of thousands of keys stalls the server proportionally to the freed keyspace. For a `SemanticCache` use case where scope-targeted invalidation routinely produces sweeps in the 10K–1M+ range (e.g., a policy version rollover in a multi-tenant deployment), this is a customer-visible latency event on every invalidation.

## Proposed change

Switch the bulk path to `UNLINK`, which queues memory reclamation on a background thread:

```python
return self._redis_client.unlink(*keys)
```

`UNLINK` returns the same count semantics. It is available on Redis 4+ so there is no minimum-version concern for current RedisVL targets.

If there is a reason to preserve `DEL` semantics for some callers, expose `use_unlink: bool = True` and default to `True`.

## Backwards compatibility

`UNLINK` is a strict superset of `DEL` for this use case. The only observable difference is that the keys may not be immediately reclaimed from a `MEMORY USAGE` standpoint — which is the point. Behavior in tests and at the protocol level is identical.

## Notes

Surfaced while writing a scoped semantic caching architecture spec. Also relevant to `SemanticCache.drop()`, which calls `drop_keys` under the hood when invoked with `keys=`.

LizerAIDev:
## Summary

 currently uses , which reclaims memory synchronously on the main thread. For large key deletions (10K–1M+ keys), this causes visible latency spikes.

This PR switches to , which queues memory reclamation on a background thread, returning immediately to the caller.

## Changes

- **sync ** ():  → 
- **async ** ():  → 

## Context

This is the same optimization pattern already applied to  in #613, now extended to .

Closes #600.

---

> [!NOTE]
> **Low Risk**
> Low risk behavior-preserving change that swaps `DEL` for `UNLINK` in cache/index key removal paths; main risk is environments using Redis versions without `UNLINK` support.
> 
> **Overview**
> Switches key-removal operations from `DEL`/`delete` to **Redis `UNLINK`** to make large deletions non-blocking.
> 
> This updates `EmbeddingsCache` drop helpers (`drop_by_key`, `mdrop_by_keys`, `adrop_by_key`, `amdrop_by_keys`) and `SearchIndex`/`AsyncSearchIndex` `drop_keys` to unlink keys asynchronously, with a small docstring note added in `EmbeddingsCache.drop_by_key`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ab884671d98e823f4b7bce39d88bca9aca29048. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/extensions/cache/embeddings/embeddings.py
+++ b/redisvl/extensions/cache/embeddings/embeddings.py
@@ -512,6 +512,8 @@ class EmbeddingsCache(BaseCache):
     def drop_by_key(self, key: str) -> None:
         """Remove an embedding from the cache by its Redis key.
 
+        Uses UNLINK instead of DELETE for better performance by freeing memory asynchronously.
+
         Args:
             key (str): The full Redis key for the embedding.
 
@@ -520,7 +522,7 @@ class EmbeddingsCache(BaseCache):
             cache.drop_by_key("embedcache:1234567890abcdef")
         """
         client = self._get_redis_client()
-        client.delete(key)
+        client.unlink(key)
 
     def mdrop_by_keys(self, keys: list[str]) -> None:
         """Remove multiple embeddings from the cache by their Redis keys.
@@ -542,7 +544,7 @@ class EmbeddingsCache(BaseCache):
 
         with client.pipeline(transaction=False) as pipeline:
             for key in keys:
-                pipeline.delete(key)
+                pipeline.unlink(key)
             pipeline.execute()
 
     def mdrop(self, contents: Iterable[bytes | str], model_name: str) -> None:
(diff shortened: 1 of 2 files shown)


vishal-bala:
This has been fixed in #616, but thank you for contributing.

PR #618, opened 2026-05-25
Add SemanticCache TTL refresh controls

### LINKED ISSUE #603: SemanticCache.check() unconditionally refreshes TTL on hit (sliding-window with no opt-out)
## Current behavior

`SemanticCache.check` (`redisvl/extensions/cache/llm/semantic.py:432`) refreshes the TTL of every matched entry on every hit:

```python
# Refresh TTL on all found keys
for key in redis_keys:
    self.expire(key)
```

This is a sliding-window TTL: a popular entry effectively lives as long as it keeps getting hit. There is no constructor flag, per-call argument, or per-entry attribute to disable this.

## Why this is a problem

Sliding window is the wrong default in any domain where the *popularity* of an entry correlates with the *cost* of it being stale:

- a procedure superseded by a regulatory bulletin that the cache has not yet invalidated
- a balance-patch-affected gameplay answer in a live-ops surface
- a pricing or eligibility answer that changed at midnight

In these domains the most-asked question is the most likely to be cached and the most likely to fan out a stale answer. The standard mitigation today — store an absolute `expires_at` numeric and filter on it at read time — works but requires every customer to discover the pattern independently.

## Proposed API

```python
SemanticCache(
    name="cache",
    redis_url=...,
    refresh_ttl_on_hit=True,    # current behavior, default unchanged
)

cache.check(
    prompt=...,
    refresh_ttl=None,           # per-call override
)
```

`refresh_ttl_on_hit=False` would make `check()` a pure read. Customers wanting absolute expiry would set this to `False`, store a numeric `expires_at` filterable field, and filter on it.

## Backwards compatibility

Default remains `True`. Existing behavior unchanged. Opt-in behavioral change for customers who need it.

## Notes

Surfaced while writing a scoped semantic caching architecture spec; came up repeatedly when stress-testing the design against regulated and fast-mutating domains where sliding-window TTL is actively hostile.

omribz156:
## Summary
- add `refresh_ttl_on_hit` to `SemanticCache` so the default sliding-window TTL behavior stays unchanged but can be disabled at cache construction
- add `refresh_ttl` per-call overrides for `check()` and `acheck()`
- document the opt-out in the LLM cache TTL guide and add sync/async integration coverage

Closes #603

## Verification
- `uv run black --check redisvl/extensions/cache/llm/semantic.py tests/integration/test_llmcache.py`
- `uv run python -m compileall redisvl/extensions/cache/llm/semantic.py tests/integration/test_llmcache.py`
- `python -c "import json, pathlib; json.loads(pathlib.Path('docs/user_guide/03_llmcache.ipynb').read_text(encoding='utf-8')); print('notebook json ok')"`
- `git diff --check`

I also attempted `uv run pytest tests/integration/test_llmcache.py -k "ttl_refresh or ttl_refresh_setting"`, but this local machine does not have Docker on PATH, so testcontainers failed before running the selected tests with `FileNotFoundError: [WinError 2]` while invoking `docker compose`.

This was implemented with Codex assistance and manually reviewed before submission.

---

> [!NOTE]
> **Low Risk**
> Backward-compatible API change with default TTL refresh unchanged; only adds opt-out knobs and tests around existing cache hit behavior.
> 
> **Overview**
> Adds configurable control over whether cache lookups extend entry lifetime in `SemanticCache`, while keeping the existing sliding-window TTL behavior as the default.
> 
> **`refresh_ttl_on_hit`** (default `True`) on `SemanticCache` disables TTL extension on every hit when set to `False`. **`check()` / `acheck()`** accept optional **`refresh_ttl`** to override that setting per lookup (e.g. `refresh_ttl=False` for read-only probes). The base LLM cache interface and `LangCacheSemanticCache` accept the same parameter for API consistency; LangCache logs a warning because it does not implement TTL refresh.
> 
> The LLM cache user guide documents the new options and examples, and integration tests cover sync/async skip, cache-level off, and per-call override via mocked `expire` / `aexpire`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d333967187eda0f5a6964d5ab2faacd95989e31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/extensions/cache/llm/semantic.py
+++ b/redisvl/extensions/cache/llm/semantic.py
@@ -54,6 +54,7 @@ class SemanticCache(BaseLLMCache):
         redis_url: str = "redis://localhost:6379",
         connection_kwargs: dict[str, Any] = {},
         overwrite: bool = False,
+        refresh_ttl_on_hit: bool = True,
         **kwargs,
     ):
         """Semantic Cache for Large Language Models.
@@ -77,6 +78,8 @@ class SemanticCache(BaseLLMCache):
                 for the redis client. Defaults to empty {}.
             overwrite (bool): Whether or not to force overwrite the schema for
                 the semantic cache index. Defaults to false.
+            refresh_ttl_on_hit (bool): Whether cache hits refresh matched
+                entries' TTL by default. Defaults to True.
 
         Raises:
             TypeError: If an invalid vectorizer is provided.
@@ -92,6 +95,8 @@ class SemanticCache(BaseLLMCache):
             redis_url=redis_url,
             connection_kwargs=connection_kwargs,
         )
+
+        self.refresh_ttl_on_hit = refresh_ttl_on_hit
 
         # Handle the deprecated dtype parameter
         dtype = kwargs.pop("dtype", None)
@@ -353,6 +358,7 @@ class SemanticCache(BaseLLMCache):
         return_fields: list[str] | None = None,
         filter_expression: FilterExpression | None = None,
         distance_threshold: float | None = None,
+        refresh_ttl: bool | None = None,
     ) -> list[dict[str, Any]]:
         """Checks the semantic cache for results similar to the specified prompt
         or vector.
@@ -377,6 +383,9 @@ class SemanticCache(BaseLLMCache):
                 the full cache will be searched.
             distance_threshold (Optional[float]): The threshold for semantic
                 vector distance.
+            refresh_ttl (Optional[bool]): Whether this cache hit should refresh
+                matched entries' TTL. If None, uses the cache-level
+                refresh_ttl_on_hit setting.
 
         Returns:
             List[Dict[str, Any]]: A list of dicts containing the requested
@@ -430,9 +439,10 @@ class SemanticCache(BaseLLMCache):
             return_fields,  # type: ignore
         )
 
-        # Refresh TTL on all found keys
-        for key in redis_keys:
-            self.expire(key)
+        refresh_ttl = self.refresh_ttl_on_hit if refresh_ttl is None else refresh_ttl
+        if refresh_ttl:
+            for key in redis_keys:
+                self.expire(key)
 
         return cache_hits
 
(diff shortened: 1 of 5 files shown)


tests/integration/test_llmcache.py
```diff
@@ -409,6 +467,39 @@ async def test_async_ttl_refresh(cache_with_ttl, vectorizer):
...
+    cache_with_ttl.refresh_ttl_on_hit = False
+
+    async with cache_with_ttl:
+        await cache_with_ttl.astore(prompt, response, vector=vector)
+
+        with patch.object(cache_with_ttl, "aexpire", new_callable=AsyncMock) as aexpire:
+            check_result = await cache_with_ttl.acheck(vector=vector, refresh_ttl=True)
+
+    assert len(check_result) == 1
+    aexpire.assert_called_once()
```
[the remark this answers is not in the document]
omribz156: Thanks, fixed in 3dfafaa. The async TTL-refresh assertions now use assert_not_awaited() and assert_awaited_once(). Verified with black --check, compileall, and git diff --check.

redisvl/extensions/cache/llm/semantic.py
```diff
@@ -353,6 +358,7 @@ def check(
         return_fields: list[str] | None = None,
         filter_expression: FilterExpression | None = None,
         distance_threshold: float | None = None,
+        refresh_ttl: bool | None = None,
```
vishal-bala, line 361: Since this class inherits from `BaseLLMCache`, you'll want to make the corresponding update on there as well. Otherwise type-checkers will complain

vishal-bala (commented):
Looks good aside from one minor comment! Should be good to go once we have that addressed.

redisvl/extensions/cache/llm/semantic.py
omribz156: Thanks, fixed in 362ca90. I added the matching refresh_ttl parameter to BaseLLMCache.check() and BaseLLMCache.acheck(), plus the sync docstring note, so SemanticCache stays compatible with the base interface. Verified with python -m compileall on base.py/semantic.py and git diff --check.

vishal-bala:
Looks like your code hasn't run the formatter before committing - there's a `make` command to get those checks done automatically, should fix the failing lint checks 👍

vishal-bala:
This is coming along, please resolve the merge conflicts with `main` and we should be nearly set 👍

PR #624, opened 2026-06-09
make add_route public

### LINKED ISSUE #623: SemanticRouter: public `add_route` method (+ two persistence bugs)
# SemanticRouter: public `add_route` method (+ two persistence bugs)

## Summary

`SemanticRouter` cannot add a brand-new route to a running instance through
its public API. The primitive exists privately as `_add_routes()` and is
already used by the test suite, but invoking it post-init also surfaces a
latent bug: the persisted router config in Redis (`<name>:route_config`)
drifts away from the in-memory `self.routes`, so the new route disappears on
the next `SemanticRouter.from_existing(name)`. `remove_route()` has the same
drift.

This issue proposes a public `add_route(route)` method and a single
companion fix to `remove_route()` that closes both drift cases. References
below are against `redisvl/extensions/router/semantic.py` at HEAD.

## Motivation

Today the only supported runtime mutations are:

- `add_route_references(route_name, references)` — extends an **existing**
  route. Raises `ValueError` if the route does not exist (line 747).
- `remove_route(route_name)` — drops a route's references and removes it
  from `self.routes` (line 533).

There is no supported way to add a brand-new route to an already-initialized
router. Applications that want to grow their route set at runtime (e.g.
admin-managed semantic routers, agentic tool registration, user-driven
intent capture) have to either:

- recreate the router with the full new route list and `overwrite=True`,
  re-embedding every reference and incurring downtime, or
- reach into the private `_add_routes()` API.

## Existing private API

`_add_routes(routes: list[Route])` (line 281) already performs the embed +
write path:

1. `self.vectorizer.embed_many(...)` over each route's references.
2. Builds `(reference_id, route_name, reference, vector)` payloads and keys
   via `_route_ref_key`.
3. `self._index.load(...)` writes references to Redis.
4. Appends to `self.routes` only when `self.get(route.name)` is `None`
   (line 311–312).

The integration test `test_add_route` (`tests/integration/test_semantic_router.py:146`)
already exercises it directly:

```python
semantic_router._add_routes(new_routes)
```

## Bug 1 — `_add_routes` does not persist `route_config`

`__init__` writes the router config to Redis once (line 116):

```python
self._index.client.json().set(f"{self.name}:route_config", ".", self.to_dict())
```

`add_route_references()` re-writes it at the tail via
`self._update_router_state()` (line 749). `_add_routes()` does neither. As a
result, a route added post-init is queryable in-process but invisible to a
subsequent `SemanticRouter.from_existing(name)`, which reconstructs the
router from the `<name>:route_config` JSON (line 154).

The existing `test_add_route` does not catch this because it never round-trips
through `from_existing` after mutating.

## Bug 2 — `remove_route` does not persist `route_config`

`remove_route()` (line 533) drops the reference keys and updates
`self.routes` in memory, but never calls `_update_router_state()`. The
removed route reappears on the next `from_existing`. Same drift, same
test-coverage gap.

## Bug 3 — `_add_routes` silently no-ops on name collision

Line 311–312:

```python
if not self.get(route.name):
    self.routes.append(route)
```

When a caller passes a route whose name already exists, the references are
still re-embedded and re-written to the index, but `self.routes` is not
updated to reflect them — and no error or warning is emitted. The right tool
for that case is `add_route_references()`; the collision branch should
either raise or be unreachable from the new public method.

### Public method

```python
def add_route(self, route: Route) -> str:
    """Add a new route to the SemanticRouter.

    Embeds the route's references, writes them to the Redis index,
    appends the route to self.routes, and persists the updated router
    config so the route survives SemanticRouter.from_existing(name).

    Args:
        route: A fully-formed Route (name, references, distance_threshold,
            optional metadata).

    Returns:
        The added route's name.

    Raises:
        ValueError: If a route with this name already exists on the router.
            Use add_route_references(route_name, references) to extend an
            existing route.
    """
    if self.get(route.name) is not None:
        raise ValueError(
            f"Route {route.name!r} already exists on router {self.name!r}; "
            f"use add_route_references() to add references to it"
        )
    self._add_routes([route])
    self._update_router_state()
    return route.name
```

### Companion fix to `remove_route`

Append a single `self._update_router_state()` call at the end of
`remove_route()` when the route existed (i.e. inside the `else` branch where
`self.routes` is already mutated, line 549). The caller already expects the
route to be gone — this just makes that durable across `from_existing`.

## Tests

Add to `tests/integration/test_semantic_router.py`:

- `test_add_route_public` — happy path: `router.add_route(Route(...))`,
  assert `router.get(name)` returns it, assert it matches a query of one of
  its references.
- `test_add_route_survives_from_existing` — add a route, then
  `SemanticRouter.from_existing(name)`, assert the new route is present
  (regression for Bug 1).
- `test_add_route_duplicate_raises` — `router.add_route` with an existing
  name raises `ValueError` (Bug 3).
- `test_remove_route_survives_from_existing` — remove a route, reload via
  `from_existing`, assert the route is gone (regression for Bug 2).

The existing `test_add_route` (which exercises `_add_routes` directly)
should either be renamed `test__add_routes_private` to flag its intent or be
removed in favor of the public-API tests above.

## Backwards compatibility

- `add_route` is additive. No existing caller breaks.
- Persisting `route_config` in `remove_route` is a strict bug fix; any
  caller depending on the non-durable removal is already broken across
  reboots.
- `_add_routes` remains for internal use (`__init__` and the new
  `add_route`). Callers that have been reaching into the private API
  directly can migrate to `add_route` and get the persistence behavior for
  free.

rbs333:
This resolves #623

### DIFF
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -530,6 +530,33 @@ class SemanticRouter(BaseModel):
 
         return top_route_matches
 
+    def add_route(self, route: Route) -> str:
+        """Add a new route to the SemanticRouter.
+
+        Embeds the route's references, writes them to the Redis index,
+        appends the route to ``self.routes``, and persists the updated router
+        config so the route survives :meth:`SemanticRouter.from_existing`.
+
+        Args:
+            route (Route): A fully-formed Route (name, references,
+                distance_threshold, optional metadata).
+
+        Returns:
+            str: The added route's name.
+
+        Raises:
+            ValueError: If a route with this name already exists on the router.
+                Use :meth:`add_route_references` to extend an existing route.
+        """
+        if self.get(route.name) is not None:
+            raise ValueError(
+                f"Route {route.name!r} already exists on router {self.name!r}; "
+                f"use add_route_references() to add references to it"
+            )
+        self._add_routes([route])
+        self._update_router_state()
+        return route.name
+
     def remove_route(self, route_name: str) -> None:
         """Remove a route and all references from the semantic router.
 
@@ -547,6 +574,7 @@ class SemanticRouter(BaseModel):
                 ]
             )
             self.routes = [route for route in self.routes if route.name != route_name]
+            self._update_router_state()
 
     def delete(self) -> None:
         """Delete the semantic router index."""
--- a/tests/integration/test_semantic_router.py
+++ b/tests/integration/test_semantic_router.py
@@ -170,12 +170,127 @@ def test_add_route(semantic_router):
         assert match.name == "politics"
 
 
+def test_add_route_public(semantic_router):
+    new_route = Route(
+        name="politics",
+        references=[
+            "are you liberal or conservative?",
+            "who will you vote for?",
+            "political speech",
+        ],
+        metadata={"type": "politics"},
+        distance_threshold=0.3,
+    )
+    added_name = semantic_router.add_route(new_route)
+    assert added_name == "politics"
+
+    route = semantic_router.get("politics")
+    assert route is not None
+    assert route.name == "politics"
+    assert "political speech" in route.references
+
+    redis_version = semantic_router._index.client.info()["redis_version"]
+    if is_version_gte(redis_version, "7.0.0"):
+        match = semantic_router("political speech")
+        assert match is not None
+        assert match.name == "politics"
+
+
+def test_add_route_survives_from_existing(
+    client, redis_url, routes, redis_test_name, hf_vectorizer
+):
+    skip_if_no_redis_search(client)
+    skip_if_redis_version_below(client, "7.0.0")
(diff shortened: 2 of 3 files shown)


redisvl/extensions/router/semantic.py
```diff
@@ -530,6 +530,33 @@ def route_many(
...
+                Use :meth:`add_route_references` to extend an existing route.
+        """
+        if self.get(route.name) is not None:
+            raise ValueError(
+                f"Route {route.name!r} already exists on router {self.name!r}; "
+                f"use add_route_references() to add references to it"
+            )
+        self._add_routes([route])
+        self._update_router_state()
+        return route.name
```
nkanu17, line 558: ✅

nkanu17 (approved):

PR #629, opened 2026-06-16
feat(mcp): support multiple active index bindings (RAAE-1604)

vishal-bala:
## Motivation

The RedisVL MCP server currently binds to exactly one Redis index per process. That single-binding assumption is enforced by a config validator and baked throughout the codebase — single-resource server state, single-binding convenience accessors on `MCPConfig`, and the search/upsert tools. Before the server can expose multiple logical indexes from a single endpoint ([RAAE-1603](https://redislabs.atlassian.net/browse/RAAE-1603)), that assumption has to be removed and replaced with a real multi-binding model.

This PR ([RAAE-1604](https://redislabs.atlassian.net/browse/RAAE-1604)) does exactly that, and nothing more: it reshapes the configuration and runtime model so the server can start, inspect, validate, and serve one *or many* bindings, while keeping existing single-index configs and callers behaving identically. It is the foundation the rest of the epic (discovery via `list-indexes`, index routing on `search-records`/`upsert-records`, docs) builds on, so it intentionally does not yet add any new request parameters or tools.

## Implementation

The core of the change is a new immutable `BindingRuntime` (in `redisvl/mcp/runtime.py`) that bundles everything a tool call needs for one logical index: the binding config, the connected `AsyncSearchIndex`, its effective (inspected + overridden) schema, an optional vectorizer, the resolved native-hybrid-search capability, and the effective read-only flag. The server now holds a `dict[str, BindingRuntime]` keyed by logical id instead of a single set of `_index`/`_vectorizer` fields. Startup iterates every configured binding and inspects, validates, and initializes each one independently — each binding owns its own Redis client — with all-or-nothing teardown so a single bad binding fails startup cleanly without leaking connections.

On the config side, the "exactly one configured index binding" validator is gone (we now simply require at least one binding with non-blank ids), and the schema-inspection, runtime-mapping, and search-validation methods move from `MCPConfig` onto `MCPIndexBindingConfig` where they naturally belong per binding. The single-binding convenience accessors on `MCPConfig` are removed. Each binding gains optional `description` and `read_only` fields, and a binding's effective write availability is computed as global `--read-only` OR the per-index `read_only`. Tool resolution goes through a new `server.resolve_binding(index_id)` helper that defaults to the sole binding when one is configured (preserving backward compatibility) and returns an `invalid_request` error when an index is omitted with multiple bindings configured or when an unknown id is given. The search and upsert tools were re-threaded to operate on a resolved `BindingRuntime` rather than reaching into single-binding server accessors.

Additional notes:

- Native-hybrid-search support is now probed eagerly per binding at startup and stored on the `BindingRuntime`, replacing the previous lazy single-index cache.
- The concurrency semaphore is a single process-wide ceiling sized from the maximum `max_concurrency` across bindings; the request timeout is sourced per-binding and passed explicitly into `run_guarded`.
- `get_index()` / `get_vectorizer()` are retained as thin convenience wrappers over `resolve_binding(None)`.
- Implemented test-first: new coverage for multi-binding config loading, `description`/`read_only` defaults, `resolve_binding` routing semantics, semaphore sizing, per-binding teardown, and three integration tests (multi-binding startup, global read-only override, and a single invalid binding failing startup), alongside the updated single-index tests that confirm backward compatibility.

## Verification

- `mypy` clean across all source files; `black`/`isort` formatted.
- 182 MCP unit tests pass.
- 44 MCP integration tests pass (2 skipped on Redis-version gates) against Redis 8.

[RAAE-1603]: https://redislabs.atlassian.net/browse/RAAE-1603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RAAE-1604]: https://redislabs.atlassian.net/browse/RAAE-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Medium Risk**
> MCP startup/teardown and binding resolution affect how indexes are served; the required `run_guarded(..., timeout_seconds=)` change breaks custom MCP extensions that call it directly.
> 
> **Overview**
> **MCP** moves from a single enforced index binding to a **`dict` of `BindingRuntime`** entries: startup inspects and initializes each configured index independently (own client, vectorizer, hybrid probe, effective read-only), with **`resolve_binding(index_id)`** defaulting when only one index is configured and rejecting ambiguous or unknown ids. Config drops the “exactly one binding” rule and **`MCPConfig`** convenience accessors; per-binding **`description`**, **`read_only`**, and schema/search helpers live on **`MCPIndexBindingConfig`**. Search/upsert tools read from the resolved runtime; **`run_guarded`** now requires **`timeout_seconds=`** per binding (breaking for direct callers).
> 
> Also in this release: **`SearchIndex.drop_keys`** uses **`UNLINK`** instead of **`DEL`**; semantic router **`delete()`** removes the standalone route-config key; **`sql-redis>=0.7.1`** with docs for **`hybrid_vector_search` / FT.HYBRID**; auto-release publishes to PyPI via **`pypa/gh-action-pypi-publish`** with OIDC; version **0.22.0**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd2a28a844570746ab3ba9e20339e34836517e2c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

---

### ⚠️ Breaking change for downstream authors

`RedisVLMCPServer.run_guarded(operation_name, awaitable)` now requires a keyword-only `timeout_seconds` argument (sourced from each binding's `request_timeout_seconds`). Any code that subclasses `RedisVLMCPServer` or calls `run_guarded` directly (custom tools/plugins) must update its call sites to pass `timeout_seconds=`, otherwise it raises `TypeError: run_guarded() missing 1 required keyword-only argument: 'timeout_seconds'` at call time. This repository has no CHANGELOG file, so this note serves as the migration callout (per review feedback on #629).

### DIFF
--- a/redisvl/mcp/server.py
+++ b/redisvl/mcp/server.py
@@ -10,7 +10,9 @@ from redis import __version__ as redis_py_version
 from redisvl.exceptions import RedisSearchError
 from redisvl.index import AsyncSearchIndex
 from redisvl.mcp.auth import build_auth_provider, resolve_auth_config
-from redisvl.mcp.config import MCPConfig, load_mcp_config
+from redisvl.mcp.config import MCPConfig, MCPIndexBindingConfig, load_mcp_config
+from redisvl.mcp.errors import MCPErrorCode, RedisVLMCPError
+from redisvl.mcp.runtime import BindingRuntime
 from redisvl.mcp.settings import MCPSettings
 from redisvl.mcp.tools.search import register_search_tool
 from redisvl.mcp.tools.upsert import register_upsert_tool
@@ -47,17 +49,15 @@ class _LifecycleState(Enum):
 
 
 class RedisVLMCPServer(FastMCP):
-    """MCP server exposing RedisVL capabilities for one existing Redis index."""
+    """MCP server exposing RedisVL capabilities for one or many existing indexes."""
 
     _LifecycleState = _LifecycleState
 
     def __init__(self, settings: MCPSettings):
-        """Create a server shell with lazy config, index, and vectorizer state."""
+        """Create a server shell with lazy config and per-binding runtime state."""
         self.mcp_settings = settings
         self.config: MCPConfig | None = None
-        self._index: AsyncSearchIndex | None = None
-        self._vectorizer: Any | None = None
-        self._supports_native_hybrid_search: bool | None = None
+        self._bindings: dict[str, BindingRuntime] = {}
         self._semaphore: asyncio.Semaphore | None = None
         self._tools_registered = False
 
@@ -90,12 +90,10 @@ class RedisVLMCPServer(FastMCP):
         """Load config, inspect the configured index, and initialize dependencies."""
         async with self._transition_lock:
             await self._begin_startup()
-            client = None
             try:
-                client = await self._initialize_runtime_resources()
+                await self._initialize_runtime_resources()
                 await self._mark_running()
             except Exception:
-                await self._teardown_runtime(client)
                 await self._mark_stopped()
                 raise
 
@@ -112,22 +110,64 @@ class RedisVLMCPServer(FastMCP):
             finally:
                 await self._mark_stopped()
 
-    async def get_index(self) -> AsyncSearchIndex:
-        """Return the initialized async index or fail if startup has not run."""
-        if self._index is None:
-            raise RuntimeError("MCP server has not been started")
-        return self._index
+    def resolve_binding(self, index_id: str | None) -> BindingRuntime:
+        """Resolve the runtime for a logical index id, honoring single-index defaults.
 
-    async def get_vectorizer(self) -> Any:
-        """Return the initialized vectorizer or fail if startup has not run."""
-        if self.config is None:
+        - ``None`` with exactly one configured binding returns that binding,
+          preserving backward-compatible single-index behavior.
+        - ``None`` with multiple bindings is an ``invalid_request``; the caller
+          must name an index.
+        - An unknown id is an ``invalid_request``.
+
+        Write-availability is not enforced here; that is the upsert tool's job.
+        """
+        if not self._bindings:
             raise RuntimeError("MCP server has not been started")
-        if self._vectorizer is None:
+
+        if index_id is None:
+            if len(self._bindings) == 1:
+                return next(iter(self._bindings.values()))
+            available = ", ".join(sorted(self._bindings))
+            raise RedisVLMCPError(
+                "index is required when multiple indexes are configured; "
+                f"available: {available}",
+                code=MCPErrorCode.INVALID_REQUEST,
+                retryable=False,
+            )
+
+        runtime = self._bindings.get(index_id)
+        if runtime is None:
+            available = ", ".join(sorted(self._bindings))
+            raise RedisVLMCPError(
+                f"Unknown index '{index_id}'; available: {available}",
+                code=MCPErrorCode.INVALID_REQUEST,
+                retryable=False,
+            )
+        return runtime
+
+    async def get_index(self, index_id: str | None = None) -> AsyncSearchIndex:
+        """Return an initialized async index, defaulting to the sole binding."""
+        return self.resolve_binding(index_id).index
+
+    async def get_vectorizer(self, index_id: str | None = None) -> Any:
+        """Return an initialized vectorizer, defaulting to the sole binding."""
+        runtime = self.resolve_binding(index_id)
+        if runtime.vectorizer is None:
             raise RuntimeError("MCP server vectorizer is not configured")
-        return self._vectorizer
+        return runtime.vectorizer
 
-    async def run_guarded(self, operation_name: str, awaitable: Awaitable[Any]) -> Any:
-        """Run a coroutine under the configured concurrency and timeout limits."""
+    async def run_guarded(
+        self,
+        operation_name: str,
+        awaitable: Awaitable[Any],
+        *,
+        timeout_seconds: float,
+    ) -> Any:
+        """Run a coroutine under the global concurrency cap and a request timeout.
+
+        The timeout is sourced per-binding by the caller; the concurrency
+        semaphore is a single process-wide ceiling shared across all bindings.
+        """
         del operation_name
         semaphore = self._semaphore
         if semaphore is None:
@@ -140,8 +180,7 @@ class RedisVLMCPServer(FastMCP):
                     self._close_awaitable(awaitable)
                     raise RuntimeError("MCP server is not running")
 
-                config = self.config
-                if config is None:
+                if self.config is None:
                     self._close_awaitable(awaitable)
                     raise RuntimeError("MCP server is not running")
 
@@ -149,33 +188,32 @@ class RedisVLMCPServer(FastMCP):
                 self._active_requests_drained.clear()
 
             try:
-                return await asyncio.wait_for(
-                    awaitable,
-                    timeout=config.runtime.request_timeout_seconds,
-                )
+                return await asyncio.wait_for(awaitable, timeout=timeout_seconds)
             finally:
                 async with self._request_state_lock:
                     self._active_requests -= 1
                     if self._active_requests == 0:
                         self._active_requests_drained.set()
 
-    def _build_vectorizer(self) -> Any:
-        """Instantiate the configured vectorizer class from validated config."""
-        if self.config is None:
-            raise RuntimeError("MCP server config not loaded")
-        if self.config.vectorizer is None:
+    @staticmethod
+    def _build_vectorizer(binding: MCPIndexBindingConfig) -> Any:
+        """Instantiate a binding's configured vectorizer class from its config."""
+        if binding.vectorizer is None:
             raise RuntimeError("MCP server vectorizer is not configured")
 
-        vectorizer_class = resolve_vectorizer_class(self.config.vectorizer.class_name)
-        return vectorizer_class(**self.config.vectorizer.to_init_kwargs())
+        vectorizer_class = resolve_vectorizer_class(binding.vectorizer.class_name)
+        return vectorizer_class(**binding.vectorizer.to_init_kwargs())
 
-    def _validate_vectorizer_dims(self, schema: IndexSchema) -> None:
+    @staticmethod
+    def _validate_vectorizer_dims(
+        binding: MCPIndexBindingConfig, vectorizer: Any, schema: IndexSchema
+    ) -> None:
         """Fail startup when vectorizer dimensions disagree with schema dimensions."""
-        if self.config is None or self._vectorizer is None:
+        if vectorizer is None:
             return
 
-        configured_dims = self.config.get_vector_field_dims(schema)
-        actual_dims = getattr(self._vectorizer, "dims", None)
+        configured_dims = binding.get_vector_field_dims(schema)
+        actual_dims = getattr(vectorizer, "dims", None)
         if (
             configured_dims is not None
             and actual_dims is not None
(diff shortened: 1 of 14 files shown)


vishal-bala:
CI is failing because of flaky tests that will be fixed by https://github.com/redis-developer/sql-redis/pull/39

redisvl/mcp/server.py
```diff
@@ -259,12 +296,23 @@ async def _teardown_runtime(self, client: Any | None = None) -> None:
...
                 await client.aclose()
 
+    async def _teardown_runtime(self) -> None:
+        """Release every binding's runtime resources and clear terminal state."""
+        bindings = list(self._bindings.values())
+        self._bindings = {}
+        self.config = None
+        self._semaphore = None
+
+        for runtime in bindings:
```
nkanu17, line 316: Teardown loop leaks connections on partial failure

`_close_resources` has an internal `try/finally` that ensures `index.disconnect()` runs even if `vectorizer.aclose()` throws. But `_close_resources` itself can still raise (e.g. if `index.disconnect()` throws a network error).

If that happens on binding 1, the exception propagates out of this loop immediately. Bindings 2 through N never reach `_close_resources`. Their Redis connections are never closed, causing a connection leak on every failed teardown. This matters most in the partial-startup-failure path where `_teardown_runtime` is called from the `except` block in `_initialize_runtime_resources`.

Suggested fix:
```python
for runtime in bindings:
    try:
        await self._close_resources(
            index=runtime.index, vectorizer=runtime.vectorizer
        )
    except Exception:
        logger.warning(
            "error closing binding %s during teardown",
            runtime.binding_id,
            exc_info=True,
        )
```
Best-effort teardown: log the failure, keep going, clean up everything possible.

redisvl/mcp/server.py
```diff
@@ -259,12 +296,23 @@ async def _teardown_runtime(self, client: Any | None = None) -> None:
...
                 await index.disconnect()
             elif client is not None:
                 await client.aclose()
 
+    async def _teardown_runtime(self) -> None:
+        """Release every binding's runtime resources and clear terminal state."""
+        bindings = list(self._bindings.values())
+        self._bindings = {}
+        self.config = None
+        self._semaphore = None
```
nkanu17, line 314: `_tools_registered` is never reset in teardown

`_teardown_runtime` clears `_bindings`, `config`, and `_semaphore` but leaves `_tools_registered = True`. If startup ever fails after `_register_tools()` succeeds (unlikely with the current call order but possible if it shifts), a subsequent `startup()` call will hit the guard at the top of `_register_tools` and skip registration entirely, leaving the server with no tools.

Add `self._tools_registered = False` here alongside the other state resets.

redisvl/mcp/server.py
```diff
@@ -112,22 +110,64 @@ async def shutdown(self) -> None:
...
+        if runtime is None:
+            available = ", ".join(sorted(self._bindings))
+            raise RedisVLMCPError(
+                f"Unknown index '{index_id}'; available: {available}",
+                code=MCPErrorCode.INVALID_REQUEST,
+                retryable=False,
+            )
+        return runtime
+
+    async def get_index(self, index_id: str | None = None) -> AsyncSearchIndex:
```
nkanu17: nit: `get_index()` and `get_vectorizer()` are footguns on multi-binding servers

Both methods call `resolve_binding(None)` which throws `INVALID_REQUEST` on any multi-binding server. A future tool author who finds these convenience methods and uses them:

```python
index = await server.get_index()  # breaks silently on multi-binding
```

gets a runtime error only when deployed with a multi-binding config. No TypeError at import time, no static analysis warning unless mypy is run against the caller.

Consider adding a deprecation warning or removing them entirely and directing all tool authors to use `resolve_binding(index_id)` directly.

redisvl/mcp/server.py
```diff
@@ -112,22 +110,64 @@ async def shutdown(self) -> None:
...
+        return self.resolve_binding(index_id).index
+
+    async def get_vectorizer(self, index_id: str | None = None) -> Any:
+        """Return an initialized vectorizer, defaulting to the sole binding."""
+        runtime = self.resolve_binding(index_id)
+        if runtime.vectorizer is None:
+            raise RuntimeError("MCP server vectorizer is not configured")
+        return runtime.vectorizer
+
+    async def run_guarded(
```
nkanu17, line 157: nit: breaking api change

Adding `timeout_seconds` as a required keyword-only argument is the right Python pattern. But this is a breaking change to a previously 2-argument public method. Any code subclassing `RedisVLMCPServer` or calling `run_guarded` directly (custom tools, plugins) will get:

```
TypeError: run_guarded() missing 1 required keyword-only argument: 'timeout_seconds'
```

This only surfaces at runtime when the codepath is hit, not at import time. Please add a CHANGELOG entry and a note in the PR description so downstream authors know to update their call sites.

nkanu17 (changes requested):
I really like the design!

Binding runtime dataclass is nice and like how resolve binding works as the single routing entry point shared across all tools keeps the behavior consistent and how it's backward compatible for single binding servers works

One blocking issue before merge: 

the teardown loop in `_teardown_runtime` does not protect subsequent bindings if `_close_resources` raises on an earlier one. 
- A failed disconnect on binding 1 will propagate out of the loop immediately, leaving bindings 2 through N with open Redis connections that are never closed. 
- In a partial startup failure this is exactly the path that gets exercised, so the leak is not hypothetical. 
- Wrapping each iteration in a try/except that logs and continues is a small fix and should be done before this lands.

redisvl/mcp/server.py
```diff
@@ -329,29 +378,63 @@ def _verify_auth_not_stale(self) -> None:
...
+    async def _initialize_runtime_resources(self) -> None:
+        """Load config and initialize every configured binding independently."""
         self.config = load_mcp_config(self._config_path)
         self._verify_auth_not_stale()
-        self._semaphore = asyncio.Semaphore(self.config.runtime.max_concurrency)
-        self._supports_native_hybrid_search = None
-        timeout = self.config.runtime.startup_timeout_seconds
+        # The semaphore is a single process-wide concurrency ceiling shared by
+        # all bindings; take the max configured limit across bindings.
+        self._semaphore = asyncio.Semaphore(
```
nkanu17, line 401: nit: Semaphore semantics need a clarifying comment
```suggestion
        # max_concurrency is configured per-binding but applied here as a single
        # process-wide ceiling shared across all bindings. Taking max() means the
        # most permissive binding sets the global cap: 5 bindings each with
        # max_concurrency=2 yields Semaphore(2), not Semaphope(10).
        # Consider renaming to global_max_concurrency in a future config version
        # to avoid misleading operators who expect per-index slot allocation.
        self._semaphore = asyncio.Semaphore(
```

redisvl/mcp/tools/upsert.py
```diff
@@ -319,14 +320,14 @@ async def upsert_records(
...
-            _prepare_record_for_storage(record, server=server, index=index)
-            for record in prepared_records
+            _prepare_record_for_storage(record, rt=rt) for record in prepared_records
         ]
 
         try:
             keys = await server.run_guarded(
                 "upsert-records",
                 index.load(loadable_records, id_field=id_field),
+                timeout_seconds=runtime.request_timeout_seconds,
```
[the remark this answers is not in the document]
vishal-bala: Added an `rt.effective_read_only` guard at the top of `upsert_records()` that raises `FORBIDDEN` before any processing — covers the case where the global flag is off but a specific binding has `read_only: true` (commit 3a0d1fe).

redisvl/mcp/server.py
vishal-bala: Wrapped each `_close_resources()` call in `try/except` — a failed disconnect on one binding now logs a warning and continues to the next rather than propagating out of the loop (commit 3a0d1fe).

redisvl/mcp/server.py
vishal-bala: Added `self._tools_registered = False` to `_teardown_runtime()` alongside the other state resets (commit 3a0d1fe).

redisvl/mcp/server.py
vishal-bala: Added the clarifying comment explaining the "most permissive binding sets the cap" semantics (commit 3a0d1fe).

redisvl/mcp/server.py
vishal-bala: Fixed in 3a0d1fe: each binding's close is wrapped in try/except that logs and continues, so a failure on one binding no longer leaks the others' connections. Added a regression test in b6b2db1 (`test_teardown_continues_when_a_binding_fails_to_close`).

redisvl/mcp/server.py
vishal-bala: Fixed in 3a0d1fe: `_teardown_runtime` now resets `self._tools_registered = False` alongside the other state resets; covered by the teardown regression test in b6b2db1.

redisvl/mcp/server.py
vishal-bala: Done in 3a0d1fe: the comment now spells out the process-wide `max()` semantics (e.g. five bindings each with max_concurrency=2 yield Semaphore(2), not 10).

redisvl/mcp/server.py
vishal-bala: Removed both methods entirely in 2028ecd (they had no non-test callers); all access now goes through `resolve_binding`. Integration tests use new `mcp_binding_index` / `mcp_binding_vectorizer` helpers in tests/conftest.py that delegate to `resolve_binding` and preserve the same errors.

redisvl/mcp/server.py
vishal-bala: There is no CHANGELOG file in this repo, so documenting here per the request: `RedisVLMCPServer.run_guarded` now takes a required keyword-only `timeout_seconds` argument — a breaking change for any subclass or direct caller, which must pass `timeout_seconds=`. Happy to add this to the PR description too if preferred.

redisvl/mcp/server.py
```diff
@@ -259,12 +299,31 @@ async def _teardown_runtime(self, client: Any | None = None) -> None:
...
             elif client is not None:
                 await client.aclose()
 
+    async def _teardown_runtime(self) -> None:
+        """Release every binding's runtime resources and clear terminal state."""
+        bindings = list(self._bindings.values())
+        self._bindings = {}
+        self.config = None
+        self._semaphore = None
+        self._tools_registered = False
```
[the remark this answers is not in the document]
vishal-bala: Fixed in cb9a12b by removing the `_tools_registered = False` reset from `_teardown_runtime`. You're right: the lifecycle allows startup() again from STOPPED and `_register_tools()` runs in `_initialize_runtime_resources`, so resetting the flag made a restart re-register the same tool names. Tools are registered once on the FastMCP instance and their closures resolve the live binding at call time, so they survive teardown — the flag now correctly stays set across a stop/start.

redisvl/mcp/server.py
vishal-bala: Follow-up: a later Bugbot review flagged that resetting this flag on every teardown makes a stop/start re-register tools on the same FastMCP instance. After tracing it, the reset was reverted in cb9a12b. Your underlying concern still holds without it: `_tools_registered` is only set after registration succeeds, and registrations persist on the instance across teardown, so a retry that skips registration still has working tools.

redisvl/mcp/server.py
```diff
@@ -90,12 +93,10 @@ async def startup(self) -> None:
         """Load config, inspect the configured index, and initialize dependencies."""
         async with self._transition_lock:
             await self._begin_startup()
-            client = None
             try:
-                client = await self._initialize_runtime_resources()
+                await self._initialize_runtime_resources()
                 await self._mark_running()
             except Exception:
-                await self._teardown_runtime(client)
```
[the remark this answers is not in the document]
vishal-bala: Fixed in 118b01d. Moved binding teardown into startup()'s exception handler so it covers any post-begin failure (including _mark_running), not just failures inside _initialize_runtime_resources — resources are no longer leaked when init succeeds but a later step raises. _initialize_runtime_resources dropped its now-redundant inner teardown (a binding that fails mid-build still closes its own bare client in _initialize_binding), leaving a single fail-closed teardown path. Added a regression test for the init-succeeds-then-post-step-fails case.

nkanu17 (approved):
Thank you for the changes! Looks good! 🚀

PR #630, opened 2026-06-16
feat(mcp): add list-indexes discovery tool (RAAE-1605)

vishal-bala:
> Stacked on #629 (RAAE-1604). Review/merge that first; this PR targets the 1604 branch so the diff is scoped to the discovery tool.

## Motivation

Once a single MCP server can expose multiple logical indexes ([RAAE-1604](https://redislabs.atlassian.net/browse/RAAE-1604)), clients need a lightweight way to discover what's available so they can pick the right index instead of guessing. This PR ([RAAE-1605](https://redislabs.atlassian.net/browse/RAAE-1605)) adds an always-registered, read-only `list-indexes` tool for exactly that, and it grounds discovery in the schema the server already inspected at startup rather than asking users to re-declare field metadata in config.

## Implementation

The new tool (`redisvl/mcp/tools/list_indexes.py`) returns one entry per configured binding, in configured order: the logical `id`, an optional `description`, an `upsert_available` flag, the shared filterable `fields`, and — only when explicitly configured — a `limits` object. `upsert_available` is simply `not effective_read_only`, so it already reflects both the global `--read-only` flag and the per-index `read_only` policy resolved at startup. The `fields` list is built from the binding's effective (inspected + overridden) schema that already lives on its `BindingRuntime`, so the output stays consistent with what the index actually contains; the vector field and the configured default embed-source text field are omitted because they are implementation inputs rather than fields a client would filter on. The Redis index name (`redis_name`) is deliberately never exposed. Limits are included only when the operator set them explicitly — detected via the runtime model's `model_fields_set` — so defaults don't masquerade as deliberate overrides; per the contract this covers `max_limit` and `max_upsert_records`.

The tool is registered unconditionally during the server's tool registration (alongside `search-records` and the conditionally-registered `upsert-records`) and is gated by the same read scope as search when auth is enabled, since it is read-only.

- Output is deterministic and ordered by configured binding.
- No new configuration surface or settings are required.

## Verification

- `mypy` clean; `black`/`isort` formatted.
- New unit coverage: field omission (vector + embed-source), description/limits inclusion rules, `redis_name` secrecy, read-only reflection, single- and multi-binding output, and tool registration.
- New integration test starts a real two-binding server (one vector, one fulltext) and asserts the discovered fields come from the inspected schema and follow the omission rules.
- Full MCP suite green: 178 unit + 45 integration (2 skipped on Redis-version gates) against Redis 8.

[RAAE-1604]: https://redislabs.atlassian.net/browse/RAAE-1604?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[RAAE-1605]: https://redislabs.atlassian.net/browse/RAAE-1605?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

---

> [!NOTE]
> **Low Risk**
> Read-only discovery metadata with no new config; behavior is additive and gated like existing search tools when auth is enabled.
> 
> **Overview**
> Adds a read-only **`list-indexes`** MCP tool so clients can discover logical indexes on multi-binding servers before calling **`search-records`** or **`upsert-records`**.
> 
> The tool is **always registered** during `_register_tools` (alongside search; upsert remains conditional). Each binding is returned in config order with **`id`**, optional **`description`**, **`upsert_available`** (`not effective_read_only`), filterable **`fields`** from the startup-inspected schema (vector and default embed-source text omitted), and **`limits`** only when **`max_limit`** / **`max_upsert_records`** were explicitly set. **`redis_name`** is never exposed; empty bindings raise **`RuntimeError`** like other pre-startup paths. When auth is on, the tool uses the same **read scope** as search.
> 
> Unit and integration tests cover payload rules, registration, and a two-binding startup scenario.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 45ce686e0ef11a4872cab82836aed3eccf5be659. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/mcp/tools/list_indexes.py
+++ b/redisvl/mcp/tools/list_indexes.py
@@ -1,0 +1,85 @@
+from typing import Any
+
+from redisvl.mcp.auth import ensure_tool_scope
+
+DEFAULT_LIST_INDEXES_DESCRIPTION = (
+    "List the logical indexes configured on this server. Each entry reports the "
+    "index id, an optional description, whether upsert is available, the "
+    "filterable fields discovered from the index, and any explicitly configured "
+    "limits. Call this first on a multi-index server to choose the correct "
+    "index for search-records or upsert-records."
+)
+
+# Runtime limits surfaced to clients, included only when explicitly configured.
+_LIMIT_FIELDS = ("max_limit", "max_upsert_records")
+
+
+def _binding_fields(rt: Any) -> list[dict[str, str]]:
+    """Return a binding's shared filterable fields from its inspected schema.
+
+    The vector field and the configured default embed-source text field are
+    omitted: they are implementation inputs, not fields a client filters on.
+    """
+    embed_source = rt.binding.runtime.default_embed_text_field
+    fields: list[dict[str, str]] = []
+    for field in rt.schema.fields.values():
+        field_type = str(getattr(field.type, "value", field.type))
+        if field_type.lower() == "vector":
+            continue
+        if field.name == embed_source:
+            continue
+        fields.append({"name": field.name, "type": field_type})
+    return fields
+
+
+def _binding_limits(rt: Any) -> dict[str, int]:
+    """Return runtime limits that were explicitly configured for the binding.
+
+    Defaults are intentionally excluded so the output reflects deliberate
+    overrides rather than implementation defaults.
+    """
+    runtime = rt.binding.runtime
+    configured = runtime.model_fields_set
+    return {
+        name: getattr(runtime, name) for name in _LIMIT_FIELDS if name in configured
+    }
+
+
+def _describe_binding(rt: Any) -> dict[str, Any]:
+    """Build the deterministic discovery payload for a single binding."""
+    entry: dict[str, Any] = {"id": rt.binding_id}
+    if rt.binding.description is not None:
+        entry["description"] = rt.binding.description
+    # Reflects both global read-only and the per-index read_only policy.
+    entry["upsert_available"] = not rt.effective_read_only
+    entry["fields"] = _binding_fields(rt)
+    limits = _binding_limits(rt)
+    if limits:
+        entry["limits"] = limits
+    return entry
+
+
+def list_indexes(server: Any) -> dict[str, Any]:
+    """Return the discovery payload for every configured binding.
+
+    The Redis index name (``redis_name``) is intentionally never exposed.
+    """
+    return {
+        "indexes": [_describe_binding(rt) for rt in server._bindings.values()],
+    }
+
+
+def register_list_indexes_tool(server: Any) -> None:
+    """Register the always-available, read-only `list-indexes` MCP tool."""
+    description = (
+        getattr(server.mcp_settings, "tool_list_indexes_description", None)
+        or DEFAULT_LIST_INDEXES_DESCRIPTION
+    )
+
+    async def list_indexes_tool():
+        """FastMCP wrapper for the `list-indexes` tool."""
+        read_scope = getattr(getattr(server, "auth_config", None), "read_scope", None)
+        ensure_tool_scope(server, read_scope)
+        return list_indexes(server)
+
+    server.tool(name="list-indexes", description=description)(list_indexes_tool)
--- a/redisvl/mcp/tools/list_indexes.py
+++ b/redisvl/mcp/tools/list_indexes.py
@@ -1,6 +1,10 @@
-from typing import Any
+from typing import TYPE_CHECKING, Any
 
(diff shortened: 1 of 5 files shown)


vishal-bala:
CI tests are failing due to flaky issues being fixed by https://github.com/redis-developer/sql-redis/pull/39

redisvl/mcp/tools/list_indexes.py
```diff
@@ -0,0 +1,92 @@
...
+        ],
+    }
+
+
+def register_list_indexes_tool(server: "RedisVLMCPServer") -> None:
+    """Register the always-available, read-only `list-indexes` MCP tool."""
+    description = (
+        getattr(server.mcp_settings, "tool_list_indexes_description", None)
+        or DEFAULT_LIST_INDEXES_DESCRIPTION
+    )
```
nkanu17: Dead code, `description` is not used.
```suggestion
```

redisvl/mcp/tools/list_indexes.py
```diff
@@ -0,0 +1,92 @@
...
+
+    async def list_indexes_tool():
+        """FastMCP wrapper for the `list-indexes` tool."""
+        read_scope = getattr(getattr(server, "auth_config", None), "read_scope", None)
```
nkanu17: This might need a rethink:

The double `getattr` with `None` fallbacks might not work as intended/fragile:
-  if `auth_config` exists but `read_scope` lives somewhere else, we’ll silently end up with `None`, and then `ensure_tool_scope` gets called with `None`. That could accidentally skip auth enforcement.

Claude tells me this also happens in `search.py`,

nkanu17 (changes requested):
This generally looks great! I ran into one thing that specified in the comments, it is worth checking the auth access pattern across the tools.

redisvl/mcp/tools/list_indexes.py
vishal-bala: Fixed in e8d750c. `description` was technically used (passed to `server.tool(..., description=description)`), but the `tool_list_indexes_description` setting it read doesn't exist on MCPSettings (only tool_search/upsert_description do), so that override branch was always None — dead. Removed it and pass the default description constant directly.

redisvl/mcp/tools/list_indexes.py
vishal-bala: Fixed in e8d750c. read_scope/write_scope are typed `str | None` fields on MCPAuthConfig, so I now read `auth_config.read_scope` directly instead of `getattr(..., "read_scope", None)`. The old form would fail open — silently yielding None and skipping the scope check — if the field were renamed; direct access fails loud. Applied the same fix to search.py and upsert.py (commit 3e441b2 on the 1604 branch) so all three wrappers match.

nkanu17 (approved):
lgtm

redisvl/mcp/tools/list_indexes.py
vishal-bala: Already addressed in the current branch (commit 2276985): the dead `tool_list_indexes_description` override was removed and the default description constant is passed directly to `server.tool(...)`.

redisvl/mcp/tools/list_indexes.py
vishal-bala: Already addressed (commit 2276985): the wrapper now reads `auth_config.read_scope` directly off the typed MCPAuthConfig instead of the fragile double-getattr, so a renamed field fails loud rather than silently skipping auth. The same fix was applied to search.py/upsert.py and landed with #629.

redisvl/mcp/tools/list_indexes.py
```diff
@@ -0,0 +1,91 @@
...
+    """Return the discovery payload for every configured binding.
+
+    The Redis index name (``redis_name``) is intentionally never exposed.
+    """
+    return {
+        "indexes": [
+            _describe_binding(binding_runtime)
+            for binding_runtime in server._bindings.values()
+        ],
+    }
```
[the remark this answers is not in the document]
vishal-bala: Fixed in 45ce686: list_indexes now raises the same "MCP server has not been started" RuntimeError as resolve_binding when _bindings is empty, instead of returning {"indexes": []}. Added a regression test.

vishal-bala:
Failing tests are because of the Redis 8.8 flakiness issue - proceeding with the merge

PR #635, opened 2026-06-23
fix(router): remove persisted route_config key on SemanticRouter.delete()

### LINKED ISSUE #634: SemanticRouter.delete() leaves orphaned {name}:route_config key in Redis
## Summary

`SemanticRouter.delete()` drops the router's search index but never removes the JSON config key the router persists at `{name}:route_config`. That key is leaked into Redis for every router that is deleted — affecting both real users and the test suite.

This was split out of the test-isolation hardening work (#546) because it requires a product-code change and is logically distinct from the test-only naming/cleanup fixes.

## Root cause

- The config key is written on init and on route changes:
  - `redisvl/extensions/router/semantic.py:116` — `self._index.client.json().set(f"{self.name}:route_config", ".", self.to_dict())`
  - `redisvl/extensions/router/semantic.py:881` — same write on route mutation
- `delete()` only drops the index, not the config key:
  - `redisvl/extensions/router/semantic.py:579` — `def delete(self): self._index.delete(drop=True)`
- `clear()` similarly only clears indexed docs.

So after `delete()`, `{name}:route_config` remains in Redis.

## Impact

- **Product:** Any user calling `router.delete()` leaves a stranded `:route_config` key. A later `SemanticRouter.from_existing(name)` (`semantic.py:154`) could read stale config from a router that was supposedly deleted.
- **Tests:** Confirmed orphaned keys in:
  - The `semantic_router` fixture in `tests/integration/test_semantic_router.py` (clear() + delete(), config key survives).
  - `tests/integration/test_key_separator_handling.py` — `test_semantic_router_uses_index_separator` (`test_router_sep`), `test_router_respects_modified_key_separator` (`router_sep_test`), `test_router_with_prefix_ending_in_separator` (`router_trailing_test`) create routers with fixed names and never clean up; the index can be dropped but the config key persists.

## Proposed change

- Make `SemanticRouter.delete()` (and the async equivalent if present) also remove `{self.name}:route_config`, so deletion fully removes router state.
- Update the affected tests to rely on that (and/or wrap their router creation in try/finally), converting the fixed router names in `test_key_separator_handling.py` to per-test names via the `redis_test_name` fixture.

## Acceptance criteria

- After `router.delete()`, no `{name}:route_config` key remains in Redis.
- `SemanticRouter.from_existing()` on a deleted router name fails/returns nothing rather than reading stale config.
- Router-related integration tests leave no `:route_config` keys behind after teardown.

vishal-bala:
## Motivation

A `SemanticRouter` persists its configuration as a standalone JSON key, `{name}:route_config`, written when the router is constructed and rewritten on every route change. `delete()`, however, only drops the router's search index — it never removes that key. The index is built over hash documents, and `FT.DROPINDEX ... DD` deletes only the documents the index actually tracks, so the JSON config key falls outside its reach and survives every call to `delete()`.

The immediate consequence is orphaned `:route_config` keys accumulating in Redis. The subtler one is a correctness gap: `from_existing()` loads a router straight from that key, so a router that was supposedly deleted can be silently resurrected from stale config. It also leaves residue behind in the test suite, where it contributed to the broader Redis-backed isolation problems tracked in #546.

## Fix

`delete()` now removes `{name}:route_config` explicitly after dropping the index, so deleting a router fully clears its state. The change is confined to `SemanticRouter.delete()`; there is no async router variant to mirror.

A regression test constructs a router, confirms the config key exists, calls `delete()`, and then asserts both that the key is gone and that `from_existing()` on the same name raises rather than reloading stale config — locking in the behaviour described above.

Closes #634.

---

> [!NOTE]
> **Low Risk**
> Small lifecycle fix in router teardown with a focused regression test; no changes to routing, auth, or data semantics beyond complete cleanup on delete.
> 
> **Overview**
> **`SemanticRouter.delete()`** now drops the search index and explicitly deletes the standalone Redis JSON key `{name}:route_config`, which is not removed by `FT.DROPINDEX` because it sits outside the index.
> 
> This prevents orphaned config keys and stops **`from_existing()`** from reloading a router that was supposed to be deleted.
> 
> An integration test asserts the config key is gone after **`delete()`** and that **`from_existing()`** raises when no valid config remains.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 23805cff1c06371b2377abeee442b09dd1b09a74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -577,8 +577,11 @@ class SemanticRouter(BaseModel):
             self._update_router_state()
 
     def delete(self) -> None:
-        """Delete the semantic router index."""
+        """Delete the semantic router index and its persisted route config."""
         self._index.delete(drop=True)
+        # The route config is stored as a standalone JSON key that is not
+        # tracked by the search index, so it must be removed explicitly.
+        self._index.client.delete(f"{self.name}:route_config")  # type: ignore
 
     def clear(self) -> None:
         """Flush all routes from the semantic router index."""
--- a/tests/integration/test_semantic_router.py
+++ b/tests/integration/test_semantic_router.py
@@ -685,6 +685,34 @@ def test_from_existing(client, redis_url, routes, redis_test_name):
                 router.delete()
 
 
+def test_delete_removes_route_config_key(
+    client, routes, hf_vectorizer, redis_test_name
+):
+    """delete() must remove the persisted {name}:route_config key (issue #634)."""
+    skip_if_no_redis_search(client)
+
+    router = SemanticRouter(
+        name=redis_test_name("test_router_delete"),
+        routes=routes,
+        routing_config=RoutingConfig(max_k=2),
+        redis_client=client,
+        overwrite=True,
+        vectorizer=hf_vectorizer,
+    )
+    config_key = f"{router.name}:route_config"
+
+    # The config key is written when the router is constructed.
+    assert client.exists(config_key)
+
+    router.delete()
+
+    # delete() must drop the index AND remove the standalone config key, so a
+    # subsequent from_existing() finds no stale config to load.
+    assert not client.exists(config_key)
+    with pytest.raises(ValueError):
+        SemanticRouter.from_existing(name=router.name, redis_client=client)
+
+
 def test_get_route_references(semantic_router):
     # Get references for a specific route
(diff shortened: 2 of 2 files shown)


limjoobin (approved):
LGTM!

nkanu17 (approved):

PR #636, opened 2026-06-23
test: harden Redis-backed test isolation under pytest-xdist (#546)

vishal-bala:
## Motivation

The integration suite runs under `pytest -n auto`, and now and then a single matrix job fails on a test that passes everywhere else — and passes again on rerun. The most recent example was `test_hybrid_query_with_geo_filter` returning 0 results instead of 3 on exactly one Python/redis-py/Redis combination. These failures read as version-specific, but they aren't: they're a test-isolation problem wearing a disguise.

The key fact is how the suite isolates workers. Each xdist worker gets its **own** Redis container, so two tests can only interfere when they run on the **same** worker *and* reach for the **same** Redis key names. For that to actually corrupt a result, two things have to line up:

1. a **shared resource name** (a name scoped only to the worker, or a hard-coded one), and
2. a **way for state to outlive its test** — cleanup that only runs on the success path, or an index recreated with `overwrite=True` but no `drop=True`, leaving stale documents under a reused prefix.

When both hold, one test sees another's leftovers — empty results, inflated counts, dtype mismatches — and which test loses depends entirely on the order the scheduler picked. That's the retry-passing, pseudo-random signature.

#543 introduced a `redis_test_name` helper that mints per-test-unique names and began migrating fixtures onto it. This PR finishes that migration for the modules that can genuinely cause flakiness, and tightens their recreate/cleanup so leftovers can't survive.

## What changed

**Recreate group** — per-test names + `create(overwrite=True, drop=True)`:
- `test_hybrid.py` and `test_aggregation.py` (these two shared the *identical* `user_index_{worker_id}` / `v1_{worker_id}` — the direct cause of the geo-filter failure), plus `test_search_results.py`, `test_svs_integration.py`, and the shared `conftest.py` index fixtures.

**Fixed-name collisions** — per-test names + guaranteed teardown:
- `from_existing_complex` (hard-coded `"test"`) and `no_proactive_module_validation` (`"my_index"`) in the sync + async search-index suites.
- The three `test_embedcache_warnings.py` tests sharing `"test_cache"`.
- The shared `"doc"` prefix / `"test-index"` name across `test_no_proactive_module_checks.py`.

**Raw `worker_id` collision** — unique names + cleanup:
- The two `test_llmcache.py` tests that both created `float64_cache_{worker_id}` and never cleaned up.

**Cluster tests** — the one place cross-worker collisions are real, since every cluster test points at a single hard-coded `redis://localhost:7001`: `test_redis_cluster_support.py` and `test_cluster_pipelining.py` now use per-test names, `drop=True`, and `try/finally` cleanup.

## Scope and follow-ups

- The `SemanticRouter.delete()` leak of its `{name}:route_config` key is a product-code gap, not a test-only fix, so it's split into #634 (which also owns the dependent `test_key_separator_handling.py` cleanup).
- The remaining #546 items are **benign hygiene** — leaks under per-test-unique names that can't contaminate another test, only accumulate harmless keys in a worker's throwaway Redis. Their exact remaining scope is enumerated in a comment on #546.

## Testing

Honest caveat for reviewers: this was developed in an environment without Docker, so I could **not** run the integration suite. `black`, `isort`, and `mypy` pass. Please run `make test` and a repeated `pytest -n auto` pass to confirm the flake is gone before merging — this is a test-only change, so behavior is fully validated by the suite itself.

---

> [!NOTE]
> **Low Risk**
> Changes are confined to the test suite; production library behavior is unchanged.
> 
> **Overview**
> **Hardens integration tests** so parallel `pytest-xdist` workers and shared Redis/cluster endpoints do not leave stale keys or collide on index/cache names.
> 
> Fixtures and tests now use **`redis_test_name`** (per-test unique index/prefix/cache names) instead of **`worker_id`-only** or hard-coded names like `"test"` / `"my_index"` / `"test_cache"`. Shared **`conftest`** index fixtures (`flat_index`, `hnsw_index`, async variants) and modules such as **`test_hybrid`**, **`test_aggregation`**, **`test_search_results`**, and **`test_svs_integration`** adopt that naming and call **`create(overwrite=True, drop=True)`** so documents under a reused prefix are cleared before load.
> 
> **Teardown** is tightened with **`try/finally`**, **`delete(drop=True)`**, and cache **`clear()`** / **`aclear()`** on failure paths (e.g. `from_existing_complex`, module-validation tests, **`test_llmcache`** dtype caches, **`test_embedcache_warnings`**). **Cluster** tests use per-test names and **`drop=True`** because all workers share **`localhost:7001`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c4e6b33adf577d252f4d22a99d98fe1fa6983189. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -518,7 +518,7 @@ def flat_index(sample_data, redis_url, redis_test_name):
     )
 
     # create the index (no data yet)
-    index.create(overwrite=True)
+    index.create(overwrite=True, drop=True)
 
     # Prepare and load the data
     def hash_preprocess(item: dict) -> dict:
@@ -575,7 +575,7 @@ async def async_flat_index(sample_data, redis_url, redis_test_name):
     )
 
     # create the index (no data yet)
-    await index.create(overwrite=True)
+    await index.create(overwrite=True, drop=True)
 
     # Prepare and load the data
     def hash_preprocess(item: dict) -> dict:
@@ -631,7 +631,7 @@ async def async_hnsw_index(sample_data, redis_url, redis_test_name):
     )
 
     # create the index (no data yet)
-    await index.create(overwrite=True)
+    await index.create(overwrite=True, drop=True)
 
     # Prepare and load the data
     def hash_preprocess(item: dict) -> dict:
@@ -687,7 +687,7 @@ def hnsw_index(sample_data, redis_url, redis_test_name):
     )
 
     # create the index (no data yet)
-    index.create(overwrite=True)
+    index.create(overwrite=True, drop=True)
 
     # Prepare and load the data
     def hash_preprocess(item: dict) -> dict:
--- a/tests/integration/test_aggregation.py
+++ b/tests/integration/test_aggregation.py
@@ -8,13 +8,13 @@ from tests.conftest import skip_if_redis_version_below
 
 
 @pytest.fixture
-def index(multi_vector_data, redis_url, worker_id):
+def index(multi_vector_data, redis_url, redis_test_name):
 
     index = SearchIndex.from_dict(
         {
             "index": {
-                "name": f"user_index_{worker_id}",
-                "prefix": f"v1_{worker_id}",
+                "name": redis_test_name("user_index"),
+                "prefix": redis_test_name("v1"),
                 "storage_type": "hash",
             },
             "fields": [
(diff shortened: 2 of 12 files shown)


limjoobin (approved):
LGTM! It is clear that the `create(overwrite=True, drop=True)` and `delete(drop=True)` patterns were used in tandem to ensure that namespace isolation is guaranteed on both setup and teardown, rather than depending on test success and teardown running. This ensures that the index is definitely clean when tests are run on it. 

Additionally, it was the right call to use `redis_test_name` to ensure no namespace collisions across tests.

nkanu17 (approved):
lgtm! thank you for addressing this

PR #637, opened 2026-06-23
fix(router): remove orphaned :route_config key on SemanticRouter.delete() (#634)

### LINKED ISSUE #634: SemanticRouter.delete() leaves orphaned {name}:route_config key in Redis

ngochuyinfo-crypto:
Fixes #634.

## Problem
`SemanticRouter` persists config at `{name}:route_config` but `delete()` only dropped the search index, leaving that JSON key orphaned. A later `from_existing(name)` could read stale config from a deleted router.

## Fix
`delete()` now also removes the `{name}:route_config` key.

## Tests
- Add `test_delete_removes_route_config_key`: asserts config key is gone after `delete()` and `from_existing()` raises instead of reading stale config.
- Wrap fixed-name routers in `test_key_separator_handling.py` in try/finally so they clean up their index + config key.

## Verification
```
4 passed in 138.97s
```
Ran against redis:8.4 (Search + JSON) with the real HF vectorizer.

---

> [!NOTE]
> **Low Risk**
> Small, targeted lifecycle fix in router teardown plus test cleanup; no auth or core routing logic changes.
> 
> **Overview**
> **`SemanticRouter.delete()`** now removes the persisted **`{name}:route_config`** JSON key in addition to dropping the search index, so deleted routers do not leave stale config that **`from_existing()`** could still load.
> 
> A new integration test checks the config key is absent after **`delete()`** and that **`from_existing()`** fails on that name. Key-separator integration tests that create fixed-name routers now use **`try`/`finally`** with **`router.delete()`** so index and config are cleaned up.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce14afe80bce793875a759cb0633b66f750006c1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/extensions/router/semantic.py
+++ b/redisvl/extensions/router/semantic.py
@@ -577,8 +577,9 @@ class SemanticRouter(BaseModel):
             self._update_router_state()
 
     def delete(self) -> None:
-        """Delete the semantic router index."""
+        """Delete the semantic router index and its persisted config."""
         self._index.delete(drop=True)
+        self._index.client.delete(f"{self.name}:route_config")  # type: ignore
 
     def clear(self) -> None:
         """Flush all routes from the semantic router index."""
--- a/tests/integration/test_key_separator_handling.py
+++ b/tests/integration/test_key_separator_handling.py
@@ -99,21 +99,26 @@ class TestKeySeparatorHandling:
             overwrite=True,
         )
 
-        # Modify the index schema to use custom separator
-        router._index.schema.index.key_separator = "|"
-        router._index.schema.index.prefix = "router"
+        try:
+            # Modify the index schema to use custom separator
+            router._index.schema.index.key_separator = "|"
+            router._index.schema.index.prefix = "router"
 
-        # Check that route reference keys use the custom separator
-        route_key = router._route_ref_key(router._index, "test_route", "ref123")
+            # Check that route reference keys use the custom separator
+            route_key = router._route_ref_key(router._index, "test_route", "ref123")
 
-        # Should use custom separator
-        assert "|" in route_key, f"Route key doesn't use custom separator: {route_key}"
-        assert (
-            route_key.count(":") == 0
-        ), f"Route key uses default separator: {route_key}"
-        assert (
-            route_key == "router|test_route|ref123"
-        ), f"Unexpected route key: {route_key}"
+            # Should use custom separator
+            assert (
+                "|" in route_key
+            ), f"Route key doesn't use custom separator: {route_key}"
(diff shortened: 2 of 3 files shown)


vishal-bala:
Thanks for your contribution, but we already have a PR in-flight for this issue: #635 

(Seems like this is a bot in any case)

PR #652, opened 2026-07-23
feat: GoogleGenAIVectorizer — migrate Google embeddings to the supported google-genai SDK

### LINKED ISSUE #620: VertexAI vectorizer uses deprecated model garden SDK (removal June 2026)
`VertexAIVectorizer._client` calls `TextEmbeddingModel.from_pretrained()` and `MultiModalEmbeddingModel.from_pretrained()` from `vertexai.language_models` / `vertexai.vision_models`. These are deprecated as of June 24, 2025 and will be removed June 24, 2026.

Triggers on every vectorizer init:
```
UserWarning: This feature is deprecated as of June 24, 2025 and will be removed on June 24, 2026.
vertexai/_model_garden/_model_garden_models.py:278
```

Needs migration to the new `google-genai` SDK. Reproduced on 0.18.2.

vishal-bala:
## Why

`VertexAIVectorizer` builds embeddings through Google's Vertex AI **model-garden** SDK (`TextEmbeddingModel` from `vertexai.language_models`, `MultiModalEmbeddingModel` from `vertexai.vision_models`). Google deprecated those modules on 2025-06-24 with a stated removal date of 2026-06-24, so today every `VertexAIVectorizer()` init prints:

```
UserWarning: This feature is deprecated as of June 24, 2025 and will be removed on June 24, 2026.
```

To be precise about the state of the world (verified against PyPI, not just the docs): the modules are **not yet hard-removed** — `google-cloud-aiplatform==1.162.0` still imports and runs them, and the `2.0.0` that would drop them is **yanked**. So the old path *works today but is living on borrowed time*: noisy, on a removal clock, and on an SDK Google has stopped investing in for gen-AI.

The replacement Google points to is [`google-genai`](https://pypi.org/project/google-genai/). Moving to it silences the warning, runs on the supported SDK, unlocks the Gemini Developer API and real async, and sheds the ~200 MB memory overhead of `google-cloud-aiplatform` ([python-aiplatform#6191](https://github.com/googleapis/python-aiplatform/issues/6191)).

Closes #620 (text embeddings; see *Caveats* for multimodal).

## The mental model

Three ideas carry the whole change:

**1. One client, two backends.** `google-genai` unifies what used to be two SDKs. The only difference between enterprise Vertex AI and the consumer Gemini API is how you build the client; after that, `embed_content` is identical (and `client.aio` gives real async for free):

```python
from redisvl.utils.vectorize import GoogleGenAIVectorizer

# Vertex AI (GCP project auth)
vectorizer = GoogleGenAIVectorizer(
    model="gemini-embedding-001",
    api_config={"project_id": "your-gcp-project", "location": "us-central1"},
)

# Gemini Developer API (single key)
vectorizer = GoogleGenAIVectorizer(api_config={"api_key": "..."})  # or set GEMINI_API_KEY

embedding = vectorizer.embed("Hello, world!")
embeddings = await vectorizer.aembed_many(["a", "b"], batch_size=2)
```

**2. Explicit beats ambient.** Auth can come from `api_config` or the environment, and real deployments often have *both* a `GOOGLE_CLOUD_PROJECT` and a stray `GOOGLE_API_KEY` lying around. So backend selection is deterministic and documented — explicit override → explicit `api_config` creds → environment — rather than "guess from whatever's set":

```python
# Force a backend regardless of ambient env:
GoogleGenAIVectorizer(api_config={"backend": "gemini", "api_key": "..."})
vectorizer.backend  # -> "gemini"  (resolved backend is observable)
```

If both a project and a key are resolvable from the environment, Vertex wins; if neither is, you get a `ValueError` that names both option sets.

**3. Errors must retry, and must not leak.** Transient API errors propagate into the tenacity `@retry` (the old vectorizer wrapped everything in `ValueError`, which its own retry filter then excluded — so it never actually retried). Input-type errors fail fast as `TypeError`. And because the Gemini key rides in a request header that transport errors can echo, we never interpolate a raw provider exception into a message on a key-bearing path.

Embeddings are returned exactly as the provider produces them — no post-processing — matching every other vectorizer in the library. `output_dimensionality` is an optional passthrough for Google's Matryoshka reduction (shorter vectors, ~4× less index memory) that also sets `dims` correctly and is fixed for the vectorizer's lifetime:

```python
v = GoogleGenAIVectorizer(output_dimensionality=768)
v.dims  # 768
```

(Note: Google does not re-normalize reduced-dim vectors. That's invisible under the COSINE metric, which is scale-invariant, and left to the caller for inner-product / L2 — we don't silently mutate provider output.)

## What changed

- **New** [`redisvl/utils/vectorize/googlegenai.py`](redisvl/utils/vectorize/googlegenai.py) — `GoogleGenAIVectorizer` (both backends, sync + async, caching, secrets-safe errors). Registered in the `Vectorizers` enum and `vectorizer_from_dict`.
- **Deprecation** — `VertexAIVectorizer` now emits a *teaching* `DeprecationWarning` (names the new class, gives the model/dim mapping, links #620). `deprecated_class` is now idempotent per instance, so the already-deprecated `VertexAITextVectorizer` alias (which subclasses it) warns exactly **once**.
- **Dependencies** — new `google-genai` extra (`pip install redisvl[google-genai]`), added to `all`. The legacy `vertexai` extra is unchanged, but its `<2.0.0` cap is now **documented** as intentional (it keeps the still-working 1.x line that ships the deprecated modules, incl. the multimodal path).
- **Tests** — offline unit tests (fake `google.genai` module + an autouse env-scrub fixture, since CI's `make test` exports `GCP_*` and would otherwise skew backend detection) covering backend resolution, both embed paths, async, retry, the secrets-never-leak guarantee, and the single-deprecation-warning behavior; integration params wired for both backends.
- **Docs** — API reference, the vectorizers user-guide notebook, `installation.md`, and `utilities.md`.
- **CI** — the existing `GEMINI_API_KEY` secret is wired into the fork-guarded `service-tests` job only.

## Caveats & future steps

- **Text embeddings only.** Multimodal (image/video) is still public preview on `google-genai`; `VertexAIVectorizer` remains the (deprecated) path for it, and its migration is tracked in #620. The new class lives at the package root with a backend-neutral name precisely so multimodal can be added later without a rename.
- **The old class still works** — deprecate-now / remove-next-major. No user is broken by this PR.
- **Notebook validation & live APIs (out of scope here).** The new notebook cell is marked `# NBVAL_SKIP`, matching the existing precedent in `13_langcache`, so it adds no API calls to notebook validation. However, several *existing* cells in `04_vectorizers.ipynb` (and `03`, `06`) already make live provider API calls under `--nbval-lax`. Making notebook validation credential-free across the board is a worthwhile, separate cleanup.
- **`google-genai` version floor** is pinned `>=1.0.0`; worth confirming the earliest release that exposes `embed_content` + `.aio` + `EmbedContentConfig` before merge.

## Verification

- `make format` clean; `mypy redisvl` clean; pre-commit hooks pass.
- New unit tests pass; full offline unit suite green.
- `python -m tests.test_imports redisvl` passes with `google-genai` **not** installed (lazy import).

---

> [!NOTE]
> **Medium Risk**
> Touches embedding/auth paths and default model/dimension changes (reindex required); legacy Vertex remains for multimodal until a follow-up.
> 
> **Overview**
> Adds **`GoogleGenAIVectorizer`** on the supported `google-genai` SDK for text embeddings, with auto-selection of **Vertex AI** (project/location/ADC) or the **Gemini Developer API** (`GEMINI_API_KEY`), plus sync/async batching, retries, and safer error messages that avoid leaking credentials.
> 
> **`VertexAIVectorizer`** and **`VertexAITextVectorizer`** are now deprecated in favor of the new class; **`deprecated_class`** warns only once per instance so subclass chains do not double-warn. Wiring includes a new **`google_genai`** `Vectorizers` enum value, **`vectorizer_from_dict`**, optional extra **`redisvl[google-genai]`** (and **`all`**), **`GEMINI_API_KEY`** in service CI, docs/notebook migration guidance, and unit + integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c543fbd7ec90fb95cba6a2809be9b84b6fcbb61e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -37,6 +37,7 @@ jobs:
           OPENAI_API_KEY: ${{ secrets.OPENAI_API_KEY }}
           GCP_LOCATION: ${{ secrets.GCP_LOCATION }}
           GCP_PROJECT_ID: ${{ secrets.GCP_PROJECT_ID }}
+          GEMINI_API_KEY: ${{ secrets.GEMINI_API_KEY }}
           COHERE_API_KEY: ${{ secrets.COHERE_API_KEY }}
           MISTRAL_API_KEY: ${{ secrets.MISTRAL_API_KEY }}
           VOYAGE_API_KEY: ${{ secrets.VOYAGE_API_KEY }}
--- a/docs/api/vectorizer.rst
+++ b/docs/api/vectorizer.rst
@@ -8,7 +8,8 @@ Vectorizers
    compatibility:
 
    - ``VoyageAITextVectorizer`` → Use ``VoyageAIVectorizer`` instead
-   - ``VertexAITextVectorizer`` → Use ``VertexAIVectorizer`` instead
+   - ``VertexAITextVectorizer`` → Use ``GoogleGenAIVectorizer`` instead
+     (``VertexAIVectorizer`` is itself deprecated; see below)
    - ``BedrockTextVectorizer`` → Use ``BedrockVectorizer`` instead
    - ``CustomTextVectorizer`` → Use ``CustomVectorizer`` instead
 
@@ -59,11 +60,27 @@ VertexAIVectorizer
 .. currentmodule:: redisvl.utils.vectorize.vertexai
 
 .. note::
-    For backwards compatibility, an alias ``VertexAITextVectorizer`` is available
-    in the ``redisvl.utils.vectorize.text`` module. This alias is deprecated
-    as of version 0.13.0 and will be removed in a future major release.
+    ``VertexAIVectorizer`` is **deprecated**. It uses Google's Vertex AI
+    model-garden SDK, which Google has deprecated with a scheduled removal. Use
+    :class:`~redisvl.utils.vectorize.googlegenai.GoogleGenAIVectorizer` for text
+    embeddings on the supported ``google-genai`` SDK. The alias
+    ``VertexAITextVectorizer`` (in ``redisvl.utils.vectorize.text``) is likewise
+    deprecated. Multimodal (image/video) migration is tracked in
+    `issue #620 <https://github.com/redis/redis-vl-python/issues/620>`_.
 
 .. autoclass:: VertexAIVectorizer
+   :show-inheritance:
+   :members:
+
+
+GoogleGenAIVectorizer
+=====================
+
+.. _googlegenaivectorizer_api:
+
+.. currentmodule:: redisvl.utils.vectorize.googlegenai
+
+.. autoclass:: GoogleGenAIVectorizer
    :show-inheritance:
    :members:
 
--- a/docs/concepts/utilities.md
+++ b/docs/concepts/utilities.md
@@ -35,7 +35,7 @@ Vectorizers handle batching internally, breaking large batches into provider-app
 
 ### Supported Providers
 
-RedisVL includes vectorizers for OpenAI, Azure OpenAI, Cohere, HuggingFace (local), Mistral, Google Vertex AI, AWS Bedrock, VoyageAI, and others. See the {doc}`/api/vectorizer` for the complete list. You can also create custom vectorizers that wrap any embedding function.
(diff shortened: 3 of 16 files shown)


rbs333 (approved):
Good cleanup here def better than what exists today

justin-cechmanek (approved):
👍🏼

PR #608, opened 2026-05-08
docs: migrate from Sphinx to mkdocs material

nkanu17:
## Summary

Migrate the documentation pipeline from Sphinx to mkdocs material so the
public site renders with the same visual language used by sql-redis,
loads faster, and exposes a cleaner Python API surface generated by
mkdocstrings + griffe.

image

### Documentation toolchain
- Drop the entire `docs/_extension/`, `docs/_static/`, `docs/_templates/`,
  `docs/conf.py`, `docs/Makefile`, `docs/make.bat` Sphinx tree and every
  `.rst` source file.
- Add `mkdocs.yml` plus a Material-themed site under `docs/` rooted at
  `docs/index.md` with the Redis Vector Library hero, navigation tree,
  branded color palette (Redis red, Space Grotesk / Space Mono), and a
  full mkdocstrings-driven API reference under `docs/api/`.
- Convert the architecture, queries, search-and-indexing, extensions,
  field-attributes, mcp, and utilities concept pages from Sphinx-flavored
  myst to plain Markdown.
- Wire the existing user-guide notebooks (`docs/user_guide/*.ipynb`) into
  the site via `mkdocs-jupyter` and add `docs/user_guide/index.md` plus an
  installation page.
- Add machine-readable agent docs under `docs/for-ais-only/` (repository
  map, build/test, public API surface, codebase tour) and a top-level
  `AGENTS.md` entry point. The site emits `llms.txt` / `llms-full.txt`
  via `mkdocs-llmstxt`.
- Custom branding lives in `docs/stylesheets/redis-brand.css` plus
  `docs/assets/` (favicon, hero logo in Redis red, architecture
  diagram).

### Build pipeline
- Replace the Sphinx block in `.readthedocs.yaml` with a `mkdocs:` block
  and switch the build to `uv sync --frozen --group docs`.
- Update `Makefile` so `make docs-build` and `make docs-serve` invoke
  `mkdocs` instead of `sphinx-build` / `sphinx-autobuild`.
- Add a new `[dependency-groups].docs` block in `pyproject.toml`
  pinning `mkdocs`, `mkdocs-material`, `mkdocs-jupyter`,
  `mkdocs-autorefs`, `mkdocs-llmstxt`, `mkdocs-redirects`,
  `mkdocs-section-index`, `mkdocs-git-revision-date-localized-plugin`,
  and `mkdocstrings[python]`. Lockfile refreshed.
- Ignore the generated `site/` output via `.gitignore`.

### CI
[...]

PR #654, opened 2026-07-25
docs: add exception reference and error handling guide

### LINKED ISSUE #496: Document RedisVL exception hierarchy and error handling patterns
## Summary
Document RedisVL exception hierarchy and practical error handling patterns.

**Level:** Beginner

## Current State
- Custom exceptions are defined in `redisvl/exceptions.py`.
- Documentation does not currently consolidate exception guidance in one discoverable place.

## Proposed Change
- Add an API docs page for exceptions.
- Add user-facing guidance for common error categories and recommended handling.
- Include concise try/except examples for schema validation and query failures.

## Definition of Done
- Exception docs are present and discoverable in generated docs.
- `make docs-build` passes.
- Examples align with current exception classes.

## Out of Scope
- Redesigning exception class hierarchy.

TheSaiEaranti:
Fixes #496

Adds `docs/api/exceptions.rst` and links it from the API toctree, so the exception hierarchy has one discoverable home.

## What's on the page

- **The hierarchy** — all five classes in `redisvl.exceptions`, and a note that `redis-py` exceptions such as `redis.exceptions.ConnectionError` are *not* part of it. RedisVL wraps those and chains the original with `raise ... from e`, so `__cause__` still holds it. That seemed like the detail most likely to trip someone writing an `except` clause.
- **A table** mapping each exception to the operations that raise it, e.g. `SchemaValidationError` → `load()`, `RedisSearchError` → `create()` / `delete()` / `search()` / `aggregate()`.
- **Four `try`/`except` examples**, including the two the issue asks for.
- **Autodoc reference** for each class, following the `:members:` / `:show-inheritance:` pattern used in `schema.rst`.

## Examples are drawn from real raise sites

Rather than writing plausible-looking snippets, I traced each exception to where it is actually raised so the examples match current behaviour:

| Exception | Verified raise site | What the example shows |
| --- | --- | --- |
| `SchemaValidationError` | `storage.py` `_preprocess_and_validate_objects` | `validate_on_load=True`, since validation is off by default and the error cannot otherwise occur |
| `QueryValidationError` | `index.py` `_validate_query` | `ef_runtime` on a `flat` vector field — the concrete case that method rejects |
| `RedisSearchError` | `create`, `delete`, `search`, `aggregate` | wrapping a Redis-side failure |
| `RedisModuleVersionError` | `_check_svs_support`, reached from `create()` | ordering `except` clauses, since it subclasses `RedisVLError` and *not* `RedisSearchError` |

The `from_yaml(..., validate_on_load=True)` form matches the example already in the `SearchIndex` docstring.

## Definition of Done

- **Exception docs present and discoverable** — `docs/_build/html/api/exceptions.html` renders, all five classes appear with their docstrings, `RedisModuleVersionError.for_svs_vamana` is documented, and the page is linked from the API index as "Exceptions".
- **`make docs-build` passes** — exit 0. The build reports 11 warnings; all 11 are pre-existing and point at `query.py` / `aggregate.py` docstrings inherited from redis-py, `index.md` heading levels, and the MCP how-to cross-references. None reference `exceptions.rst`.
- **Examples align with current exception classes** — per the table above.

`codespell` is clean on both files. No Python changed, so no behaviour change.

## Notes

- Docs-only, so this needs the **`auto:documentation`** label per CONTRIBUTING. I can't set labels as an outside contributor — could a maintainer add it?
- I deliberately left `uv.lock` out. Running `make docs-build` with a recent `uv` rewrites it to lockfile revision 3 (a ~250 line diff) which has nothing to do with this change.
- If #608 (Sphinx → mkdocs) lands first, this page is a single autodoc file and cheap to port; happy to do that conversion if you'd prefer to wait for it.

---

> [!NOTE]
> **Low Risk**
> Documentation-only change with no runtime or API behavior impact.
> 
> **Overview**
> Adds a dedicated **Exceptions** API page and links it from the API toctree so `redisvl.exceptions` is easy to find.
> 
> The new `exceptions.rst` documents the five-class hierarchy under `RedisVLError`, clarifies that `redis-py` errors are wrapped as `RedisSearchError` with chaining on `__cause__`, and includes a table mapping each exception to typical entry points (`load()`, `query()`, `create()`, etc.). It also provides four `try`/`except` patterns (schema validation on load, query validation, version vs Redis failures, and catching `RedisVLError`) plus autoclass reference blocks matching the style used elsewhere in the API docs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 891bfe9fbf4010153b6e6cb8c9d5f7a766ff6f4b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/api/exceptions.rst
+++ b/docs/api/exceptions.rst
@@ -1,0 +1,200 @@
+***********
+Exceptions
+***********
+
+RedisVL defines its custom exceptions in ``redisvl.exceptions``. Every one of them
+inherits from :class:`RedisVLError`, so catching that single base class is enough to
+handle any error the library raises on its own behalf. Catch the more specific
+subclasses when you want to react differently to, for example, a schema validation
+failure than to a Redis connection problem.
+
+.. code-block:: text
+
+    Exception
+    └── RedisVLError
+        ├── RedisSearchError
+        ├── SchemaValidationError
+        ├── QueryValidationError
+        └── RedisModuleVersionError
+
+.. note::
+
+   Exceptions raised by the underlying ``redis-py`` client, such as
+   ``redis.exceptions.ConnectionError``, are not part of this hierarchy. Where
+   RedisVL performs an index or search operation on your behalf it wraps those
+   errors in a :class:`RedisSearchError` and chains the original exception, so the
+   underlying cause is still available on ``__cause__``.
+
+
+When each error is raised
+=========================
+
+.. list-table::
+   :widths: 28 42 30
+   :header-rows: 1
+
+   * - Exception
+     - Raised when
+     - Typical entry points
+   * - :class:`SchemaValidationError`
+     - An object does not match the index schema. Only raised when the index was
+       created with ``validate_on_load=True``.
+     - ``load()``
+   * - :class:`QueryValidationError`
+     - A query is not valid for the index it targets, for example setting
+       ``ef_runtime`` on a vector field that uses the ``flat`` algorithm.
+     - ``query()``
+   * - :class:`RedisSearchError`
+     - An index or search operation fails, including errors returned by Redis
+       itself.
+     - ``create()``, ``delete()``, ``search()``, ``aggregate()``
+   * - :class:`RedisModuleVersionError`
+     - The connected Redis or Redis Search version does not support a requested
+       feature, such as an ``svs-vamana`` vector field.
+     - ``create()``
+   * - :class:`RedisVLError`
+     - A load operation fails for a reason not covered by a more specific error.
+       Also the base class for everything above.
+     - ``load()``
+
+All of the above apply equally to :class:`~redisvl.index.SearchIndex` and
+:class:`~redisvl.index.AsyncSearchIndex`.
+
+
+Handling errors
+===============
+
+Validating data on load
+-----------------------
+
+Schema validation is off by default. Pass ``validate_on_load=True`` to have RedisVL
+check each object against the index schema before writing it, and raise
+:class:`SchemaValidationError` on the first object that does not match.
+
+.. code-block:: python
+
+    from redisvl.index import SearchIndex
+    from redisvl.exceptions import SchemaValidationError
+
+    index = SearchIndex.from_yaml(
+        "schema.yaml",
+        redis_url="redis://localhost:6379",
+        validate_on_load=True,
+    )
+
+    try:
+        index.load(data)
+    except SchemaValidationError as e:
+        # The message identifies the offending object by its position in the
+        # input and describes which field failed and why.
+        print(f"Invalid record: {e}")
+
+The error message reports the index of the object within the batch you passed, so a
+failure part way through a large load still points at a specific record.
+
+Handling query failures
+-----------------------
+
+:class:`QueryValidationError` signals a query that cannot run against this index. It
+is a programming error rather than a transient one, so it is usually worth failing
+loudly instead of retrying.
+
+.. code-block:: python
+
+    from redisvl.query import VectorQuery
+    from redisvl.exceptions import QueryValidationError
+
+    query = VectorQuery(
+        vector=[0.1, 0.2, 0.3],
+        vector_field_name="embedding",
+        return_fields=["title"],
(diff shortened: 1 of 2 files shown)


docs/api/exceptions.rst
```diff
@@ -0,0 +1,200 @@
...
+        raise
+
+Because RedisVL chains the underlying exception when it wraps one, ``e.__cause__``
+still holds the original ``redis-py`` error where there was one.
+
+
+Exception classes
+=================
+
+.. currentmodule:: redisvl.exceptions
```
vishal-bala: This directive needs to be above the prose that uses it. Sphinx resolves bare `` :class:`RedisVLError` `` against the current module, and until this line there isn't one — so it looks for a top-level `RedisVLError` while autodoc registered `redisvl.exceptions.RedisVLError`. Result: 12 unresolved refs (lines 6, 25, 39, 43, 47, 51, 55, 70, 98, 122).

Moving it to just under the page title fixes all 12 at once, and matches how `schema.rst` orders it. The autoclass blocks below still work, since `currentmodule` applies to the rest of the file.

For comparison once fixed: the `~redisvl.index.SearchIndex` ref on line 60 already links correctly, because it's fully qualified.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,200 @@
+***********
+Exceptions
+***********
+
+RedisVL defines its custom exceptions in ``redisvl.exceptions``. Every one of them
+inherits from :class:`RedisVLError`, so catching that single base class is enough to
```
vishal-bala, line 8: "enough to handle any error the library raises on its own behalf" holds for the operations you document, but argument validation happens in constructors, outside that net. Concretely, with the parameter this page features: `VectorQuery(..., ef_runtime=-1)` raises `ValueError: ef_runtime must be positive`, not a `RedisVLError` — and it fires before the `try` block in the line 107 example opens. A sentence noting that constructor/argument validation raises standard `ValueError` would set expectations well.

docs/api/exceptions.rst
```diff
@@ -0,0 +1,200 @@
...
+failure than to a Redis connection problem.
+
+.. code-block:: text
+
+    Exception
+    └── RedisVLError
+        ├── RedisSearchError
+        ├── SchemaValidationError
+        ├── QueryValidationError
+        └── RedisModuleVersionError
```
vishal-bala, line 22: Minor completeness note: `redisvl/mcp/errors.py:22` defines `RedisVLMCPError(Exception)`, which isn't a `RedisVLError` subclass. Since line 5 reads as exhaustive, either scoping that sentence to the core index/query APIs or adding a pointer would keep this accurate as MCP grows.

vishal-bala (commented):
Hi! Thanks for contributing to RedisVL.

This looks good overall — the page is well-organized, and I verified the exception hierarchy, table, and examples against the source without finding inaccuracies.

One thing needs fixing before merge: `.. currentmodule::` sits below the prose, so 12 of the 14 `:class:` references render as plain text instead of links — including all five names in the "When each error is raised" table. It's a one-line move; details inline.

Worth knowing why your Definition of Done didn't catch this: your build report is accurate (I reproduced 11 warnings, none referencing `exceptions.rst`), but Sphinx only reports unresolved Python cross-references under `-n`. So `make docs-build` passing can't detect this class of problem — `sphinx -b html -n docs docs/_build/html` surfaces it.

Two smaller notes inline, both optional.

TheSaiEaranti:
Thanks for the thorough review, all three points are in the latest commit.

The currentmodule directive now sits right under the page title, and I removed the duplicate above the autoclass section since it applies to the rest of the file. Verified with the sphinx -n build you suggested: zero unresolved reference warnings for exceptions.rst now. Good to know that make docs-build can't catch this class of problem, I'll use -n for docs changes going forward.

Both optional notes are in too. The intro now scopes the base-class claim to the core index and query APIs with a pointer to redisvl.mcp.errors.RedisVLMCPError, and the note block mentions that constructor validation raises standard ValueError before any try block opens, using your ef_runtime example.

vishal-bala (approved):

PR #657, opened 2026-07-31
fix: prevent SearchIndex/AsyncSearchIndex finalizer from leaking every instance

### LINKED ISSUE #656: Memory leak: SearchIndex / AsyncSearchIndex instances are never garbage collected
## Summary

`SearchIndex` and `AsyncSearchIndex` instances are never garbage collected. Every index
ever constructed is retained for the lifetime of the process, so applications that build
an index per request grow unboundedly and are eventually OOM-killed.

The root cause is a `weakref.finalize` misuse in `redisvl/index/index.py`:

```python
# SearchIndex.__init__
if self._owns_redis_client:
    weakref.finalize(self, self.disconnect)          # bound method -> strong ref to self

# AsyncSearchIndex.__init__
if self._owns_redis_client:
    weakref.finalize(self, sync_wrapper(self.disconnect))
```

`self.disconnect` is a **bound method**, so the finalizer callback holds a strong
reference to `self`. Per the [`weakref.finalize` docs](https://docs.python.org/3/library/weakref.html#weakref.finalize),
"the callback must not hold a reference to obj, otherwise obj will never be garbage
collected." `weakref.finalize` stores its callback in a module-level registry that is a
genuine GC root, so this creates a self-sustaining reference the cyclic GC cannot break.
The finalizer never fires, `disconnect()` is never called, and the index — along with its
schema, all field objects, its lock, and its Redis client — is retained forever.

A finalizer intended to release connections is the very thing preventing their release.

## Reproduction (no Redis server required)

```python
import gc
from redisvl.index import SearchIndex, AsyncSearchIndex

D = {
    "fields": [
        {"name": "a", "type": "tag"},
        {"name": "b", "type": "numeric"},
        {"name": "v", "type": "vector",
         "attrs": {"dims": 8, "algorithm": "hnsw",
                   "distance_metric": "cosine", "datatype": "float32"}},
    ],
}

for cls in (SearchIndex, AsyncSearchIndex):
    gc.collect()
    before = sum(1 for o in gc.get_objects() if type(o) is cls)
    for _ in range(300):
        i = cls.from_dict(D)
        del i
    gc.collect()
    after = sum(1 for o in gc.get_objects() if type(o) is cls)
    print(f"{cls.__name__:17s}: {before} -> {after} live after 300 create/delete")
```

**Actual:**
```
SearchIndex      : 0 -> 300 live after 300 create/delete
AsyncSearchIndex : 0 -> 300 live after 300 create/delete
```

**Expected:** both `-> 0`. Every instance survives an explicit `del` plus a full `gc.collect()`.

Confirmed on `main` (0.24.0) and reproduced back to 0.18.2 — the offending line is
unchanged across that range.

## Real-world impact

Reported by a user running a document-ingestion service that constructs a `SearchIndex`
once per upload request: ~33.5 KB leaked per request, perfectly linear over 5,000
requests with no plateau; container RSS climbed 3.82 GB → 5.72 GB over 42 hours, strictly
monotonic, and pods reached their 7 GiB limit and were OOM-killed roughly every 4 days.
The leak is spread thinly across many small retained objects, so it is largely invisible
to a heap profiler and shows up only in RSS — making it hard to attribute.

## Suggested fix

Register a finalizer callback that does **not** reference `self`. Note the client is
created lazily (it is `None` at `__init__` when the index owns it), so the callback must
be able to reach the *current* client at finalization time — capturing the constructor
argument by value is not sufficient. A small mutable holder shared between the instance
and a module-level close helper satisfies both constraints.

## Workaround for users

Reuse a single `SearchIndex` per logical index rather than constructing one per operation.
Note: calling `disconnect()` explicitly does **not** help — the finalizer *registration*
itself is what retains the object, regardless of whether the callback ever runs.

## Environment

| | |
|---|---|
| redisvl | 0.18.2 through 0.24.0 (all affected) |
| redis-py | 7.4.1 |
| Python | 3.12.11 |
| OS | Linux (container) |

Reported by Ivan Devitsyn. Also reproduced with the async client.

nkanu17:
## Background

An external user reported that every `SearchIndex` and `AsyncSearchIndex` instance is retained for the lifetime of the process, and attributed unbounded memory growth in a document-ingestion service (which constructs an index per upload request) to it.

Their production telemetry, quoted here as reported and **not** independently verified: about 33.5 KB leaked per request, container RSS climbing 3.82 GB to 5.72 GB over 42 hours at roughly 21k requests/hour, and pods OOM-killed about every 4 days.

What was verified locally against origin/main (v0.24.0):

- 300 of 300 constructed indexes of each class survive `del` plus a full `gc.collect()`.
- Retained memory over 3000 construct-and-drop cycles, measured as RSS delta after `gc.collect()`: about 5.1 KB per construction with a 3-field schema, and about 29 KB with a 25-field schema (the instance drags its `IndexSchema`, every field object, its lock, and its Redis client along with it). With this fix both fall to noise (roughly 0.1 KB) and zero instances are retained.
- Explicit `disconnect()` does not mitigate the leak, and indexes given a caller-provided client never leaked. Both corollaries follow from the mechanism below.

The bug was introduced by #280 and first shipped in 0.4.0, so every release from 0.4.0 through 0.24.0 is affected (wider than the 0.18.2+ range in the original report).

## Description

`SearchIndex.__init__` registered a finalizer whose callback is a bound method of the instance being finalized:

```python
if self._owns_redis_client:
    weakref.finalize(self, self.disconnect)
```

A bound method holds a strong reference to its instance, and `weakref.finalize` stores its callback in a module-level registry that is a genuine GC root until the finalizer fires. The finalizer can only fire once the instance is unreachable, but the instance can never become unreachable while the registry holds the callback that references it. The finalizer therefore never fires, `disconnect()` never runs, and the index (schema, field objects, locks, and Redis client) is immortal.

`AsyncSearchIndex` had the same defect one level removed: `sync_wrapper(self.disconnect)` returns a closure over the bound method.

This is the documented constraint of [`weakref.finalize`](https://docs.python.org/3/library/weakref.html#weakref.finalize): the callback must not hold a reference to the object being finalized.

Reproduction (no Redis needed): construct and drop 300 indexes, then `gc.collect()`. Before this change, all 300 remain alive for both classes; after, zero. This also explains the two corollaries above: the registry entry retains the instance whether or not `disconnect()` is ever called, and no finalizer is registered at all on the caller-provided-client path.

## Fix

Bind the finalizer to the client, never to `self`. Binding at `__init__` alone is not enough: in the default `from_dict`/`from_yaml` + `redis_url` path the client does not exist yet at construction and is created lazily, so an `__init__`-bound finalizer would capture `None` and silently stop closing real connections.

Instead, the finalizer is registered at every site where an owned client is created or replaced, and detached whenever that client is closed or swapped out:

- Module-level close callbacks (`_close_owned_sync_client`, `_close_owned_async_client`) that take the client as an argument and cannot reference the index. The async callback reuses the existing shutdown-hardened `sync_wrapper`.
- `_register_client_finalizer` / `_detach_client_finalizer` helpers on `BaseSearchIndex`. Registration detaches any previous finalizer first, so a replaced client is never closed later by a stale finalizer.
- Registration sites: `__init__` (covers the `from_existing` constructors, which pass an owned client), lazy creation (sync `_redis_client` property, async `_get_client`), and the deprecated `connect()` / `set_client()` paths.
- Both `disconnect()` implementations detach the finalizer before closing, so an explicitly disconnected index is not double-closed at collection time.

Behavior is unchanged for indexes that do not own their client: no finalizer is registered and the caller's client is never touched.

## Testing

Test-driven: all tests were written first and confirmed failing against unmodified origin/main, then the fix was applied.

- `tests/unit/test_index_gc_finalizer.py` (11 tests, no Redis required): instances of both classes are collectable after `del` + `gc.collect()`; repeated construct/drop cycles do not accumulate; a lazily created owned client is closed exactly once on collection; explicit `disconnect()` closes once and detaches; injected clients are never closed. Before the fix 8 of 11 failed (the 3 passing were the unowned-client cases, consistent with the diagnosis).
- `tests/integration/test_index_gc_finalizer_integration.py` (6 tests, real Redis via testcontainers): an owned client that performed a real round trip is closed exactly once on index collection and its pool sockets are torn down; same for the async client through `sync_wrapper`; injected sync and async clients survive collection and still answer PING; the index-per-request pattern leaves no live instances. Before the fix 4 of 6 failed.
- Full non-API suite locally (`pytest -n auto`, `redis:8.4`): 1976 passed. Six `test_sql_redis_hybrid.py` tests failed locally at the time, and were verified to fail identically on unmodified origin/main, so they were unrelated to this change. Correction added after merge: their cause was a stale local virtualenv holding `sql-redis` 0.6.0 while `pyproject.toml` requires `>=0.7.1`. After `uv sync --all-extras` all six pass locally. They were never broken in the repository or in CI.
- `make format`, `make check-types` (mypy), and pre-commit all pass.

### On the `redis:latest` CI job

One matrix job (`Python 3.13 - redis-py 7.x [redis:latest]`) fails with two `test_hybrid.py` assertions returning zero rows. This looks unrelated to this change:

- `tests/docker-compose.yml` already documents this failure mode: Redis 8.8 changed the RediSearch worker-pool default to a multithreaded background executor, "which surfaces a race where FT.SEARCH can return nil fields for docs that expire mid-query -- the source of flaky redis:latest CI runs".
- Nightly runs on `main` fail the same way on `redis:latest` with a rotating cast of tests (`test_unf_noindex_integration` on Jul 31, `test_query_cosine_distance_un_normalized` on Jul 29), all with the same signature of fewer documents than expected.
- Both failing tests pass locally against `redis:latest` (8.10.0), repeatedly, under `-n auto`.
- Structurally, this change only closes a client the index itself created. `RedisConnectionFactory.get_redis_connection` builds a fresh pool per call via `Redis.from_url` with no caching, so a finalizer firing on one index cannot affect any other client's connections or document visibility.

Two separate observations for whoever picks up the `redis:latest` flakiness, neither of which blocks this PR:

1. The compose mitigation is inert. The official `redis` image does not read `REDIS_ARGS` (that is a redis-stack convention), so `--search-workers 0` is never applied. Confirmed locally: a container started with `REDIS_ARGS=--maxmemory 123mb` reports `maxmemory 0`, and one started with the compose settings reports `search-workers 14`. The flags belong in a compose `command:` entry.
2. That said, the root cause of the `redis:latest` failures is still unattributed. Loading documents and querying immediately, 150 iterations against `redis:latest` (8.10.0) with `search-workers 14` and again with `search-workers 0`, produced no shortfall in either configuration (vector and filter queries both saw all documents every time). So the worker-pool race described in the compose comment was not reproducible here, and fixing item 1 should not be assumed to fix the flake.

## Notes for users on affected versions (0.4.0 through 0.24.0)

- Prefer one long-lived index object per logical index rather than one per operation; this avoids the leak entirely and is also cheaper (no connection churn).
- Calling `disconnect()` does not mitigate the leak on affected versions; only instance reuse does.

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -129,6 +129,37 @@ def _sql_executor_cache_key(sql_redis_options: dict[str, Any]) -> str:
     return json.dumps(sql_redis_options, sort_keys=True, default=repr)
 
 
+def _close_owned_sync_client(client: SyncRedisClient) -> None:
+    """weakref.finalize callback that closes an index-owned sync client.
+
+    Must stay a module-level function taking the client as an argument: a
+    callback that references the index (e.g. a bound method) would be kept
+    alive by the finalizer registry and prevent the index from ever being
+    garbage collected.
+    """
+    try:
+        client.close()
+    except Exception:
+        # Finalizers may run during interpreter shutdown; never propagate.
+        pass
+
+
+def _close_owned_async_client(client: AsyncRedisClient) -> None:
+    """weakref.finalize callback that closes an index-owned async client.
+
+    Same constraint as :func:`_close_owned_sync_client`: must not reference
+    the index instance.
+    """
+
+    async def _aclose() -> None:
+        await client.aclose()
+
+    try:
+        sync_wrapper(_aclose)()
+    except Exception:
+        pass
+
+
 # Default number of documents processed per round-trip in bulk operations.
 DEFAULT_BULK_BATCH_SIZE = 500
 
@@ -346,8 +377,42 @@ class BaseSearchIndex:
 
     schema: IndexSchema
 
+    # Module-level callback (set per subclass) used to close an owned client
+    # when the index is garbage collected. Must never be a bound method.
+    _finalizer_close_client: Callable[[Any], None]
+
+    _client_finalizer: "weakref.finalize | None" = None
+
     def __init__(*args, **kwargs):
         pass
+
+    def _register_client_finalizer(self, client: Any) -> None:
+        """Register a finalizer that closes ``client`` once this index is
+        garbage collected.
+
+        Call this at every point where the index creates or takes over a
+        client it owns. The finalizer callback must only capture the client,
+        never ``self``: ``weakref.finalize`` keeps the callback alive in a
+        module-level registry until it fires, so a callback holding the
+        instance (a bound method, or any closure over ``self``) would keep
+        the instance reachable forever and the finalizer would never run.
+        """
+        self._detach_client_finalizer()
+        if client is not None and getattr(self, "_owns_redis_client", False):
+            self._client_finalizer = weakref.finalize(
+                self, type(self)._finalizer_close_client, client
+            )
+
+    def _detach_client_finalizer(self) -> None:
+        """Cancel the pending client finalizer, if any.
+
+        Called when the owned client is closed or replaced so the finalizer
+        neither double-closes a client nor keeps a replaced client alive.
+        """
+        finalizer = self._client_finalizer
+        if finalizer is not None:
+            finalizer.detach()
+        self._client_finalizer = None
 
     @property
     def _storage(self) -> BaseStorage:
@@ -600,8 +665,13 @@ class SearchIndex(BaseSearchIndex):
 
         self._validated_client = kwargs.pop("_client_validated", False)
         self._owns_redis_client = kwargs.pop("_owns_redis_client", redis_client is None)
-        if self._owns_redis_client:
-            weakref.finalize(self, self.disconnect)
+        self._client_finalizer = None
+        # Close the owned client when this index is garbage collected. When
+        # the client is created lazily, registration happens at creation time
+        # instead (see the _redis_client property).
+        self._register_client_finalizer(redis_client)
+
+    _finalizer_close_client = staticmethod(_close_owned_sync_client)
 
     def disconnect(self):
         """Disconnect from the Redis database."""
@@ -609,6 +679,7 @@ class SearchIndex(BaseSearchIndex):
         if self._owns_redis_client is False:
             logger.info("Index does not own client, not disconnecting")
             return
+        self._detach_client_finalizer()
         if self.__redis_client:
             self.__redis_client.close()
         self.__redis_client = None
@@ -698,6 +769,7 @@ class SearchIndex(BaseSearchIndex):
                         redis_url=self._redis_url,
                         **kwargs,
                     )
+                    self._register_client_finalizer(self.__redis_client)
         if not self._validated_client and self._lib_name:
             # Only set lib name for user-provided clients
             RedisConnectionFactory.validate_sync_redis(
@@ -730,6 +802,7 @@ class SearchIndex(BaseSearchIndex):
         self.__redis_client = RedisConnectionFactory.get_redis_connection(
             redis_url=redis_url, **kwargs
         )
+        self._register_client_finalizer(self.__redis_client)
 
     @deprecated_function("set_client", "Pass connection parameters in __init__.")
     def set_client(self, redis_client: SyncRedisClient, **kwargs):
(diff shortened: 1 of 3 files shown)


nkanu17:
@codex review

abrookins (approved):
Makes sense to me! 👍 Good find.

PR #659, opened 2026-08-03
test: apply Redis server flags via compose command, not REDIS_ARGS

### LINKED ISSUE #658: test: docker-compose passes server flags via REDIS_ARGS, which the official redis image ignores
## Summary

`tests/docker-compose.yml` passes Redis server flags through the `REDIS_ARGS` environment variable. `REDIS_ARGS` is a redis-stack image convention: the redis-stack entrypoint reads it and appends it to the server command line. The official `docker.io/library/redis` image used by this compose file does not read it at all, so **none of these flags are ever applied** to the test container.

```yaml
services:
  redis:
    image: "${REDIS_IMAGE:-redis:8.4}"
    environment:
      # Pin the RediSearch worker pool to 0 (single-threaded, foreground
      # execution). Redis 8.8 changed the default from 0 to a multithreaded
      # background executor, which surfaces a race where FT.SEARCH can return
      # nil fields for docs that expire mid-query -- the source of flaky
      # redis:latest CI runs. WORKERS=0 avoids the race and is already the
      # default on the pinned 8.2/8.4 images, so this is a no-op there.
      - "REDIS_ARGS=--save '' --appendonly no --search-workers ${REDIS_SEARCH_WORKERS:-0}"
```

The affected settings are `--save ''`, `--appendonly no`, and `--search-workers`.

## Evidence

Verified locally against `redis:latest` (Redis 8.10.0):

A container started with the compose file's environment reports the multithreaded default rather than 0:

```
$ docker run -d -e "REDIS_ARGS=--save '' --appendonly no --search-workers 0" redis:latest
$ docker exec <id> redis-cli CONFIG GET search-workers
search-workers
14
```

A distinctive setting confirms the variable is ignored entirely rather than partially parsed:

```
$ docker run -d -e "REDIS_ARGS=--maxmemory 123mb" redis:latest
$ docker exec <id> redis-cli CONFIG GET maxmemory
maxmemory
0
```

(`appendonly no` happens to match the image default, so it is not evidence either way.)

## Impact

This is test infrastructure only. No shipped `redisvl` code is affected, so no released version has a defect from this. The consequence is that the test Redis does not have the configuration the compose file says it has, and the `--search-workers 0` pin added to reduce flaky CI runs on newer Redis images has never taken effect.

## Suggested fix

Pass the flags as real server arguments via a compose `command:` entry instead of an environment variable:

```yaml
    command: ["redis-server", "--save", "", "--appendonly", "no", "--search-workers", "${REDIS_SEARCH_WORKERS:-0}"]
```

Verification worth doing as part of the fix:

- `REDIS_IMAGE=redis:latest docker compose -f tests/docker-compose.yml up -d`, then `redis-cli CONFIG GET search-workers` returns 0.
- Same check against the pinned `redis:8.2` and `redis:8.4` images, where 0 is already the default so the change should be a no-op.
- Confirm `--save ''` still results in no RDB save points, since quoting an empty argument in compose command form differs from shell quoting.

## Note on the flaky CI runs

The compose comment attributes flaky `redis:latest` runs to a Redis 8.8 multithreaded-search race where `FT.SEARCH` can return nil fields for docs expiring mid-query. That attribution should be treated as unconfirmed, and this fix should not be assumed to resolve the flakiness.

Attempting to reproduce it: loading 7 documents and querying immediately with no wait, 150 iterations against `redis:latest` 8.10.0 with `search-workers 14`, and again with `search-workers 0`, produced no shortfall in either configuration. Vector queries and filter queries both saw all 7 documents on every iteration.

The `redis:latest` matrix jobs do fail nondeterministically with a consistent signature of fewer documents than expected, on a rotating cast of tests (`test_unf_noindex_integration`, `test_query_cosine_distance_un_normalized`, and `test_hybrid.py` hybrid queries returning zero rows). Root-causing that is separate work from this issue.

nkanu17:
Fixes #658

## Problem

`tests/docker-compose.yml` passed Redis server flags through the `REDIS_ARGS` environment variable. `REDIS_ARGS` is a redis-stack image convention: its entrypoint reads the variable and appends it to the server command line. The official `docker.io/library/redis` image that this compose file uses never reads it, so `--save ''`, `--appendonly no`, and `--search-workers` were all silently discarded.

The practical effect was that the `--search-workers 0` pin, added to keep newer Redis images on the same query execution model as the pinned ones, had never taken effect on any image.

This is test infrastructure only. No shipped `redisvl` code is affected, so no released version carries a defect from this.

## Fix

The flags are now passed as real server arguments via a compose `command:` entry, with the `${REDIS_SEARCH_WORKERS:-0}` parameterization preserved. The accompanying comment was rewritten to explain the actual mechanism (the official image ignores `REDIS_ARGS`) and no longer asserts the FT.SEARCH nil-field race as established fact, for the reason in the next section.

Exec (list) form turned out to express `--save ''` correctly, so no shell-form fallback was needed. `docker inspect -f '{{json .Args}}'` confirms the argv reaching the server:

```
["redis-server","--save","","--appendonly","no","--search-workers","0"]
```

## Verification

Containers were started from this compose file and inspected directly. On `redis:latest` (8.10.0):

```
CONFIG GET search-workers  ->  0        (was 14 before this change)
CONFIG GET save            ->  ""       (zero-length string, no save points)
CONFIG GET appendonly      ->  no
```

The same three checks pass on `redis:8.4` (8.4.5) and `redis:8.2` (8.2.8), where `search-workers 0` is already the default so the pin is a no-op, and both containers start normally. `--search-workers` is accepted on all three images, so no version gating is needed.

Parameterization still works: `REDIS_SEARCH_WORKERS=4` yields `CONFIG GET search-workers -> 4` on both `redis:latest` and `redis:8.4`, with `save` and `appendonly` unchanged.

Integration tests exercising the testcontainers `DockerCompose` fixture that drives this file:

- `tests/integration/test_search_index.py`: 46 passed
- `tests/integration/test_hybrid.py`: 12 passed, 2 skipped

## Why this matters: it fixes the flaky redis:latest jobs

The comment being replaced attributed flaky `redis:latest` runs to the Redis 8.8 change that enables a multithreaded RediSearch worker pool. That attribution is correct, and because `REDIS_ARGS` was ignored, `search-workers` has been at the image default on every `redis:latest` CI job, so the intended mitigation was never in effect.

Reproduced with 8 concurrent threads each creating an index, loading 3 documents, and querying immediately expecting all 3, 60 iterations per thread (480 total), against `redis:latest` (8.10.0):

| configuration | shortfalls | documents seen at first query |
|---|---|---|
| `search-workers 14` (image default), 1 CPU | 17/480 | 0, 1, or 2 instead of 3 |
| `search-workers 14` (image default), unconstrained CPU | 17/480 | 0, 1, or 2 instead of 3 |
| `search-workers 0` | **0/480** | always 3 |

The partial-result distribution matches the observed CI failures: zero documents (`test_hybrid.py` returning 0 of 7) and 1 or 2 of 3 (`test_unf_noindex_integration`).

The writes are not lost. Across 30 shortfall events, all 3 hashes were already in the keyspace every time, and the index caught up in every case within 0.8 to 2.6 ms (median 1.6 ms). So `FT.SEARCH` briefly loses read-your-writes consistency under concurrent write and query load, and `search-workers 0` eliminates it.

A single-threaded probe against an idle server shows no shortfall at all, which is why an earlier attempt of mine failed to reproduce this. Concurrency is required to surface it.

This is server behavior and also affects library users who load and then immediately query under concurrency. That is tracked separately for documentation and is not addressed by this PR.

### DIFF
--- a/tests/docker-compose.yml
+++ b/tests/docker-compose.yml
@@ -3,14 +3,26 @@ services:
     image: "${REDIS_IMAGE:-redis:8.4}"
     ports:
       - "6379"
-    environment:
-      # Pin the RediSearch worker pool to 0 (single-threaded, foreground
-      # execution). Redis 8.8 changed the default from 0 to a multithreaded
-      # background executor, which surfaces a race where FT.SEARCH can return
-      # nil fields for docs that expire mid-query -- the source of flaky
-      # redis:latest CI runs. WORKERS=0 avoids the race and is already the
-      # default on the pinned 8.2/8.4 images, so this is a no-op there.
-      - "REDIS_ARGS=--save '' --appendonly no --search-workers ${REDIS_SEARCH_WORKERS:-0}"
+    # Server flags have to be passed as real arguments. REDIS_ARGS is a
+    # redis-stack image convention: the official docker.io/library/redis image
+    # used here never reads it, so anything set that way is silently ignored.
+    #
+    # --save "" disables RDB save points and --appendonly no disables the AOF;
+    # test data is disposable.
+    #
+    # --search-workers pins the RediSearch worker pool to 0 (single-threaded,
+    # foreground execution). Redis 8.8 changed the default from 0 to a
+    # multithreaded background executor, so pinning it keeps newer images on the
+    # same query execution model as the pinned 8.2/8.4 images, where 0 is
+    # already the default (a no-op there). Override with REDIS_SEARCH_WORKERS.
+    command:
+      - "redis-server"
+      - "--save"
+      - ""
+      - "--appendonly"
+      - "no"
+      - "--search-workers"
+      - "${REDIS_SEARCH_WORKERS:-0}"
     deploy:
       replicas: 1
       restart_policy:


nkanu17:
Closing as a duplicate of #653, which already contains this fix and predates it by over a week.

The functional change is identical: same exec-form `command:` with the same seven arguments and the same `${REDIS_SEARCH_WORKERS:-0}` parameterization. A diff of the two branches on `tests/docker-compose.yml` shows only comment wording, and #653's comment is the better one, noting the 8.8.0 default of 12 and documenting the override for reproducing the race locally.

My verification is recorded on #653 rather than duplicated here: `search-workers` 0 applied on redis:latest (8.10.0), `CONFIG GET save` returning an empty string so exec form does express `--save """ correctly, `appendonly no`, all confirmed on redis:8.4 and redis:8.2 too, and `REDIS_SEARCH_WORKERS=4` overriding as expected.

Issue #658 is likewise a duplicate of a problem already diagnosed in #653 and can be closed with it.

PR #661, opened 2026-08-03
fix: do not close a caller-provided client passed to set_client()

### LINKED ISSUE #660: set_client() keeps client ownership, so the index closes a caller-provided client
## Summary

An index created with `redis_url` owns the client it lazily creates, so `_owns_redis_client` is `True`. The deprecated `set_client()` does not reset that flag, so after a caller swaps in their own client the index still believes it owns it and will close it. `__init__(redis_client=...)` already gets this right and marks such a client as not owned.

Before 0.25.0 this was latent, because the client finalizer never fired (see #657). Now that the finalizer works, the index closes the caller's client when it is garbage collected.

Affects both `SearchIndex` and `AsyncSearchIndex`, on 0.25.0.

## Reproduction

```python
import gc
from unittest import mock

SCHEMA = {
    "fields": [{"name": "a", "type": "tag"}],
}

caller_client = mock.MagicMock()
index = SearchIndex.from_dict(SCHEMA, redis_url="redis://localhost:6379")
with mock.patch("redisvl.index.index.RedisConnectionFactory.validate_sync_redis"):
    index.set_client(caller_client)

print("owns:", index._owns_redis_client)      # True, should be False
del index
gc.collect()
print("caller client closed:", caller_client.close.called)   # True, should be False
```

Observed on 0.25.0:

```
owns: True
caller client closed: True
```

The async path behaves the same, awaiting `aclose()` on the caller's client. Calling `disconnect()` explicitly closes it too, for the same reason.

## Impact

Low severity but real. redis-py clients recover from `close()` and `aclose()` by reconnecting on next use, so the practical effect is unexpected connection churn on a client the caller still owns rather than a permanently broken client. Verified: `ping()` returns `True` after both `close()` and `aclose()` for sync and async clients.

The exposure is further limited because `set_client()` is deprecated.

## Additional problem in the same method

The sync `set_client()` also abandons the client the index created for itself, without closing it. It overwrites `__redis_client` and (since 0.25.0) detaches that client's finalizer, so nothing ever closes it. The async `set_client()` does not have this problem because it awaits `disconnect()` before swapping.

## Suggested fix

`set_client()` should mark the client as not owned, since a caller-provided client is by definition not the index's to close, and it should release the previously owned client first.

The deprecated async `connect()` needs care here: it creates its own client and then delegates to `set_client()`, so a naive ownership flip would leave a client the index created with nobody to close it. That path must keep ownership. Routing both through a small internal helper that takes ownership as a parameter handles this.

Worth covering with tests on all three entry points: constructor injection (already correct), `set_client()` (must not own), and `connect()` (must own).

nkanu17:
Fixes #660

Client ownership fixes for the deprecated `connect()` and `set_client()` paths. Based on `main`, no longer stacked.

## The bug (#660)

`__init__(redis_client=...)` already treats a caller-provided client as not owned and never closes it. The deprecated `set_client()` did not follow that rule: an index created with `redis_url` owns the client it lazily creates, so `_owns_redis_client` stays `True`, and `set_client()` never reset it. The index went on believing it owned a client it had never created.

Latent before 0.25.0 because the finalizer never fired (#657). Now that it works, the index closes the caller's client on garbage collection, and `disconnect()` closes it too. Verified on 0.25.0 before this change, both classes:

```
_owns_redis_client after __init__(redis_url=...):  True
_owns_redis_client after set_client(caller):       True
caller's client closed on GC (sync):               True
caller's client aclosed on GC (async):             True
```

Severity is low: redis-py clients recover from `close()` and `aclose()` by reconnecting, so the effect is unexpected connection churn rather than a broken client, and `set_client()` is deprecated. Confirmed directly, `ping()` returns `True` after both.

The trap was the deprecated async `connect()`, which creates its own client and then delegated to `set_client()`. A plain ownership flip would have left a client the index created with nobody to close it. Both now route through `_swap_client(client, *, owns: bool)`, so `connect()` keeps ownership and `set_client()` gives it up.

Ownership is now consistent across every entry point:

| Entry point | Who created the client | Index owns it |
|---|---|---|
| `__init__(redis_client=...)` | caller | no (unchanged) |
| `__init__(redis_url=...)` then lazy creation | index | yes (unchanged) |
| `connect()` (deprecated) | index | yes (**fixed**, was whatever it was before) |
| `set_client()` (deprecated) | caller | no (**fixed**, was yes) |

## Two related bugs found while fixing it

**Sync `set_client()` abandoned the index's own client.** It overwrote the client, and since 0.25.0 the overwrite also detached that client's finalizer, so nothing closed it. It now releases the previous client first.

**`connect()` leaked the client it replaced.** Same mechanism: registering a finalizer for the new client detaches the old client's, so an owned client being replaced had nothing left to close it. Reachable by calling `connect()` twice, or after a client was lazily created from `redis_url`. Also predates this PR, fixed here since this PR is already reworking ownership in these methods. The async path avoided it by going through `_swap_client()`, which disconnects an owned client first.

## Bot review findings addressed

- **Bugbot:** sync `connect()` never set `_owns_redis_client`, so after `set_client()` cleared ownership, a client `connect()` had just created was never closed. Fixed, with tests, for both classes.
- **Copilot:** `_swap_client()` assigned `_redis_client` under the lock but flipped `_owns_redis_client` outside it. Both fields plus the finalizer registration now happen together inside the lock, and the sync paths do the same under their threading lock. In a single event loop there is no await point between those statements, so the window was effectively nil, but relying on that is fragile to future edits.
- **Copilot:** `set_client()` called `disconnect()` unconditionally, which for a non-owned client only logged that it was not disconnecting. Now guarded; the SQL schema cache is still invalidated explicitly, so nothing is skipped.
- **Copilot:** the `connect()` leak above.

## One existing test changed

`test_search_index_set_client` asserted `.client is None` after `disconnect()`. That was only reachable through the bug: for a client the index does not own, `disconnect()` has always left the client in place rather than closing or clearing it, exactly as it does for constructor-injected clients. It now asserts the corrected semantics, including that the caller's client still answers `PING`. That last part matters here because the test passes a **sync** client, which `_validate_client` converts into an async wrapper sharing the caller's pool; closing that wrapper would tear down connections the caller still needs.

## Testing

Test-driven throughout; every test below was confirmed failing first.

- `tests/unit/test_index_client_ownership.py`, 16 tests: `set_client()` marks the client unowned and never closes it, on GC or via `disconnect()`, both classes; `connect()` keeps ownership and still closes its own client; `connect()` takes ownership back from an unowned state; `connect()` releases the client it replaces, including after lazy creation; previous owned clients are released on swap; indexes stay garbage collectable throughout.
- `tests/integration/test_index_client_ownership_integration.py`, 7 tests against real Redis: a caller's client still answers `PING` after the index is collected and after `disconnect()`, both classes; the index's own client has its pool sockets torn down when replaced; a client created by `connect()` is closed exactly once on collection; plus the constructor-injection baseline.

Full suite: **2005 passed**, 143 skipped, 2 xfailed, no failures. `make format`, mypy, and pre-commit all clean.

---

> [!NOTE]
> **Medium Risk**
> Changes connection lifecycle and ownership for sync/async index client swaps (including locking and finalizers); impact is limited to deprecated `connect`/`set_client` paths but mistakes could leak connections or close caller-owned clients.
> 
> **Overview**
> Fixes **#660** by aligning Redis client **ownership** across deprecated `connect()` / `set_client()` with constructor injection: caller-supplied clients are never closed on `disconnect()` or GC, while clients the index creates remain owned and are released on swap.
> 
> **Sync `SearchIndex`:** `connect()` and `set_client()` now swap clients under `_lock` via `_release_owned_client_locked()`—closing a previously **owned** client before installing the replacement, setting `_owns_redis_client` appropriately (`True` for `connect()`, `False` for `set_client()`), clearing `_validated_client`, and registering or detaching finalizers. No-op when the same client is reinstalled.
> 
> **Async `AsyncSearchIndex`:** Both paths route through new `_swap_client(client, owns=…)` (`connect()` → `owns=True`, `set_client()` → `owns=False`), replacing the old pattern that called `disconnect()` then assigned the client. Sync clients passed to the async index are converted to a **new** async client that the index owns.
> 
> **Tests:** Added `tests/unit/test_index_client_ownership.py` and `tests/integration/test_index_client_ownership_integration.py`; updated async `set_client` integration expectations (caller sync client still works after disconnect).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f42fa1a952ad1981f48705a9c0823046c9ac455e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -955,8 +955,14 @@ class SearchIndex(BaseSearchIndex):
         # create a client in between, whose finalizer would then be detached by
         # the swap below without anything closing it.
         with self._lock:
-            if client is not self.__redis_client:
-                self._release_owned_client_locked()
+            if client is self.__redis_client:
+                # The factory handed back the client already installed, so its
+                # provenance has not changed and ownership stays as it is.
+                # Claiming it here would let this index close a caller-provided
+                # client. Only reachable with a factory that reuses clients;
+                # Redis.from_url always builds a fresh one.
+                return
+            self._release_owned_client_locked()
             self.__redis_client = client
             # This index created the client, so it owns and must close it. The
             # index may have been holding a caller-provided client until now
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -969,8 +969,14 @@ class SearchIndex(BaseSearchIndex):
         """
         RedisConnectionFactory.validate_sync_redis(redis_client)
         self.invalidate_sql_schema_cache()
+        # Release the client this index created for itself, if any, before
+        # taking on the caller's client.
+        self.disconnect()
         self.__redis_client = redis_client
-        self._register_client_finalizer(redis_client)
+        # The caller owns the client they passed in, so this index must never
+        # close it. This matches __init__(redis_client=...) semantics.
+        self._owns_redis_client = False
+        self._detach_client_finalizer()
         return self
 
     def _check_svs_support(self) -> None:
@@ -2171,7 +2177,8 @@ class AsyncSearchIndex(BaseSearchIndex):
         client = await RedisConnectionFactory._get_aredis_connection(
             redis_url=redis_url, **kwargs
         )
-        await self.set_client(client)
+        # This index created the client, so it owns and must close it.
+        await self._swap_client(client, owns=True)
 
     @deprecated_function("set_client", "Pass connection parameters in __init__.")
     async def set_client(self, redis_client: AsyncRedisClient | SyncRedisClient):
@@ -2179,12 +2186,31 @@ class AsyncSearchIndex(BaseSearchIndex):
         [DEPRECATED] Manually set the Redis client to use with the search index.
         This method is deprecated; please provide connection parameters in __init__.
         """
-        redis_client = await self._validate_client(redis_client)
+        # The caller owns the client they passed in, so this index must never
+        # close it. This matches __init__(redis_client=...) semantics.
+        return await self._swap_client(redis_client, owns=False)
+
+    async def _swap_client(
+        self, redis_client: AsyncRedisClient | SyncRedisClient, *, owns: bool
+    ):
+        """Replace the active client, releasing the previous one if owned.
+
+        Args:
+            redis_client: The client to start using.
+            owns: Whether this index owns the new client and is therefore
+                responsible for closing it. Callers passing their own client
+                must use False so it is never closed on their behalf.
+        """
+        validated_client = await self._validate_client(redis_client)
         self.invalidate_sql_schema_cache()
+        # Release the client this index created for itself, if any.
         await self.disconnect()
         async with self._lock:
-            self._redis_client = redis_client
-        self._register_client_finalizer(redis_client)
+            self._redis_client = validated_client
+        self._owns_redis_client = owns
+        # No-op when owns is False; also detaches any finalizer left over from
+        # a previously owned client.
+        self._register_client_finalizer(validated_client)
         return self
 
     async def _get_client(self) -> AsyncRedisClient:
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -987,15 +987,20 @@ class SearchIndex(BaseSearchIndex):
         # Release the previous client and install the caller's under a single
         # lock hold, for the reason described in connect().
         with self._lock:
-            # Skip the release when the caller hands back the client already
-            # installed: closing it would leave this index holding a client it
-            # had just closed.
-            if redis_client is not self.__redis_client:
-                self._release_owned_client_locked()
-            self.__redis_client = redis_client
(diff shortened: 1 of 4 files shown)


nkanu17 (commented):
Remove stack

vishal-bala (changes requested):
This is a good catch, but my main gripe with it would be that it's adding some complexity and checks to a set of operations that have already been marked for deprecation since February 2025.

I'd say that my preferred path for this PR would be to actually complete that deprecation and remove those methods (in favour of passing connection parameters via `__init__`) so that we can reduce the amount of risky code and tech debt in the index.

abrookins:
I'd support that @vishal-bala !

PR #149, opened 2024-05-13
Reranker implementation using Cross Encoders (HuggingFace / SentenceT…

bsbodden:
…ransformers)

Inspired by [LangChain's Cross Encoder Reranker](https://python.langchain.com/v0.1/docs/integrations/document_transformers/cross_encoder_reranker/)

Issue #90, opened 2023-12-20
Ability to choose the RediSearch Dialect


pierrelambert:
For instance Dialect 2 does not permit to retrieve multi-value data stored in json 
Dialect 3 and 4 do permit.

tylerhutcherson:
working on this. dialect will be 2 by default, but configurable on the query class :)

tylerhutcherson:
Addressed in https://github.com/RedisVentures/redisvl/pull/95

PR #674, opened 2026-08-07
fix(deps): require sentence-transformers >=5.2.0 to unblock a patched transformers

vishal-bala:
Three open Dependabot alerts are all `transformers`, and all of them reach us through one optional extra. Nothing under `redisvl/` imports `transformers` directly.

| Alert | Severity | CVE | Patched in |
|---|---|---|---|
| #151 | high (8.0) | CVE-2026-5241 | transformers 5.5.0 |
| #149 | high (7.8) | CVE-2026-4372 | transformers 5.3.0 |
| #90 | medium (6.5) | CVE-2026-1839 | transformers 5.0.0rc3 |

All three are upstream `transformers` advisories, already public and fixed upstream; see the linked GitHub advisories for details. Two of them concern model loading, which is the code path this extra uses, so we are treating them as applying here rather than assuming otherwise. The third additionally requires `torch<2.6` (we resolve 2.13.0) and a `transformers` component redisvl does not use; it is closed here as a side effect.

Please report any security concern about redisvl through [GitHub's private vulnerability reporting](https://github.com/redis/redis-vl-python/security/advisories/new) rather than a public issue or PR comment.

## The blocker was our own constraint, not transformers

Every `sentence-transformers` release below 5.2.0 declares `transformers<5.0.0,>=4.41.0`. That ceiling sits below *every* patched release, so no amount of upgrading `transformers` could help while we capped the extra at `<4`. I checked each candidate on PyPI: 3.4.x, 4.x, 5.0.0 and 5.1.x all carry the cap, and **5.2.0 is the first release that widens it to `transformers<6.0.0`**. There is no 4.x backport to fall back on either — the transformers 4.x line ended at 4.57.6.

So the floor moves rather than the range widening:

```diff
-sentence-transformers = ["sentence-transformers>=3.4.0,<4"]
+sentence-transformers = ["sentence-transformers>=5.2.0,<6"]
```

## ⚠️ Breaking: the `sentence-transformers` and `all` extras drop st 3.x–5.1

If you pin `sentence-transformers`, move to `>=5.2.0,<6`.

Installing `redisvl[sentence-transformers]` against an older pin fails with a `ResolutionImpossible` that names **`redisvl`**, not `sentence-transformers`, so the real cause is not obvious from the error. There is also a quieter failure mode: `pip install -U redisvl` *without* re-specifying the extra never evaluates the extra's constraint at all, so an old `sentence-transformers` simply stays put with no warning.

We considered keeping a range that still admitted st 3.x/4.x and rejected it deliberately. That range would advertise a dependency combination that can never resolve a patched `transformers` — the alerts would stay open for precisely the users who took us up on the wider range.

## ⚠️ If you constrain `transformers`, require `>=5.5.0`

This PR deliberately does **not** add a `transformers` floor: redisvl does not import `transformers`, and we would rather not pin a transitive dependency of an optional extra.

That means the requirement lands on you if your project constrains `transformers` at all, or already has an older version installed — pip will not necessarily upgrade an existing version that still satisfies the range. Our lockfile resolves 5.14.1, which covers this repository's own environment, but it does not govern yours.

If you use `redisvl[sentence-transformers]` or `redisvl[all]`, please make sure your resolved `transformers` is `>=5.5.0`:

```bash
pip install -U "transformers>=5.5.0"
python -c "import transformers; print(transformers.__version__)"
```

## Compatibility

redisvl's use of the library is limited to `SentenceTransformer(model, **kwargs)`, `.encode()`, `CrossEncoder(model, **kwargs)` and `.predict()` — all stable across 3 → 5 — so no library code changed.

Two things to know if you forward extra kwargs through our constructors:

- sentence-transformers 5.x renamed `automodel_args` / `tokenizer_args` to `model_kwargs` / `processor_kwargs`. We pass `**kwargs` verbatim, so the old names now raise `TypeError`.
- This pulls `huggingface-hub` across a major version, 0.36.2 → 1.27.0 (transformers 5.x requires `>=1.5.0,<2`). redisvl has no `huggingface_hub` imports and is unaffected, but if your own code calls `hf_hub_download`, `snapshot_download` or `InferenceClient`, read the hub 1.0 notes.

**Reranker scores are unchanged.** `CrossEncoder.predict()` activation handling was rewritten in st 4.0, so this was the most likely place for a silent behavioural shift. I compared st 3.4.1 against 5.7.0 directly for both models the suite uses: `ms-marco-MiniLM-L-6-v2` resolves to `Identity()` under both and `stsb-distilroberta-base` to `Sigmoid()` under both, producing byte-identical scores. No user-visible change in `HFCrossEncoderReranker` score scale.

## Also here: Python 3.14 HuggingFace tests now actually run

`SKIP_HF` existed because `transformers` had no Python 3.14 support. It does now, and torch 2.13.0 ships cp314 wheels. Since `make test` applies no `-m` filter, the 15 Python-3.14 matrix jobs were silently skipping 119 tests; they now execute them.

Removing that scaffolding surfaced a latent bug worth calling out. Two modules rebound `requires_hf = pytest.mark.skipif(SKIP_HF, ...)` at module scope, so all 14 `@requires_hf` sites carried a *skipif* mark rather than the registered marker. `pytest -m requires_hf` selected nothing, and the `requires_hf and SKIP_HF` branch in `pytest_collection_modifyitems` was unreachable. Both aliases are gone and every site now uses `@pytest.mark.requires_hf` directly, so the marker works for the first time — and `grep pytest.mark.requires_hf` finds every real use site, which is the discoverability property whose absence hid the bug.

The marker is kept on purpose as a capability label, since these tests download model weights. The `HFTextVectorizer` entries in the three `test_vectorizers.py` param lists are marked too, closing a hole where `--run-api-tests -m "not requires_hf"` still pulled weights for 7 tests.

Not fixed here: several tests construct `SemanticRouter` / `SemanticMessageHistory` without an explicit vectorizer and so default to `HFTextVectorizer` while remaining unmarked. `-m "not requires_hf"` is therefore not yet a fully offline-safe run.

## Verification

Resolution: sentence-transformers 3.4.1 → 5.7.0, transformers 4.57.6 → **5.14.1**, huggingface-hub 0.36.2 → 1.27.0, torch 2.13.0 unchanged. `sentence-transformers` is the only lock entry that depends on `transformers`, and there is no python-conditional fallback leg, so no valid re-resolve drags it back below 5.5.0.

One trap worth recording for the next dependency-driven security fix: `uv lock` alone left `transformers` at the vulnerable 4.57.6, because uv preserves an existing pin that still satisfies the new range. It took `uv lock --upgrade-package transformers` to move it. Without that, this PR would have looked correct and closed nothing.

| Gate | Python 3.13 | Python 3.14 |
|---|---|---|
| `-m requires_hf`, six HF modules | 118 passed, 1 skipped | 118 passed, 1 skipped |
| `make test` | 2016 passed, 143 skipped, 2 xfailed | identical |

`make check-types` clean across 119 files; `pre-commit run --all-files` passes.

Threshold-sensitive tests (`SemanticRouter` at 0.3/0.2, `SemanticCache` at 0.2, `SemanticMessageHistory` at 0.5) were the main regression risk and all pass. They are structurally insulated: the extensions hardcode `flat`/`cosine`, `datatype` comes from redisvl's own `"float32"` default rather than from the model, and `.tolist()` → `array_to_buffer` recasts to the schema dtype — so an `encode()` change could only perturb values, never buffer layout. Worth noting for the future that `test_semantic_add_and_get_relevant` has only ~0.02 of headroom above its threshold and is the first assertion that would break under any embedding-model change.

## Release notes

Please label this **`auto:minor`**, not `auto:dependencies`. In `.autorc`, `auto:dependencies` is `"releaseType": "none"` — it would publish a downstream-breaking constraint change at an unchanged version, filed under "Dependency Updates" where an affected user would skim past it. The release notes should carry both warnings above: the new `sentence-transformers` floor, and that consumers pinning `transformers` must bump to `>=5.5.0` themselves.

## Follow-ups (not in this PR)

- `all` duplicates every dependency literally instead of using self-referential extras, which is why this change had to edit the same constraint twice. It has already drifted: `all` omits `mcp`'s `fastmcp` and `pydantic-settings`, so `pip install redisvl[all]` and CI's `uv sync --all-extras` are not the same set.
- `.github/actions/run-service-tests/action.yml` writes `'{"token":"$HF_TOKEN"}'` in single quotes, storing the literal string rather than the token — and huggingface-hub 1.x no longer reads `~/.huggingface/token` regardless. Dead step.
- Mark the remaining model-downloading tests so `-m "not requires_hf"` is genuinely offline-safe.
- Consider stating the new floor in `docs/user_guide/installation.md`, given how unhelpful the resolver error is.

---

> [!NOTE]
> **Medium Risk**
> Breaking optional-extra constraint change that pulls major upgrades of sentence-transformers, transformers, and huggingface-hub. No redisvl library code changed, but consumers pinning older versions will fail to resolve.
> 
> **Overview**
> Raises the optional `sentence-transformers` / `all` extras from `>=3.4.0,<4` to **`>=5.2.0,<6`** so a patched `transformers` can resolve. Older st releases pin `transformers<5.0.0`, which blocked every fix for the open Dependabot alerts (CVE-2026-5241, CVE-2026-4372, CVE-2026-1839).
> 
> **Breaking for consumers of those extras:** st 3.x–5.1 is dropped. redisvl library code is unchanged (only stable `SentenceTransformer` / `CrossEncoder` APIs are used). Callers who pin `transformers` themselves still need `>=5.5.0`.
> 
> Also removes the Python 3.14 `SKIP_HF` skip path and fixes `@pytest.mark.requires_hf` so HF tests actually run and are selectable on 3.14.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1d541f1db3ad175dbc729e93f781e8e61bcbab0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,7 +48,8 @@ mistralai = ["mistralai>=1.0.0,<2"]
 openai = ["openai>=1.1.0"]
 cohere = ["cohere>=4.44"]
 voyageai = ["voyageai>=0.2.2"]
-sentence-transformers = ["sentence-transformers>=3.4.0,<4"]
+# Floor is a security constraint, not an API one: below 5.2.0 pins transformers<5.0.0.
+sentence-transformers = ["sentence-transformers>=5.2.0,<6"]
 langcache = ["langcache>=0.11.0"]
 # Legacy Google backend for the deprecated VertexAIVectorizer. The <2.0.0 cap is
 # intentional: Google's model-garden modules (vertexai.language_models /
@@ -78,7 +79,7 @@ all = [
     "openai>=1.1.0",
     "cohere>=4.44",
     "voyageai>=0.2.2",
-    "sentence-transformers>=3.4.0,<4",
+    "sentence-transformers>=5.2.0,<6",
     "langcache>=0.11.0",
     "google-cloud-aiplatform>=1.26,<2.0.0",
     "protobuf>=5.28.0,<6.0.0",
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,7 +3,6 @@ import logging
 import os
 import re
 import subprocess
-import sys
 from datetime import datetime, timezone
 
 import pytest
@@ -12,12 +11,7 @@ from testcontainers.compose import DockerCompose
 from redisvl.index.index import AsyncSearchIndex, SearchIndex
 from redisvl.redis.connection import RedisConnectionFactory, is_version_gte
 from redisvl.redis.utils import array_to_buffer
-
-# Check if we're on Python 3.14+ where sentence-transformers may not work
-SKIP_HF = sys.version_info >= (3, 14)
-
-if not SKIP_HF:
-    from redisvl.utils.vectorize import HFTextVectorizer
+from redisvl.utils.vectorize import HFTextVectorizer
 
 logger = logging.getLogger(__name__)
 
@@ -225,8 +219,6 @@ def cluster_client(redis_cluster_url):
 
 @pytest.fixture(scope="session")
 def hf_vectorizer():
-    if SKIP_HF:
-        pytest.skip("HFTextVectorizer not supported on Python 3.14+")
     return HFTextVectorizer(
         model="sentence-transformers/all-mpnet-base-v2",
         token=os.getenv("HF_TOKEN"),
@@ -236,15 +228,11 @@ def hf_vectorizer():
 
 @pytest.fixture(scope="session")
 def hf_vectorizer_float16():
-    if SKIP_HF:
-        pytest.skip("HFTextVectorizer not supported on Python 3.14+")
     return HFTextVectorizer(dtype="float16")
 
 
 @pytest.fixture(scope="session")
 def hf_vectorizer_with_model():
-    if SKIP_HF:
-        pytest.skip("HFTextVectorizer not supported on Python 3.14+")
     return HFTextVectorizer("sentence-transformers/all-mpnet-base-v2")
 
 
@@ -465,9 +453,6 @@ def pytest_collection_modifyitems(
     skip_cluster = pytest.mark.skip(
         reason="Skipping test because Redis cluster is not available. Use --run-cluster-tests to run these tests."
     )
-    skip_hf = pytest.mark.skip(
-        reason="Skipping test because sentence-transformers is not supported on Python 3.14+"
-    )
 
     # Apply skip markers independently based on flags
     for item in items:
@@ -475,8 +460,6 @@ def pytest_collection_modifyitems(
             item.add_marker(skip_api)
         if item.get_closest_marker("requires_cluster") and not run_cluster_tests:
             item.add_marker(skip_cluster)
-        if item.get_closest_marker("requires_hf") and SKIP_HF:
-            item.add_marker(skip_hf)
 
 
 @pytest.fixture
--- a/tests/integration/test_cross_encoder_reranker.py
+++ b/tests/integration/test_cross_encoder_reranker.py
@@ -1,13 +1,8 @@
 import pytest
 
-from tests.conftest import SKIP_HF
+from redisvl.utils.rerank.hf_cross_encoder import HFCrossEncoderReranker
 
-if not SKIP_HF:
-    from redisvl.utils.rerank.hf_cross_encoder import HFCrossEncoderReranker
-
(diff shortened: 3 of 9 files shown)


nkanu17 (approved):
lgtm!

PR #677, opened 2026-08-07
feat(mcp): register and execute custom tool profiles

vishal-bala:
**Stack position: 3 of 3.** Base is #676. This completes the feature — the point at which a `custom_tools:` entry becomes a real tool.

## How locked arguments are actually unreachable

`register_profile_tool` builds each profile's wrapper signature dynamically and hands it to FastMCP, which derives the advertised input schema from that signature and marks it `additionalProperties: false`. That is what makes a locked or hidden argument genuinely unreachable rather than merely undocumented — the model cannot name an argument the schema does not contain.

I verified empirically that FastMCP derives its schema from a dynamic `__signature__`, since the whole design rests on it. The wrapper still re-checks exposure per call rather than trusting the schema alone.

The `filter` annotation is an object type, never a string, so a raw filter string is refused by the advertised schema; the wrapper refuses one too, because the schema is the client's contract and the wrapper is the server's. A `limit` cap is published as `Field(le=cap)` so the ceiling is visible to the model rather than only enforced on rejection.

## Startup validation

Catches what config load could not, because it needs the inspected schema: a locked projection or filter naming a field the bound index does not have, a locked `exists` on a field without `INDEXMISSING`, or one on a vector field. It runs before registration so a bad profile fails startup instead of leaving a half-registered tool set behind.

## Two operational hazards that would otherwise be silent

Tools register once per process, but a profile bakes its locked filter, projection, and signature in at registration time. A restart that reloads a *changed* config would therefore keep enforcing the old profiles. The dangerous direction is an operator tightening a lock and believing the restart applied it, so the server fingerprints the config its tools were built from and warns when that no longer matches.

The empty-surface warning from #668 now also names `custom_tools` as a possible cause.

## Also included

Integration coverage against real Redis, the concept and how-to documentation, and unit tests for registration, execution, description building, and per-binding lock isolation.

## Verification

- MCP unit tests: 356 passing
- `make check-types`: clean

## After this merges

The integration branch holds #675 + #676 + this, and squash-merges to `main` as one "custom tool profiles" commit. Phase 2 (auth-claim tenant injection) and v1.1 (code tools) are separate follow-ups and not in this stack.

One thing recorded for phase 2: `TokenEscaper` does not escape `|`, so a scalar claim like `acme|evil` would render `@tenant_id:{acme|evil}` — a cross-tenant OR. Harmless here because the value is ANDed under the lock, but claim injection must validate claim *characters*, not just type.

---

> [!NOTE]
> **Medium Risk**
> Profiles enforce data scoping via locked filters and schema narrowing, but incorrect locks or misunderstood AND-composition could surprise operators; tools still register once per process so config changes need a full restart.
> 
> **Overview**
> Adds **custom tool profiles**: YAML `custom_tools` entries that publish curated `search-records` wrappers under custom names, with **locked** filters/projections and **exposed** params driving a dynamic FastMCP signature (`additionalProperties: false`).
> 
> **Runtime:** New `profiles.py` registers each profile at startup (pinned index, parsed locked filter, limit caps, object-only caller filters), delegates to `search_records` with `locked_filter` / `limit_cap`, and validates locked fields against the inspected index schema before registration. **Server** wires `register_profile_tools`, extends the tool-surface fingerprint/warnings for profile config drift, and documents the feature in concepts and how-to guides.
> 
> **Tests:** Integration tests against Redis plus broad unit coverage for filter AND-scoping, schema advertisement, limits, auth, and multi-profile isolation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7cac8c7fc2be6fa11a9c77bba7b9085190c42925. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/concepts/mcp.md
+++ b/docs/concepts/mcp.md
@@ -86,7 +86,7 @@ MCP-reserved score metadata field names for the configured search mode.
 
 ## Read-Only and Read-Write Modes
 
-RedisVL MCP registers `search-records` and `list-indexes` by default (see [Tool Surface](#tool-surface) for turning a built-in off deliberately).
+RedisVL MCP registers `search-records` and `list-indexes` by default (see [Custom Tool Profiles](#custom-tool-profiles) for turning a built-in off deliberately).
 
 Write availability is enforced at two levels:
 
@@ -105,13 +105,13 @@ For configuration and the gateway boundary, see {doc}`/user_guide/how_to_guides/
 
 ## Tool Surface
 
-RedisVL MCP exposes up to three built-in tools:
+RedisVL MCP exposes up to three built-in tools, plus any configured [custom tool profiles](#custom-tool-profiles):
 
 - `list-indexes` enumerates the configured logical indexes for discovery
 - `search-records` searches a selected index using that index's server-owned search mode
 - `upsert-records` validates and upserts records into a selected writable index, embedding them only when that capability is configured
 
-Any of the three can be turned off with `server.builtin_tools` — useful for a server that should only ever read, or one that should not advertise discovery:
+Any of the three can be turned off with `server.builtin_tools`, independently of whether custom tools are configured — useful for a server that should only ever read, or one that serves nothing but curated profiles:
 
 ```yaml
 server:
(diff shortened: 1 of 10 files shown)


limjoobin (approved):
LGTM. Functionally, this PR registers the configured custom tool profiles as real MCP tools implemented as a wrapper around `search_records`, complete with startup validation and operational guardrails.

Also, there is a failing test in the test suite. Seems like it is coming from a misconfigured AzureOpenAI API endpoint rather than the implementation in this PR breaking anything. Feature-wise, we are good to go on this PR. Still, worth a look into the failing test case though!

PR #679, opened 2026-08-09
fix: do not fail connection setup when ACL denies both CLIENT SETINFO and ECHO

### LINKED ISSUE #673: Connection fails outright when ACL denies both CLIENT SETINFO and ECHO
## Summary

RedisVL cannot create a connection at all when the Redis ACL user is denied both `CLIENT SETINFO` and `ECHO`. Connection setup raises `NoPermissionError` before any index operation runs, with a confusing message about the `echo` command:

```
redis.exceptions.NoPermissionError: User <name> has no permissions to run the 'echo' command
```

Client identification is best-effort telemetry, so it should never block connecting.

## Root cause

`redis.exceptions.NoPermissionError` is a subclass of `ResponseError`:

```python
>>> from redis.exceptions import NoPermissionError, ResponseError
>>> issubclass(NoPermissionError, ResponseError)
True
```

So an ACL user that is denied `CLIENT SETINFO` takes the `except ResponseError` branch — but the fallback `echo` call inside that branch is not itself guarded. If the ACL also denies `ECHO`, the second `NoPermissionError` propagates out of connection setup:

```python
_lib_name = make_lib_name(kwargs.get("lib_name"))
try:
    client.client_setinfo("LIB-NAME", _lib_name)
except ResponseError:
    # Fall back to a simple log echo
    if hasattr(client, "echo"):
        client.echo(_lib_name)   # <-- unguarded
return client
```

The same unguarded-fallback pattern appears in **four** places in `redisvl/redis/connection.py`:

| Method | Approx. lines |
| --- | --- |
| `RedisConnectionFactory.get_redis_connection` | 545–552 |
| `RedisConnectionFactory._get_aredis_connection` | 604–611 |
| `RedisConnectionFactory.validate_sync_redis` | 732–739 |
| `RedisConnectionFactory.validate_async_redis` | 758–765 |

The two `validate_*` methods matter too: they run when a user passes their own pre-built client, so the failure is not limited to the `redis_url` path.

## Reproduction

Reproduced against Redis 8.4.4 using the example ACL rule from Redis Software's own documentation (the `POST /v1/redis_acls` example in [Create database roles](https://redis.io/docs/latest/operate/rs/security/access-control/create-db-roles/)):

```
ACL SETUSER u on >pw ~* &* +@read +FT.INFO +FT.SEARCH
```

Under that rule the raw commands behave as expected — `FT.INFO`, `FT.SEARCH`, and `FT.AGGREGATE` are permitted; `FT._LIST` and `FT.CREATE` are denied. But:

```python

index = SearchIndex(schema, redis_url="redis://u:pw@host:6379")
index.exists()   # raises NoPermissionError from the echo call in connection setup
```

The exception comes from connection setup, not from the search command, so it fires no matter which index operation is attempted first.

## Suggested fix

Make identification fully best-effort. Either guard the fallback in its own `try/except ResponseError: pass`, or catch `NoPermissionError` explicitly before `ResponseError` and skip identification entirely. Apply the same change to all four call sites, sync and async.

## Suggested test

A regression test with an ACL user denied both `CLIENT SETINFO` and `ECHO`, asserting that a connection can still be created and used.

`tests/integration/test_search_index.py::test_exists_under_acl_without_admin_commands` is the repo's first ACL test and establishes the fixture pattern to copy: create the user via the `client` fixture, name it with `redis_test_name`, and run `ACL DELUSER` first in a `finally` block.

## Environment

- redisvl 0.25.0
- redis-py 7.4.0
- Redis 8.4.4

## Note

This is pre-existing and unrelated to the `exists()`/`FT._LIST` change; it surfaced while documenting ACL requirements.

Aryan-Pardeshi:
Fixes #673

Client identification is best-effort telemetry, so it should never be the reason a connection cannot be created. Today it can be.

`NoPermissionError` subclasses `ResponseError`, so an ACL user denied `CLIENT SETINFO` lands in the `except ResponseError` branch as intended — but the `echo` fallback inside that branch is not itself guarded. If the same ACL also denies `ECHO`, the second `NoPermissionError` propagates straight out of connection setup, and the user sees a confusing message about the `echo` command when all they did was connect.

The same unguarded-fallback shape appears in four places in `redisvl/redis/connection.py`, so all four are fixed:

- `RedisConnectionFactory.get_redis_connection`
- `RedisConnectionFactory._get_aredis_connection`
- `RedisConnectionFactory.validate_sync_redis`
- `RedisConnectionFactory.validate_async_redis`

Each fallback `echo` is now wrapped in its own `try`/`except ResponseError` that logs at debug level and continues. No other behaviour changes: a client whose ACL permits either command still gets its library name set exactly as before.

`tests/unit/test_connection_acl.py` covers all four entry points. The two `validate_*` tests construct real `Redis`/`AsyncRedis` instances rather than mocks, because `validate_sync_redis` gates on `issubclass(type(redis_client), ...)`, which a `MagicMock(spec=Redis)` does not satisfy — construction opens no socket and both commands are patched, so no server is contacted.

Verified the tests fail against unmodified `main` (4 failed) and pass with the change (4 passed). Full unit suite is green, and `isort --profile black` / `black --target-version py311` report no changes.

Happy to fold the four guards into one shared private helper if you would prefer that shape — I kept the diff local so the fix is easy to read.

---

> [!NOTE]
> **Low Risk**
> Narrow error-handling change around best-effort client identification; successful paths unchanged and covered by new unit tests.
> 
> **Overview**
> **Connection setup no longer fails** when Redis ACL denies both `CLIENT SETINFO` and the `ECHO` fallback used for library-name telemetry.
> 
> After `client_setinfo` raises `ResponseError`, the existing `echo` fallback in `get_redis_connection`, `_get_aredis_connection`, `validate_sync_redis`, and `validate_async_redis` is now wrapped in its own `try`/`except ResponseError`. Denied `ECHO` is logged at debug and setup continues; behavior is unchanged when either command is allowed.
> 
> Adds `tests/unit/test_connection_acl.py` with four unit tests (sync/async factory and validate paths) asserting both commands are attempted without raising.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 02d54097a1cdf3be8187d6bd52aec0a68cf36191. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -551,7 +551,9 @@ class RedisConnectionFactory:
                 try:
                     client.echo(_lib_name)
                 except ResponseError as e:
-                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
+                    logger.debug(
+                        f"Failed to echo lib_name due to ACL restrictions: {e}"
+                    )
         return client
 
     @staticmethod
@@ -613,7 +615,9 @@ class RedisConnectionFactory:
                 try:
                     await client.echo(_lib_name)
                 except ResponseError as e:
-                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
+                    logger.debug(
+                        f"Failed to echo lib_name due to ACL restrictions: {e}"
+                    )
         return client
 
     @staticmethod
@@ -745,7 +749,9 @@ class RedisConnectionFactory:
                 try:
                     redis_client.echo(_lib_name)
                 except ResponseError as e:
-                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
+                    logger.debug(
+                        f"Failed to echo lib_name due to ACL restrictions: {e}"
+                    )
 
         # Module validation removed - operations will fail naturally if modules are missing
 
@@ -773,7 +779,9 @@ class RedisConnectionFactory:
                 try:
                     await redis_client.echo(_lib_name)
                 except ResponseError as e:
-                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
+                    logger.debug(
+                        f"Failed to echo lib_name due to ACL restrictions: {e}"
+                    )
 
         # Module validation removed - operations will fail naturally if modules are missing
 
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -548,7 +548,10 @@ class RedisConnectionFactory:
         except ResponseError:
             # Fall back to a simple log echo
             if hasattr(client, "echo"):
-                client.echo(_lib_name)
+                try:
+                    client.echo(_lib_name)
+                except ResponseError as e:
+                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
         return client
 
     @staticmethod
@@ -607,7 +610,10 @@ class RedisConnectionFactory:
         except ResponseError:
             # Fall back to a simple log echo
             if hasattr(client, "echo"):
-                await client.echo(_lib_name)
+                try:
+                    await client.echo(_lib_name)
+                except ResponseError as e:
+                    logger.debug(f"Failed to echo lib_name due to ACL restrictions: {e}")
         return client
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


vishal-bala:
Hi, thanks for the contribution! #685 has already implemented a more robust solution that covers this issue, so we'll close this PR for now.

PR #680, opened 2026-08-09
fix: raise provider-specific errors from vectorizer _set_model_dims()

### LINKED ISSUE #485: Improve provider-specific errors in vectorizer _set_model_dims()
## Summary
Replace generic fallback exceptions in vectorizer `_set_model_dims()` with provider-specific, actionable errors.

**Level:** Beginner

## Current State
- Several vectorizer implementations catch broad `Exception` with TODO comments.
- Resulting messages are vague and slow down troubleshooting.

## Proposed Change
- Update `_set_model_dims()` error handling in:
  - `redisvl/utils/vectorize/text/openai.py`
  - `redisvl/utils/vectorize/text/cohere.py`
  - `redisvl/utils/vectorize/text/mistral.py`
  - `redisvl/utils/vectorize/text/azureopenai.py`
  - `redisvl/utils/vectorize/text/huggingface.py`
  - `redisvl/utils/vectorize/bedrock.py`
  - `redisvl/utils/vectorize/vertexai.py`
  - `redisvl/utils/vectorize/voyageai.py`
- Catch known provider/network/auth cases and include remediation hints.

## Definition of Done
- TODO fallback comments are removed.
- Errors include provider/model context and useful next-step guidance.
- Existing tests pass; add targeted tests where needed.

## Out of Scope
- Refactoring vectorizer architecture or retry policy behavior.

Aryan-Pardeshi:
Fixes #485

Every vectorizer probes its provider with a throwaway embedding call to learn the model's dimensionality. When that probe failed, all eight raised the same message under a `# fall back (TODO get more specific)` comment:

```
Error setting embedding model dimensions: <whatever the SDK said>
```

That tells the caller nothing about which provider rejected them or what to change, which is exactly the troubleshooting cost the issue describes.

Each `_set_model_dims()` now catches the exception classes its SDK actually raises and reports the provider, the model, and a concrete next step:

- `text/openai.py`, `text/azureopenai.py` — `AuthenticationError`, `PermissionDeniedError`, `NotFoundError`, `APIConnectionError`. The Azure messages talk about deployments rather than models, since Azure addresses models by deployment name and a wrong deployment is the common failure.
- `text/cohere.py` — `UnauthorizedError`, `NotFoundError`, then `cohere.core.api_error.ApiError`.
- `text/mistral.py` — `mistralai.models.SDKError`.
- `voyageai.py` — `AuthenticationError`, `InvalidRequestError`, `APIConnectionError` from `voyageai.error`.
- `bedrock.py` — `ClientError`, split on the error code so credential problems and unknown model ids give different advice, then `BotoCoreError`.
- `vertexai.py` — `PermissionDenied`, `Unauthenticated`, `NotFound`, then `GoogleAPICallError`.
- `text/huggingface.py` — `OSError` for a model that will not load, `RuntimeError` for the CUDA OOM/device-mismatch case.

Every class name was verified by importing the installed SDK rather than taken from documentation, since a catch for a class that does not exist is dead code.

Two deliberate choices:

**The broad `except Exception` stays** as a final clause in all eight. Narrowing to provider classes alone would let anything unanticipated escape raw instead of surfacing as a `ValueError`, which would be a regression rather than a fix. The generic clause now also names the provider and model.

**Imports are local to the method**, matching how these modules already treat their optional SDK dependencies. The SDK is guaranteed importable at that point because `__init__` initialises the client before calling `_set_model_dims()`.

`tests/unit/test_vectorizer_dim_errors.py` drives OpenAI, Azure OpenAI, Bedrock and HuggingFace through their real `__init__` with the network client stubbed, so it exercises the same path a user hits with a bad key. It also pins the generic fallback: an unexpected `ZeroDivisionError` must still arrive as a `ValueError`.

Verified 7 failed against unmodified `main`, 7 passed with the change. Full unit suite: 1299 passed, 11 skipped. `isort --profile black` and `black --target-version py311` report no changes.

---

> [!NOTE]
> **Low Risk**
> Changes are limited to init-time error wrapping in embedding vectorizers; successful embed paths are unchanged and failures still surface as ValueError.
> 
> **Overview**
> When vectorizers fail their startup **dimension probe**, failures are no longer reported as a generic `Error setting embedding model dimensions: …` message. **OpenAI, Azure OpenAI, Cohere, Mistral, Bedrock, Vertex AI, and VoyageAI** now raise a **`ValueError` that names the provider and model/deployment**, with the underlying SDK or retry error preserved via **`raise … from e`**. The separate **`KeyError`/`IndexError`** branch for “unexpected API response” on that path is removed in favor of the single wrapped handler.
> 
> **Hugging Face** gets the largest behavioral change: **`SentenceTransformer` load failures** (`OSError`, `RuntimeError` such as CUDA OOM) are caught in **`_initialize_client`** with guidance on model path/download and `device='cpu'`, and similar handling is added in **`_set_model_dims`** when the probe encode step fails.
> 
> **`tests/unit/test_vectorizer_dim_errors.py`** locks in model names in messages, exception chaining, HF load errors, VoyageAI’s `TypeError` path from invalid models, and the generic fallback for unexpected errors.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5a1ebea647f0f9911ee0c056ce3cd82fc40c1fb2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/utils/vectorize/text/huggingface.py
+++ b/redisvl/utils/vectorize/text/huggingface.py
@@ -113,7 +113,22 @@ class HFTextVectorizer(BaseVectorizer):
                 "Please install with `pip install sentence-transformers`"
             )
 
-        self._client = SentenceTransformer(model, **kwargs)
+        try:
+            self._client = SentenceTransformer(model, **kwargs)
+        except OSError as e:
+            # This is where a bad model name or path actually surfaces --
+            # _set_model_dims() never sees it, since loading happens here,
+            # before that method's try block runs.
+            raise ValueError(
+                f"Could not load the local embedding model '{model}'. Check the "
+                f"model name or path and that the model has been downloaded: {str(e)}"
+            ) from e
+        except RuntimeError as e:
+            raise ValueError(
+                f"The local embedding model '{model}' failed to load. On CUDA this is "
+                f"commonly an out-of-memory or device mismatch -- retry with "
+                f"device='cpu': {str(e)}"
+            ) from e
 
     def _set_model_dims(self):
         try:
--- a/redisvl/utils/vectorize/text/huggingface.py
+++ b/redisvl/utils/vectorize/text/huggingface.py
@@ -120,9 +120,22 @@ class HFTextVectorizer(BaseVectorizer):
             embedding = self._embed("dimension check")
         except (KeyError, IndexError) as ke:
             raise ValueError(f"Empty response from the embedding model: {str(ke)}")
+        except OSError as e:
+            raise ValueError(
+                f"Could not load the local embedding model '{self.model}'. Check the "
+                f"model name or path and that the model has been downloaded: {str(e)}"
+            ) from e
+        except RuntimeError as e:
+            raise ValueError(
+                f"The local embedding model '{self.model}' failed while determining its "
+                f"dimensions. On CUDA this is commonly an out-of-memory or device "
+                f"mismatch -- retry with device='cpu': {str(e)}"
+            ) from e
         except Exception as e:  # pylint: disable=broad-except
-            # fall back (TODO get more specific)
-            raise ValueError(f"Error setting embedding model dimensions: {str(e)}")
+            raise ValueError(
+                f"Error setting embedding model dimensions for local model "
+                f"'{self.model}': {str(e)}"
+            ) from e
         return len(embedding)
 
     @deprecated_argument("text", "content")
--- a/redisvl/utils/vectorize/text/huggingface.py
+++ b/redisvl/utils/vectorize/text/huggingface.py
@@ -136,9 +136,12 @@ class HFTextVectorizer(BaseVectorizer):
         except (KeyError, IndexError) as ke:
             raise ValueError(f"Empty response from the embedding model: {str(ke)}")
         except OSError as e:
+            # Unlike _initialize_client()'s OSError handler, this one fires
+            # after SentenceTransformer already loaded successfully -- the
+            # failure is in the dimension probe/encode call, not the load.
             raise ValueError(
-                f"Could not load the local embedding model '{self.model}'. Check the "
-                f"model name or path and that the model has been downloaded: {str(e)}"
+                f"The local embedding model '{self.model}' failed while determining its "
+                f"dimensions: {str(e)}"
             ) from e
         except RuntimeError as e:
             raise ValueError(
--- a/redisvl/utils/vectorize/text/openai.py
+++ b/redisvl/utils/vectorize/text/openai.py
@@ -2,7 +2,7 @@ import os
 from typing import TYPE_CHECKING, Any
 
 from pydantic import ConfigDict
-from tenacity import RetryError, retry, stop_after_attempt, wait_random_exponential
+from tenacity import retry, stop_after_attempt, wait_random_exponential
 from tenacity.retry import retry_if_not_exception_type
 
 if TYPE_CHECKING:
@@ -152,58 +152,14 @@ class OpenAITextVectorizer(BaseVectorizer):
         Raises:
             ValueError: If embedding dimensions cannot be determined
         """
-        import openai
-
         try:
             # Use the parent embed() method which handles caching
             embedding = self._embed("dimension check")
             return len(embedding)
-        except (ValueError, RetryError) as e:
-            # _embed()/_embed_many() are @retry-decorated with
-            # retry_if_not_exception_type(TypeError), so the ValueError they
-            # raise on a permanent failure (bad credentials, unknown model...)
-            # is itself retried until tenacity gives up and raises RetryError.
-            # Unwrap that first, then unwrap the ValueError it wraps, to reach
-            # the real SDK exception either way.
-            root: BaseException = e
-            if isinstance(root, RetryError):
-                root = root.last_attempt.exception() or root
-            # _embed()/_embed_many() wrap the SDK's own exception in a ValueError
-            # (or TypeError for VoyageAI's InvalidRequestError below), so dispatch
-            # on the wrapped cause instead of the wrapper -- catching the provider
-            # SDK's exception type here would never fire otherwise.
-            cause = root.__cause__ or root.__context__ or root
-            if isinstance(cause, (KeyError, IndexError)):
-                raise ValueError(
-                    f"Unexpected response from the OpenAI API: {str(cause)}"
-                ) from e
-            if isinstance(
-                cause, (openai.AuthenticationError, openai.PermissionDeniedError)
-            ):
-                raise ValueError(
-                    f"OpenAI rejected the credentials used while determining embedding "
-                    f"dimensions for model '{self.model}'. Check the api_key given in "
-                    f"api_config, or the OPENAI_API_KEY environment variable: {str(cause)}"
-                ) from e
-            if isinstance(cause, openai.NotFoundError):
-                raise ValueError(
-                    f"OpenAI does not recognize the embedding model '{self.model}'. Check "
-                    f"the model name against the provider's current model list: {str(cause)}"
-                ) from e
-            if isinstance(cause, openai.APIConnectionError):
-                raise ValueError(
-                    f"Could not reach the OpenAI API while determining embedding dimensions "
-                    f"for model '{self.model}'. Check network access and any proxy "
-                    f"configuration: {str(cause)}"
-                ) from e
-            raise ValueError(
-                f"Error setting embedding model dimensions for OpenAI model "
-                f"'{self.model}': {str(e)}"
-            ) from e
         except Exception as e:  # pylint: disable=broad-except
             raise ValueError(
                 f"Error setting embedding model dimensions for OpenAI model "
-                f"'{self.model}': {str(e)}"
+                f"'{self.model}': {e}"
             ) from e
 
     @deprecated_argument("text", "content")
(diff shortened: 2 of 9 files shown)


Aryan-Pardeshi:
Good catch, and it went deeper than the first fix. `_embed()`/`_embed_many()` already wrap the SDK's own exception in a generic `ValueError` before `_set_model_dims()` sees it, so catching the provider exception type directly never fired — confirmed.

There was a second layer underneath that too: `_embed`/`_embed_many` are `@retry`-decorated with `retry_if_not_exception_type(TypeError)`, which does not exempt that `ValueError`. So a permanent failure like bad credentials gets retried 6 times with exponential backoff before tenacity gives up and raises `RetryError`, wrapping the `ValueError`, which wraps the real SDK exception.

`_set_model_dims()` now unwraps `RetryError` first, then `__cause__`/`__context__`, before dispatching. Tests for OpenAI, Bedrock and HuggingFace now drive the real `_embed()`/`_initialize_client()` code (patching only the network client, not `_embed` itself) so they'd have caught this the first time. HuggingFace also got a real fix for the case you flagged separately — a bad model name now raises where it actually happens, in `_initialize_client()`'s `SentenceTransformer()` call, not in `_set_model_dims()`.

Full unit suite: 1303 passed, no regressions.

Aryan-Pardeshi:
Good catch again — pushed. `_embed_many()` re-raises `InvalidRequestError` as `TypeError` specifically so the retry decorator skips retrying it (a bad model id can't ever succeed). That meant it reached `_set_model_dims()` as a bare, unwrapped `TypeError`, never caught by the `(ValueError, RetryError)` tuple, so it fell to the generic fallback instead of the unrecognized-model message.

Added `TypeError` to the caught tuple and a test that drives the real `_embed_many()` code (only the client's `.embed()` call is stubbed), which fails against the previous except tuple and passes now. Full suite: 1304 passed.

Aryan-Pardeshi:
Both real, pushed a fix for each.

High: right, cohere.UnauthorizedError/NotFoundError only exist in cohere 5.0+, but the package still declares cohere>=4.44 and the embed path still handles 4.x list responses. Those were referenced unconditionally at import/isinstance time, so any 4.x install would crash on construction. Now using getattr(cohere, "UnauthorizedError", None) etc., and the ApiError import is wrapped in try/except ImportError -- falls through to the generic message on 4.x instead of crashing.

Low: fixed the copy-paste wording, HF's OSError handler now says "failed while determining its dimensions" like the RuntimeError one, since it fires after SentenceTransformer already loaded.

Full suite: 1304 passed, 11 skipped, no regressions.

vishal-bala (changes requested):
Hi, thanks for contributing to RedisVL! I appreciate the idea you're going for here, but the current implementation adds a lot of bloated logic and comments for limited lift to an end user. If the goal is to expose the underlying specific error more concretely, I think we can do that by simply chaining the exception into the `ValueError`

```python
except (...) as ke:
    raise ValueError("...") from ke
```

That should make it sufficiently visible in the traceback and actionable by others catching this exception, while also maintaining the standard of code quality we expect for this project.

Aryan-Pardeshi:
@vishal-bala thanks — agreed, and I took the whole thing out rather than trimming it.

`5a1ebea` collapses every vectorizer's `_set_model_dims()` to one `except Exception as e: raise ValueError(...) from e`. The provider-specific handlers, the `RetryError.last_attempt.exception()` unwrapping and the `__cause__`/`__context__` dispatch are all gone; `from e` already puts the real SDK error in the traceback, which is what the extra code was reimplementing badly. The message still names the provider and the model so the actionable part survives.

Net diff across the 8 files is **+80 / -512**.

Verified: `black --check` and `isort --check-only` clean on the touched files, `mypy` reports "Success: no issues found in 7 source files". The rewritten `tests/unit/test_vectorizer_dim_errors.py` covers OpenAI/Bedrock/HuggingFace end to end against real SDK failures plus a parametrized cause-chaining check for the rest — I ran those standalone rather than through pytest, since `tests/conftest.py`'s session-scoped `redis_container` fixture needs Docker for even pure unit tests.

The Bugbot comments above all predate this rewrite and point at handlers that no longer exist.

Aryan-Pardeshi:
The latest commit already simplifies this to the single chained `raise ValueError(...) from ke` you suggested, and addresses the earlier Bugbot findings about the specific-exception handlers not being reachable. Should be ready for another look whenever you get a chance.

redisvl/utils/vectorize/text/openai.py
```diff
@@ -156,11 +156,11 @@ def _set_model_dims(self) -> int:
...
         except Exception as e:  # pylint: disable=broad-except
-            # fall back (TODO get more specific)
-            raise ValueError(f"Error setting embedding model dimensions: {str(e)}")
+            raise ValueError(
+                f"Error setting embedding model dimensions for OpenAI model "
+                f"'{self.model}': {e}"
+            ) from e
 
     @deprecated_argument("text", "content")
     @retry(
```
vishal-bala, line 166: This decorator is what swallows the cause. `retry_if_not_exception_type(TypeError)` doesn't exempt the `ValueError` that `_embed` raises, so six attempts run and tenacity wraps the result in `RetryError`. Adding `reraise=True` here lets the original propagate, and your new message then shows the real reason.

Same change needed in `mistral.py`, `bedrock.py`, and on `voyageai.py`'s `_embed_many` decorator. `azureopenai.py:214` already has it if you want a reference.

redisvl/utils/vectorize/text/openai.py
```diff
@@ -156,11 +156,11 @@ def _set_model_dims(self) -> int:
             # Use the parent embed() method which handles caching
```
vishal-bala, line 156: Small one while you're here: this comment says the opposite of the code. `embed()` is the caching path and `_embed()` bypasses it, which is what you want for a throwaway probe. The other seven vectorizers say "Call the protected _embed method to avoid caching this test embedding". Could you match them?

redisvl/utils/vectorize/text/huggingface.py
```diff
@@ -113,16 +113,47 @@ def _initialize_client(self, model: str, **kwargs):
...
+        try:
+            self._client = SentenceTransformer(model, **kwargs)
+        except OSError as e:
+            # This is where a bad model name or path actually surfaces --
+            # _set_model_dims() never sees it, since loading happens here,
+            # before that method's try block runs.
+            raise ValueError(
+                f"Could not load the local embedding model '{model}'. Check the "
+                f"model name or path and that the model has been downloaded: {str(e)}"
+            ) from e
```
vishal-bala, line 125: `OSError` is wider than it looks. `huggingface_hub`'s `GatedRepoError`, `LocalEntryNotFoundError`, `OfflineModeIsEnabled`, and `HfHubHTTPError` all subclass it, and so does `requests.exceptions.RequestException`. So a gated model with a perfectly correct name, or a proxy blocking huggingface.co, both get told to check the model name and whether it's downloaded. `RepositoryNotFoundError` is the only one that advice actually fits.

Either catch `RepositoryNotFoundError` specifically, or widen the wording to cover the rest, along the lines of "check the model name or path, your network access to huggingface.co, and whether the repo requires accepting a licence".

tests/unit/test_vectorizer_dim_errors.py
```diff
@@ -0,0 +1,227 @@
...
+    except Exception clause."""
+    import voyageai.error
+
+    from redisvl.utils.vectorize.voyageai import VoyageAIVectorizer
+
+    bad_model = voyageai.error.InvalidRequestError("model not found")
+    mock_client = MagicMock()
+    mock_client.embed.side_effect = bad_model
+
+    def _fake_init(self, *a, **k):
```
vishal-bala, line 201: This test passes without testing anything. `_fake_init` doesn't set `_token_batching_supported`, which the real `_initialize_client` sets and `_embed_many` reads, so construction dies on `AttributeError` before your stubbed `.embed()` is ever called: `mock_client.embed.call_count` is 0. The `InvalidRequestError` to `TypeError` path your docstring describes never runs, and the assertions on lines 210-211 are weak enough not to catch it.

It also costs about 18 seconds, because `AttributeError` isn't `TypeError` so it retries six times with real backoff. Setting `self._token_batching_supported = True` here, plus the `monkeypatch.setattr(time, "sleep", ...)` guard your OpenAI and Bedrock tests use on lines 41 and 69, fixes both.

vishal-bala (commented):
Thanks for turning this around. Dropping the per-SDK exception classes was the right call, and the chained `ValueError` is what I was after. A few things left.

The retry decorator sitting above each of these handlers decides what the handler ever sees. When `_embed` raises, tenacity retries it and then wraps whatever escaped in a `RetryError`, so the handler formats the wrapper rather than the cause.

That splits the eight providers in two. Azure sets `reraise=True`, Vertex exempts `ValueError` from retry, and Cohere and HuggingFace aren't retried at all, so those four print the provider's real error. OpenAI, Mistral, Bedrock, and VoyageAI (whose `_embed` delegates to a retried `_embed_many`) print this instead, for a bad API key:

    Error setting embedding model dimensions for OpenAI model 'text-embedding-3-small': RetryError[<Future at 0x10b2030b0 state=finished raised ValueError>]

Naming the model is a real improvement. But the reason a user needs is still a `RetryError` repr with a memory address in it, which is roughly where they started. Adding `reraise=True` to those four decorators, as Azure already does, puts the provider's own message back into the string.

Two more asks. Please update the description before this merges: it still describes the per-SDK exception classes from the earlier revision and cites 7 tests where there are now 10, so the history would record something quite different from what landed. And the test file is 227 lines against roughly 40 lines of source change. The parametrised block and the HuggingFace load test do most of the work; trimming toward 100 lines would match the rest of `tests/unit/`.

Happy to approve once those are in.

PR #683, opened 2026-08-09
feat: add delete_where()/adelete_where() and share batching internals with clear()

### LINKED ISSUE #488: Add delete_where() API for bulk index deletion by filter
## Summary
Add `delete_where()` / `adelete_where()` to delete index documents by filter expression, and refactor `clear()` to share the same internals.

**Level:** Intermediate

## Current State
- `clear()` deletes all documents via batched `FilterQuery("*")` flow.
- `drop_documents()` / `drop_keys()` only delete known IDs/keys.
- There is no public API to delete a filtered subset of index documents.

## Proposed Change
- Add:
  - `delete_where(filter_expression, batch_size=500) -> int`
  - `adelete_where(filter_expression, batch_size=500) -> int`
- Refactor both `clear()` and `delete_where()` to use shared private internals:
  - common batch-ID fetch loop
  - existing `_delete_batch()` and async equivalent as delete primitive
- Implement `clear()` as `delete_where(FilterExpression("*"))`.

## Definition of Done
- Filtered bulk deletion works on standalone and cluster clients.
- `clear()` delegates to the shared deletion internals (no duplicated batch loop).
- Integration tests cover selective match, no-op match, and full-clear behavior.
- Docs include one usage example.

## Out of Scope
- Hard-delete audit logging or soft-delete support.

Aryan-Pardeshi:
Fixes #488

Adds `delete_where()` / `adelete_where()` for deleting a filtered subset of an index, and refactors `clear()` onto the same internals.

Before this there was no public way to delete by filter: `clear()` removes everything, and `drop_documents()` / `drop_keys()` need ids or keys you already hold.

`clear()` and `_delete_where()` were the same batch loop with a different filter expression, so the loop now lives once in `_delete_where()` (and `_adelete_where()`), and `clear()` is a call to it with the match-all expression.

Two details from the original `clear()` are preserved deliberately:

- **The `max_ratio = 1.01` overshoot guard.** It stops the loop running forever against concurrent inserts. It is now computed from a `CountQuery` on the supplied filter rather than `info()["num_docs"]`, so the bound tracks the documents actually being deleted instead of the whole index — which matters once the filter is not match-all.
- **`invalidate_sql_schema_cache()`** still runs at the end of every deletion path.

`_delete_batch()` is untouched, so the cluster-specific key-at-a-time behaviour is unchanged.

`tests/integration/test_delete_where.py` covers a filtered delete leaving non-matching documents in place, a filter matching nothing returning `0` and deleting nothing, `clear()` keeping its existing contract, and a `batch_size` below the match count so the paging loop is exercised. Sync and async both covered. 7 passed.

For regression cover on the `clear()` change I also ran `tests/integration/test_search_index.py` and `test_async_search_index.py` on this branch: 95 passed.

One gap I would rather state than paper over: **I have not run the cluster suite.** `clear()` is public API with cluster-specific behaviour underneath, and while `_delete_batch()` is unchanged I would not call that fully verified. If you have a cluster CI job, that is the one to watch here.

---

> [!NOTE]
> **Medium Risk**
> Refactors public clear() and introduces destructive delete-by-filter on index data; cluster behavior relies on unchanged _delete_batch and is not covered by new cluster tests.
> 
> **Overview**
> Adds **`delete_where()`** and **`adelete_where()`** on sync/async search indexes so callers can remove documents by filter without pre-fetching ids or keys.
> 
> The shared **`_delete_where` / `_adelete_where`** path counts matches with **`CountQuery`**, pages ids with **`FilterQuery`**, deletes via existing **`_delete_batch`** ( **`DEL`** ), applies the **`1.01`** runaway cap from that count, and still calls **`invalidate_sql_schema_cache()`**. **`clear()`** now delegates to the same helper with a match-all filter instead of duplicating the loop; the overshoot bound is tied to the filter’s match count rather than **`info()["num_docs"]`**.
> 
> Integration tests cover filtered delete, no-op filters, **`clear()`** behavior, small **`batch_size`** paging, and async parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0592f2c6647793e4fb3f5271953a56fdc764ce38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -1149,9 +1149,7 @@ class SearchIndex(BaseSearchIndex):
         batch_size: int = DEFAULT_BULK_BATCH_SIZE,
     ) -> int:
         filter_expr = (
-            FilterExpression("*")
-            if filter_expression is None
-            else filter_expression
+            FilterExpression("*") if filter_expression is None else filter_expression
         )
         matched_count = cast(int, self.query(CountQuery(filter_expr)))
         max_ratio = 1.01
@@ -2511,9 +2509,7 @@ class AsyncSearchIndex(BaseSearchIndex):
         batch_size: int = DEFAULT_BULK_BATCH_SIZE,
     ) -> int:
         filter_expr = (
-            FilterExpression("*")
-            if filter_expression is None
-            else filter_expression
+            FilterExpression("*") if filter_expression is None else filter_expression
         )
         matched_count = cast(int, await self.query(CountQuery(filter_expr)))
         max_ratio = 1.01
--- a/redisvl/index/index.py
+++ b/redisvl/index/index.py
@@ -1143,26 +1143,22 @@ class SearchIndex(BaseSearchIndex):
             records_deleted_in_batch = cast(int, client.delete(*batch_keys))
         return records_deleted_in_batch
 
-    def clear(self) -> int:
-        """Clear all keys in Redis associated with the index, leaving the index
-        available and in-place for future insertions or updates.
-
-        NOTE: This method requires custom behavior for Redis Cluster because
-        here, we can't easily give control of the keys we're clearing to the
-        user so they can separate them based on hash tag.
-
-        Returns:
-            int: Count of records deleted from Redis.
-        """
-        batch_size = 500
+    def _delete_where(
+        self,
+        filter_expression: str | FilterExpression | None = None,
+        batch_size: int = DEFAULT_BULK_BATCH_SIZE,
+    ) -> int:
+        filter_expr = (
+            FilterExpression("*")
+            if filter_expression is None
+            else filter_expression
+        )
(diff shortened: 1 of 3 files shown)


Aryan-Pardeshi:
Closing this one myself — `drop_by_filter()` already exists on `main` (`index.py:1306` sync, `2638` async) and covers everything issue #488 asked for: deletes by filter expression, takes `batch_size`, plus `dry_run`, `allow_all`, and `on_progress` that this PR didn't have. It also correctly avoids the `MAXSEARCHRESULTS` limit via re-querying from offset 0, which this PR's `clear()` refactor also does but I hadn't connected to the existing API. I'll leave a note on the issue as well. Apologies for the noise.

PR #685, opened 2026-08-12
fix: let a connection open when CLIENT SETINFO is denied

### LINKED ISSUE #673: Connection fails outright when ACL denies both CLIENT SETINFO and ECHO

vishal-bala:
`RedisVL` announces itself on connect with `CLIENT SETINFO LIB-NAME` and, if that is refused, sends the same string through `ECHO`. The `ECHO` call was unguarded, so `NoPermissionError` escaped while the connection was being created, before any index operation could be attempted:

```
NoPermissionError: User <name> has no permissions to run the 'echo' command
```

An application role assembled from `@read`/`@write` hits this. `CLIENT SETINFO` is tagged `@connection` and `@slow`, `ECHO` is `@connection` and `@fast`, and neither is in `@read` or `@write` — so a rule built up from those categories never grants either.

**Subtracting `@dangerous` is not what denies them.** Measured on Redis 8.4.5, `+@all -@dangerous` permits both, so the shape that fails is a hand-written `+@read +@write` rule rather than a broad rule with exclusions — which also means Redis Cloud's predefined Read-Write shape is unaffected, while a `@read`-shaped Read-Only rule would be affected. Note too that `+@read +@write +@slow` permits `CLIENT SETINFO` while still denying `ECHO`, so the fallback was never a reliable second chance.

## Why the fallback is deleted rather than guarded

It was added in 934d269 (#155) as a telemetry breadcrumb for servers older than Redis 7.2, where `CLIENT SETINFO` does not exist, so the library name would at least appear in `MONITOR` or the slowlog. It earns nothing today:

- It fires only when `CLIENT SETINFO` errors, and the ACL rule that denies one denies the other.
- Its argument reaches nothing that reads `lib-name` — `CLIENT LIST` and `CLIENT INFO` take that field from `SETINFO` alone, and an O(1) `ECHO` only enters the slowlog with `slowlog-log-slower-than` near zero.
- redis-py already covers the old-server case one layer down, sending its own `CLIENT SETINFO` during the connection handshake under `try/except ResponseError: pass`.

The `hasattr(client, "echo")` guard goes with it, along with the comment claiming `RedisCluster` has no `echo`; both `redis.cluster.RedisCluster` and `redis.asyncio.cluster.RedisCluster` expose it.

## What is kept, deliberately

The four duplicated blocks become one sync and one async helper. The explicit `client_setinfo` call stays: redis-py's handshake reports its own name, and this overwrites it with the composed `redis-py(redisvl_v…;<wrapper>)` string that adoption metrics read. The helper docstring says so, so it does not read as redundant with the handshake.

`except ResponseError` also stays narrow. In the connection-factory path this is the first command on a freshly created connection and therefore the de-facto connectivity check, so broadening to `RedisError` would swallow `ConnectionError` and defer a real failure to some later command. `AuthenticationError` subclasses `ConnectionError`, not `ResponseError`, so `WRONGPASS` and `NOAUTH` keep propagating either way.

## Tests

`tests/unit/test_client_identification.py` — 16 cases over both twins. Mutation-checked: restoring the `ECHO` fallback fails six unit cases plus the integration test, and broadening the `except` to `Exception` fails two.

Both refusals are covered, since only one is a permission problem: a plain `ResponseError` stands for a pre-7.2 server, which is the case that lets the fallback go. Identification is asserted on all three URL shapes — sentinel, cluster, standalone — because it sits after that fan-out and moving it into one branch otherwise goes unnoticed; live cluster tests need `--run-cluster-tests` and never run in CI.

One integration test opens a connection under `+@read +@write`, with the premise pinned: `CLIENT SETINFO` must raise `NoPermissionError` for that user, so the test cannot go vacuous if Redis ever grants it to that role.

ACL user setup moves into an `acl_user` fixture, reused by the existing `-@admin` test. Rules are applied after `reset` because `ACL SETUSER` is additive and usernames are derived from the test's node id; the user is dropped before the connections it authenticated; and the fixture skips on deployments that reject `ACL SETUSER`.

## Docs

`docs/user_guide/installation.md` said a credential permitted to run neither command fails at connection time. That is no longer true. The replacement names `+client|setinfo` for anyone who wants RedisVL attributed in `CLIENT LIST`, and is explicit that this labels only the connection RedisVL opens — redis-py labels the rest of the pool as plain `redis-py`.

It also gains a cluster caveat found while verifying this fix: `CLUSTER SLOTS` is tagged `@slow` only, so `RedisCluster.from_url` cannot discover the topology under a `+@read +@write` rule and the connection fails before identification is attempted, reported as the misleading `Redis Cluster cannot be connected`. Such a deployment needs `+cluster|slots` as well.

## Not in scope

The rest of the ACL documentation pass. Four statements in `installation.md` are stale for a different reason — they predate a `create_index=False` opt-out for the extension constructors — and are corrected alongside it in a follow-up PR.

Identification still reaches only one connection: the explicit call labels whichever pooled connection it borrows, not the rest of the pool or any reconnect, and on a cluster redis-py routes it to the default node alone. Passing the composed name in as redis-py's `lib_name`/`driver_info` would fix that, but `redis>=5.0,<8.0` straddles the deprecation of `lib_name` in favour of `driver_info` and needs version-conditional handling.

---

> [!NOTE]
> **Low Risk**
> Behavior change is limited to connect-time client labeling; real connectivity and auth errors still propagate, and the main effect is fixing false failures on restrictive ACLs.
> 
> **Overview**
> Fixes connection failures for ACL roles built from **`+@read` / `+@write`** that cannot run **`CLIENT SETINFO`**. Identification is centralized in **`_identify_client`** / **`_aidentify_client`**: the library still attempts **`CLIENT SETINFO LIB-NAME`** for adoption metrics, but a **`ResponseError`** (ACL denial or pre-7.2 server) is logged at debug and ignored. The old **`ECHO`** fallback is removed because it was unguarded and failed with the same permission shape.
> 
> **`installation.md`** now states that denied identification does not block connect, documents optional **`+client|setinfo`**, and notes that cluster clients need **`+cluster|slots`** when topology discovery is not covered by read/write categories.
> 
> Tests add an **`acl_user`** fixture, integration coverage for restricted credentials, and unit tests that refused SETINFO still returns a client, **`ECHO`** is not called, and **`ConnectionError`** is not swallowed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit edadd6c342c011007cdd1cb608b993a496fc58a4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/user_guide/installation.md
+++ b/docs/user_guide/installation.md
@@ -211,7 +211,9 @@ Index enumeration is the only thing an `-@admin` rule breaks. It is reached by `
 
 Other categories gate different operations. `FT.DROPINDEX` is tagged `@dangerous` and `@write` as well as `@search`, so a policy that subtracts `@dangerous` denies `index.delete()`, `rvl index delete`, and `rvl index destroy`.
 
-Outside of Redis Search, RedisVL identifies itself on connect with `CLIENT SETINFO` and falls back to `ECHO`. A credential permitted to run neither currently fails when the connection is created, before any index operation.
+Outside of Redis Search, RedisVL identifies itself on connect with `CLIENT SETINFO`. That command is tagged `@connection` and `@slow`, and belongs to neither `@read` nor `@write`, so a rule built up from those categories never grants it. A credential that cannot run it still connects: identification only populates the `lib-name` field that `CLIENT LIST` and `CLIENT INFO` display, so a refusal is ignored (and logged, if you have configured logging at debug level). Grant `+client|setinfo` if you want RedisVL to appear as the connecting library there — note that this labels the connection RedisVL opens, while redis-py labels the rest of the pool as plain `redis-py`.
+
+Cluster deployments need one more grant. `RedisCluster` discovers the topology with `CLUSTER SLOTS`, which is tagged `@slow` only, so a credential assembled from `+@read +@write` cannot open a clustered connection at all — redis-py reports this as `Redis Cluster cannot be connected`, with the underlying permission error chained beneath it. Grant `+cluster|slots` alongside the rules above.
 
 ### Redis Cloud and Redis Software
 
--- a/redisvl/redis/connection.py
+++ b/redisvl/redis/connection.py
@@ -212,6 +212,43 @@ def make_lib_name(*args) -> str:
     return f"redis-py({custom_libs})"
 
 
+def _identify_client(client: SyncRedisClient, lib_name: str | None = None) -> None:
+    """Report RedisVL as the connecting library, tolerating a refusal.
+
+    redis-py sends its own ``CLIENT SETINFO`` during the connection handshake,
+    so this call is here to overwrite that with the composed
+    ``redis-py(redisvl_v...;<wrapper>)`` name that adoption metrics read. Keep
+    it: without it the library and any wrapper above it go unattributed.
+
+    A refusal is ignored, because ``CLIENT SETINFO`` only populates the
+    ``lib-name`` field that ``CLIENT LIST`` and ``CLIENT INFO`` display -- its
+    own documentation tells client libraries to ignore failures. Two credentials
+    hit this: one granting neither ``@connection`` nor the command itself, and
+    one on a server predating Redis 7.2, where the command does not exist.
+
+    Only ``ResponseError`` is caught. In the connection-factory path this is the
+    first command issued on a freshly created connection, which makes it the
+    de-facto connectivity check, so swallowing ``ConnectionError`` would defer a
+    genuine failure to some later and more confusing command. Note that on a
+    cluster client redis-py routes the command to the default node only, so the
+    label reaches one node rather than the whole cluster.
+    """
+    try:
+        client.client_setinfo("LIB-NAME", make_lib_name(lib_name))
+    except ResponseError as e:
+        logger.debug(f"CLIENT SETINFO was not applied, continuing without it: {e}")
+
+
+async def _aidentify_client(
+    client: AsyncRedisClient, lib_name: str | None = None
+) -> None:
+    """Async version of :func:`_identify_client`."""
+    try:
+        await client.client_setinfo("LIB-NAME", make_lib_name(lib_name))
+    except ResponseError as e:
+        logger.debug(f"CLIENT SETINFO was not applied, continuing without it: {e}")
+
+
 def convert_index_info_to_schema(index_info: dict[str, Any]) -> dict[str, Any]:
     """Convert the output of FT.INFO into a schema-ready dictionary.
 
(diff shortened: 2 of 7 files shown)


nkanu17 (approved):
lgtm!

PR #687, opened 2026-08-13
feat: let callers opt out of index creation with create_index=False

vishal-bala:
Stacked on #685, which fixes the connection-time failure this PR's credential also hits. Review that one first.

Every extension checks whether its index exists while being constructed, and that check is `FT.INFO`. A credential assembled from `+@read +@write` is denied `FT.INFO` and `FT.CREATE` together — neither command is in either category, identically on Redis 8.0.6 through 8.8.1 — so such a role cannot construct `SemanticCache`, `MessageHistory`, `SemanticMessageHistory` or `SemanticRouter` at all, even against an index it can query perfectly well:

```
RedisSearchError: Error while fetching llmcache index info:
User <name> has no permissions to run the 'FT.INFO' command
```

**There was no way to ask for less.** `overwrite=False` is the *reason* the check runs — `create()` calls `exists()` first and consults `overwrite` only afterwards — `drop` is not a constructor parameter at all, and `overwrite=True` is strictly worse, since it proceeds to `FT.DROPINDEX`. `SearchIndex` exposes no lifecycle seam either, and each constructor calls `create()` inline with no hook to subclass around.

`create_index=False` lets the caller state what RedisVL cannot ask: the index exists. It skips the existence check, the schema comparison against the live index, and creation, so the constructor issues no index command at all. It is rejected together with `overwrite=True`, which asks for the opposite.

## Why not a runtime probe

`FT.SEARCH` is permitted where `FT.INFO` is not, so probing with it looks attractive. Its reply is identical for an index and for an alias pointing at one, though, so a probe would reinstate the `create(overwrite=True, drop=True)` → `FT.DROPINDEX <alias> DD` data-loss path that #672 closed. And a credential that cannot ask whether the index exists cannot create one either, so there is nothing to work out at runtime — `exists()` keeps failing loudly instead.

## The invariant this had to fix first

Constructor-time `create()` was the de-facto eager connect. `SearchIndex.client` returns the raw client and is `None` until the lazy `_redis_client` property runs, so skipping `create()` left ten `self._index.client` sites in `redisvl/extensions/` dereferencing `None` — starting with the router's `route_config` write, which runs immediately after index setup. All ten now use `_redis_client`. Two of the `# type: ignore` comments they carried turned out to cover a real `scan_by_pattern` signature mismatch rather than the Optional, and are kept with explicit codes.

## Router semantics

`create_index=False` means the index exists, is already seeded, and is not ours to rewrite, so the router also skips writing route references and the stored `route_config`. Rewriting that blob from an unverified local route list would truncate a shared router's routes, and `JSON.SET` is `@write`, so a restricted credential can do it. `_update_router_state()` stays armed — `add_route()` and `remove_route()` are the caller acting deliberately — but that consequence is now stated in `add_route()`'s own docstring as well as the guide, since the docstring is the reference for anyone who never opens the guide.

`SemanticRouter.from_existing()` threads the flag through. It reads the stored config with `JSON.GET` and reaches `FT.INFO` only via the constructor, so with `create_index=False` it issues no index command and becomes the way to attach to a router under a restricted credential. The flag is popped before `_split_from_existing_kwargs`, which retains only `SearchIndex` init kwargs and would otherwise pass it to the Redis client constructor — where `SearchIndex.__init__` discards unknown kwargs silently, making the mistake invisible.

Separately, `routes=[]` now raises a useful error when matching instead of `max() arg is an empty sequence`. That is unconditional: emptiness is legal on either path, and a flag about index ownership should not decide it.

## Tests

`tests/unit/test_extension_create_index_flag.py` — 17 cases. The contract is "no index command at all", so they assert on the client: a `MagicMock` records every call, and `ft()` is the gate every `FT.*` command passes through. All four constructors reach zero recorded calls.

Two cases pin the lazy-connect invariant above by driving `drop(id=...)` and `get_route_references()` as the first operation — reverting any of the ten conversions otherwise breaks no test. Others pin that the flag survives as instance state, that it never reaches the router's stored config, and that `from_existing()` still verifies by default.

The default construction path is not re-tested here; the existing 189 integration tests already fail if `create()` is dropped.

One integration test round-trips `store()`/`check()` through a cache built with `create_index=False` under a real `+@read +@write -@dangerous` ACL user — the customer's rule, so the destructive commands it denies stay denied — with the premise pinned (`FT.INFO` must raise `NoPermissionError` for that user) and the negative alongside it: without the flag the same credential raises `RedisSearchError` naming `ft.info`, with `NoPermissionError` chained.

## Docs

The ACL section of `docs/user_guide/installation.md` is restructured. Four statements were falsified by this change or were already wrong: that a credential needs `@search` at all, that all four extensions always call `create()`, that enumeration is the only thing an `-@admin` rule breaks, and the advice to grant `FT.CREATE`. The operation table gains a `+@read +@write` column, the command-to-category mapping is labelled as measured rather than documented, and the wrapped error text appears verbatim under its own heading — an H3, so it has an anchor to link to.

The pre-existing key-permission material became its own section rather than being dropped, and was corrected while there: partial key-pattern overlap is denied exactly like no overlap, not filtered down to the readable subset, and `FT.CREATE` is not key-checked at all, so a credential can create an index it cannot query.

A new subsection covers what the flag gives up. An absent index fails loudly, and a vector dimension mismatch does too — but only once the index holds a document, which a freshly provisioned index will not. A wrong prefix, an `ON JSON` index written as hashes, and a differing datatype or distance metric are silent. The tell is `FT.INFO`'s `key_type`, `prefixes` and `attributes`, not `hash_indexing_failures`, which stays `0` because those keys were never indexing candidates. Router provisioning gets its own subsection, since preparing one for this mode needs embedded reference vectors rather than a hand-written `FT.CREATE`.

Two corrections worth calling out for reviewers who know this area: `clear()` is not uniform — only `SemanticCache.clear()` avoids `FT.INFO`, while the other three delegate to `SearchIndex.clear()`, which calls `info()` first — and Redis Cloud's predefined Read-Write rule reads as `@read`/`@write`-shaped from its published description, so it is a candidate for this problem rather than immune to it.

## Not in scope

`create_index=False` restrains construction only; `delete()` and `clear()` stay armed. Coupling ownership to the flag is the coherent next step, and the flag is stored as instance state so it can be added without another parameter.

`adk_redis` needs a matching field on `RedisVLCacheProviderConfig` before this reaches callers who construct through that provider. Until then the interim for the escalation remains the `exists()` monkeypatch already shared in the thread.

---

> [!NOTE]
> **Medium Risk**
> Behavior changes constructor and lifecycle semantics for four public extension APIs under restricted ACLs; mistakes with externally managed indexes can fail silently on schema mismatch, though destructive index operations are now guarded.
> 
> **Overview**
> Adds **`create_index=False`** to **`SemanticCache`**, **`MessageHistory`**, **`SemanticMessageHistory`**, and **`SemanticRouter`** so constructors skip **`FT.INFO`**, schema checks, and **`FT.CREATE`**—the path needed when a **`+@read +@write`** role cannot run index-metadata commands. **`overwrite=True`** with this flag is rejected via shared **`CREATE_INDEX_OVERWRITE_CONFLICT`** messaging.
> 
> **Index-wide `delete()` / `clear()`** (plus async cache **`adelete` / `aclear`**) now raise **`EXTERNAL_INDEX_LIFECYCLE_CONFLICT`** when the extension does not own the index. Per-entry **`drop`** remains allowed.
> 
> **`SemanticRouter`** with the flag does not seed reference vectors or write **`route_config`**; **`from_existing(..., create_index=False)`** threads the flag and avoids **`FT.*`** on attach. Matching with **`routes=[]`** fails with an explicit error instead of an empty **`max()`**.
> 
> Fixes lazy-connect breakage from skipping constructor **`create()`** by using **`_index._redis_client`** instead of **`_index.client`** across extensions.
> 
> **Docs** expand Redis ACL guidance (operation vs **`+@read +@write`** table, **`create_index=False`** usage, router provisioning, silent schema mismatches) and point **`FT.INFO`** permission errors to the new flag in **`exceptions.rst`**. **Tests** add unit coverage for “no index commands at construction” and an integration ACL round-trip.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 106f8b0deaa1d709b0de2dc244969423d5b2e45e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/user_guide/installation.md
+++ b/docs/user_guide/installation.md
@@ -186,30 +186,102 @@ The Sentinel URL format supports:
 
 ## Redis permissions (ACLs)
 
-RedisVL works through Redis Search commands, so a connecting credential needs the `@search` ACL category, or the individual `FT.*` commands. Reading an index additionally requires key permissions covering its prefix: the [ACL documentation](https://redis.io/docs/latest/operate/oss_and_stack/management/security/acl/#command-categories) describes this rule for creating, modifying, and reading an index, and in practice `FT.INFO`, `FT.SEARCH`, and `FT.AGGREGATE` are denied when the index prefix falls outside the allowed key patterns. `FT.CREATE` is not checked this way, so a credential can create an index it is then unable to read.
+RedisVL reaches Redis through Redis Search commands, but not all of them need the `@search` category. Querying and loading work under an ordinary `+@read +@write` role; it is the commands that inspect or manage an index — `FT.INFO`, `FT.CREATE`, `FT._LIST` — that need `@search` or an explicit grant.
 
-One command needs more than `@search`. Redis tags `FT._LIST` as `@admin` as well as `@search` and `@slow`, and ACL rules are applied left to right — so a rule that grants search access and then takes back administrative commands, such as `+@search -@admin` or `+@all -@admin`, denies it:
+The command-to-category mapping below was measured against live servers rather than quoted from published documentation, which does not list ACL categories per `FT.*` command. It was identical on Redis 8.0.6, 8.2.7, 8.4.5, 8.6.4, 8.8.0 and 8.8.1. Check your own deployment with `COMMAND INFO ft.info`, `ACL CAT search`, or `ACL DRYRUN <user> FT.INFO <index>`.
+
+### What each operation needs
+
+| Operation | Redis command | `+@all -@admin` | `+@read +@write` |
+|---|---|---|---|
+| `index.query()`, `index.search()`, `index.aggregate()` | `FT.SEARCH`, `FT.AGGREGATE` | Yes | Yes |
+| `index.load()` | `HSET` or `JSON.SET` (needs key access) | Yes | Yes |
+| `index.exists()`, `index.info()`, `index.clear()`, `SearchIndex.from_existing()`, `rvl index info`, `rvl stats` | `FT.INFO` | Yes | **No** |
+| `index.create()` | `FT.CREATE` | Yes | **No** |
+| `index.delete()`, `rvl index delete`, `rvl index destroy` | `FT.DROPINDEX` | Yes | Yes |
+| Enumerating indexes (see below) | `FT._LIST` | **No** | **No** |
+
+Every `Yes` above assumes key patterns that cover the index prefix — see [Key permissions](#key-permissions) — and `-@dangerous` layered on the second column additionally denies `FT.DROPINDEX`, so `index.delete()` becomes `No`. An SVS-VAMANA schema needs more than `FT.CREATE`: `index.create()` first probes capabilities with `INFO` (`@slow @dangerous`) and `MODULE LIST` (`@admin @slow @dangerous`), so both `-@admin` and `-@dangerous` policies break creation for those schemas.
+
+Two of the rows above deserve their own explanation.
+
+`FT._LIST` is tagged `@admin` as well as `@search` and `@slow`, and ACL rules are applied left to right — so a rule that grants search access and then takes back administrative commands, such as `+@search -@admin` or `+@all -@admin`, denies it:
 
 ```text
 User <name> has no permissions to run the 'FT._LIST' command
 ```
 
-Note that `FT._LIST` does not *require* `@admin`: granting `+@search` on its own permits it. Only rules that subtract `@admin` after granting search are affected. To keep such a policy and still enumerate indexes, grant the command back explicitly with `+ft._list`.
+`FT._LIST` does not *require* `@admin`: granting `+@search` on its own permits it. Only rules that subtract `@admin` after granting search are affected. To keep such a policy and still enumerate indexes, grant the command back explicitly with `+ft._list`. Enumeration is reached by `SearchIndex.listall()` and `AsyncSearchIndex.listall()`, by `rvl index listall`, and by the migration entry points that discover indexes for you: `rvl migrate helper`, `rvl migrate wizard` when no `-i/--index` is given, and `rvl migrate batch-plan --pattern`.
 
-| Operation | Redis command | Permitted by `+@all -@admin` |
-|---|---|---|
-| `index.create()`, `index.exists()` | `FT.CREATE`, `FT.INFO` | Yes |
-| `index.query()`, `index.search()`, `index.aggregate()` | `FT.SEARCH`, `FT.AGGREGATE` | Yes |
-| `index.info()`, `rvl index info`, `rvl stats` | `FT.INFO` | Yes |
-| `index.load()` | `HSET` or `JSON.SET` (needs `@write` and key access) | Yes |
-| `index.delete()`, `rvl index delete`, `rvl index destroy` | `FT.DROPINDEX` | Yes |
-| Enumerating indexes (see below) | `FT._LIST` | No |
+`FT.DROPINDEX` is tagged `@dangerous` and `@write` as well as `@search`, so a policy that subtracts `@dangerous` denies `index.delete()`, `rvl index delete`, and `rvl index destroy`.
 
-`SemanticCache`, `SemanticMessageHistory`, `MessageHistory`, and `SemanticRouter` each call `index.create()` while being constructed, so they are covered by the first row.
+### Roles built from `@read` and `@write`
 
-Index enumeration is the only thing an `-@admin` rule breaks. It is reached by `SearchIndex.listall()` and `AsyncSearchIndex.listall()`, by `rvl index listall`, and by the migration entry points that discover indexes for you: `rvl migrate helper`, `rvl migrate wizard` when no `-i/--index` is given, and `rvl migrate batch-plan --pattern`.
+`FT.INFO` is in neither `@read` nor `@write` — its only category is `@search` — and `FT.CREATE` is the same. So an application role assembled from `+@read +@write` — a natural least-privilege shape — can query and load, but cannot ask whether an index exists and cannot create one. Note that subtracting `@dangerous` is not what causes this: `+@all -@dangerous` permits both. The commands are simply never granted.
 
-Other categories gate different operations. `FT.DROPINDEX` is tagged `@dangerous` and `@write` as well as `@search`, so a policy that subtracts `@dangerous` denies `index.delete()`, `rvl index delete`, and `rvl index destroy`.
+Every extension constructor checks whether its index exists, so under such a credential all of them fail while being constructed:
+
+```text
+RedisSearchError: Error while fetching llmcache index info:
+User <name> has no permissions to run the 'FT.INFO' command
+```
+
+### "no permissions to run the 'FT.INFO' command"
+
+RedisVL does not guess its way around this. A credential that cannot ask whether the index exists also cannot create one, so there is nothing useful to infer — instead, tell RedisVL that the index is already there:
+
+```python
+cache = SemanticCache(
+    name="llmcache",
+    redis_url="redis://localhost:6379",
+    create_index=False,
+)
+```
+
+`create_index=False` is available on `SemanticCache`, `MessageHistory`, `SemanticMessageHistory` and `SemanticRouter`. It skips the existence check, the comparison of your schema against the live index, and index creation — the constructor issues no index command at all. Pass it when the index is managed externally, or when the credential cannot run `FT.INFO`. It cannot be combined with `overwrite=True`, which asks for the opposite.
+
+A `SearchIndex` used directly needs nothing special: build it with `from_dict()` or `from_yaml()`, then load and query. Two of its methods stay unavailable, because both read index metadata: `from_existing()`, which reconstructs a schema out of Redis, and `clear()`, which starts by calling `info()`.
+
+The flag also skips the SVS-VAMANA capability probe described above, since that runs inside `create()`.
+
+### Provisioning a router
+
+`SemanticRouter` with `create_index=False` writes nothing at all: not the reference vectors for its routes, and not the stored route config that `SemanticRouter.from_existing()` reads. Preparing a router for this mode therefore means constructing it once with a privileged credential — a hand-written `FT.CREATE` is not enough, because the reference vectors have to be embedded and written too. Without them the router matches nothing, which looks like a distance-threshold problem rather than an empty index.
+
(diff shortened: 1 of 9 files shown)


redisvl/extensions/cache/llm/semantic.py
```diff
@@ -148,32 +192,35 @@ def __init__(
...
+                    raise ValueError(
+                        f"Existing index {name} schema does not match the user provided schema for the semantic cache. "
+                        "If you wish to overwrite the index schema, set overwrite=True during initialization."
+                    )
 
-        # Create the search index in Redis
-        self._index.create(overwrite=self.overwrite, drop=False)
+            # Create the search index in Redis
+            self._index.create(overwrite=self.overwrite, drop=False)
+        else:
```
nkanu17, line 214: **P1 — prevent cross-alias target deletion.** This branch deliberately skips the existence/identity check, so it can now construct an extension whose `name` is an index alias. All four extension `delete()` methods (and cache `adelete()`) later call `FT.DROPINDEX <name> DD`; Redis resolves an alias and drops the separately named target index plus its documents. Normal construction does not expose this path because `SearchIndex.exists()` reports aliases as absent and creation then fails. Please reject lifecycle-wide `delete`/`adelete` when `_create_index` is false unless ownership is separately and explicitly enabled; `clear()` should be considered under the same ownership rule. Add an alias regression for every extension and the async cache path.

redisvl/extensions/router/semantic.py
```diff
@@ -123,7 +156,17 @@ def from_existing(
...
+
+        Reads the stored route config with ``JSON.GET``, so unlike
+        :meth:`SearchIndex.from_existing` this needs no ``FT.INFO``. Pass
+        ``create_index=False`` to keep it that way through construction, which
+        makes this the way to attach to a router with a credential that cannot
+        run index-metadata commands.
+        """
+        # Pulled out before the split below, which retains only SearchIndex init
+        # kwargs and would otherwise hand this to the Redis client constructor.
+        create_index = kwargs.pop("create_index", True)
```
nkanu17, line 170: **P2 — handle `overwrite` before splitting connection kwargs.** `create_index` is popped here, but `overwrite` is not. `from_existing(..., create_index=False, overwrite=True)` therefore bypasses the constructor conflict check: with a supplied client the value is silently ignored, while URL-based construction forwards it to redis-py and raises `TypeError: AbstractConnection.__init__() got an unexpected keyword argument overwrite`. Pop and reject or explicitly forward `overwrite`, and cover both connection forms.

docs/user_guide/installation.md
```diff
@@ -186,30 +186,102 @@ The Sentinel URL format supports:
...
+| Operation | Redis command | `+@all -@admin` | `+@read +@write` |
+|---|---|---|---|
+| `index.query()`, `index.search()`, `index.aggregate()` | `FT.SEARCH`, `FT.AGGREGATE` | Yes | Yes |
+| `index.load()` | `HSET` or `JSON.SET` (needs key access) | Yes | Yes |
+| `index.exists()`, `index.info()`, `index.clear()`, `SearchIndex.from_existing()`, `rvl index info`, `rvl stats` | `FT.INFO` | Yes | **No** |
+| `index.create()` | `FT.CREATE` | Yes | **No** |
+| `index.delete()`, `rvl index delete`, `rvl index destroy` | `FT.DROPINDEX` | Yes | Yes |
+| Enumerating indexes (see below) | `FT._LIST` | **No** | **No** |
+
+Every `Yes` above assumes key patterns that cover the index prefix — see [Key permissions](#key-permissions) — and `-@dangerous` layered on the second column additionally denies `FT.DROPINDEX`, so `index.delete()` becomes `No`. An SVS-VAMANA schema needs more than `FT.CREATE`: `index.create()` first probes capabilities with `INFO` (`@slow @dangerous`) and `MODULE LIST` (`@admin @slow @dangerous`), so both `-@admin` and `-@dangerous` policies break creation for those schemas.
```
nkanu17: **P2 — this overstates key-pattern requirements.** The table includes `index.create()` as a Yes, but the Key permissions section below correctly says `FT.CREATE` is not checked against index-prefix key patterns. Please carve out `FT.CREATE` here so the guide does not imply that Redis prevents creating an index the caller cannot subsequently query.

docs/user_guide/installation.md
```diff
@@ -186,30 +186,102 @@ The Sentinel URL format supports:
...
+
+The flag also skips the SVS-VAMANA capability probe described above, since that runs inside `create()`.
+
+### Provisioning a router
+
+`SemanticRouter` with `create_index=False` writes nothing at all: not the reference vectors for its routes, and not the stored route config that `SemanticRouter.from_existing()` reads. Preparing a router for this mode therefore means constructing it once with a privileged credential — a hand-written `FT.CREATE` is not enough, because the reference vectors have to be embedded and written too. Without them the router matches nothing, which looks like a distance-threshold problem rather than an empty index.
+
+Afterwards, `SemanticRouter.from_existing(name, create_index=False)` is the way to attach to it: it recovers the routes and thresholds with `JSON.GET` and needs no `FT.INFO`. Pass the full route set. Each route's distance threshold is applied from the local list, so a partial set silently narrows matching — and `add_route()` and `remove_route()` rewrite the stored config from that same list, so attaching with a subset and then adding a route drops the rest from the config every other client reads.
```
nkanu17: **P2 — this instruction cannot be followed with this API.** `SemanticRouter.from_existing()` does not accept a route set; it obtains routes from the stored `route_config`. Please say that the stored config must contain the complete route set, or move this warning to direct `SemanticRouter(..., routes=..., create_index=False)` construction.

docs/user_guide/installation.md
```diff
@@ -186,30 +186,102 @@ The Sentinel URL format supports:
...
-| `index.create()`, `index.exists()` | `FT.CREATE`, `FT.INFO` | Yes |
-| `index.query()`, `index.search()`, `index.aggregate()` | `FT.SEARCH`, `FT.AGGREGATE` | Yes |
-| `index.info()`, `rvl index info`, `rvl stats` | `FT.INFO` | Yes |
-| `index.load()` | `HSET` or `JSON.SET` (needs `@write` and key access) | Yes |
-| `index.delete()`, `rvl index delete`, `rvl index destroy` | `FT.DROPINDEX` | Yes |
-| Enumerating indexes (see below) | `FT._LIST` | No |
+### When the schema diverges
 
-`SemanticCache`, `SemanticMessageHistory`, `MessageHistory`, and `SemanticRouter` each call `index.create()` while being constructed, so they are covered by the first row.
+With `create_index=False` nothing verifies that the live index matches the schema you described. Some mismatches are loud on first use, and two are silent:
```
nkanu17: **P2 — the count is incorrect.** The table identifies three silent mismatch classes: prefix, storage type, and datatype/distance metric. Use “three” or “several.”

docs/user_guide/installation.md
```diff
@@ -186,30 +186,102 @@ The Sentinel URL format supports:
...
+`FT._LIST` does not *require* `@admin`: granting `+@search` on its own permits it. Only rules that subtract `@admin` after granting search are affected. To keep such a policy and still enumerate indexes, grant the command back explicitly with `+ft._list`. Enumeration is reached by `SearchIndex.listall()` and `AsyncSearchIndex.listall()`, by `rvl index listall`, and by the migration entry points that discover indexes for you: `rvl migrate helper`, `rvl migrate wizard` when no `-i/--index` is given, and `rvl migrate batch-plan --pattern`.
+
+`FT.DROPINDEX` is tagged `@dangerous` and `@write` as well as `@search`, so a policy that subtracts `@dangerous` denies `index.delete()`, `rvl index delete`, and `rvl index destroy`.
+
+### Roles built from `@read` and `@write`
+
+`FT.INFO` is in neither `@read` nor `@write` — its only category is `@search` — and `FT.CREATE` is the same. So an application role assembled from `+@read +@write` — a natural least-privilege shape — can query and load, but cannot ask whether an index exists and cannot create one. Note that subtracting `@dangerous` is not what causes this: `+@all -@dangerous` permits both. The commands are simply never granted.
```
nkanu17: **P2 — avoid calling the bare role least privilege.** This same guide shows that `+@read +@write` grants `FT.DROPINDEX`, so users can copy this “natural least-privilege” shape while retaining a destructive command. Recommend `+@read +@write -@dangerous` for the runtime example, with explicit grants for anything additionally required.

nkanu17 (changes requested):
Requesting changes for the alias-target deletion path introduced by create_index=False. Normal construction uses SearchIndex.exists() to reject aliases, but this mode skips that identity guard while leaving extension-wide delete operations armed. Please fail closed for lifecycle-wide destructive operations when the index is externally managed, or require a separate explicit ownership/force opt-in, and add alias regression coverage across all four extensions plus async cache deletion. The remaining inline comments are non-blocking correctness and documentation fixes.

redisvl/extensions/router/semantic.py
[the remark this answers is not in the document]
nkanu17: Resolved in 46459d6. `from_existing()` now removes `overwrite` before splitting connection kwargs, rejects `create_index=False, overwrite=True` immediately, and forwards valid overwrite values to construction. Regression coverage checks both supplied-client and URL-only forms, preventing the prior silent ignore and redis-py `TypeError`.

redisvl/extensions/cache/llm/semantic.py
```diff
@@ -148,32 +192,35 @@ def __init__(
...
+                if existing_index.schema.to_dict() != self._index.schema.to_dict():
+                    raise ValueError(
+                        f"Existing index {name} schema does not match the user provided schema for the semantic cache. "
+                        "If you wish to overwrite the index schema, set overwrite=True during initialization."
+                    )
 
-        # Create the search index in Redis
-        self._index.create(overwrite=self.overwrite, drop=False)
+            # Create the search index in Redis
+            self._index.create(overwrite=self.overwrite, drop=False)
```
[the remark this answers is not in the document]
nkanu17: Resolved in 46459d6. Extensions constructed with `create_index=False` now reject index-wide `delete()`, `adelete()`, `clear()`, and `aclear()` before issuing any Redis command. The regression suite covers both lifecycle operations across all four extensions and both async cache paths using an alias-shaped name; the focused suite passes 29/29.

nkanu17 (approved):

PR #689, opened 2026-08-13
fix(ci): make the redis-py matrix axis actually test redis-py

vishal-bala:
## Problem

The `redis-py-version` axis of the test matrix was inert. All 45 jobs ran redis-py 7.4.0, regardless of the version in their name.

The job pinned a version with `uv pip install "redis>=6,<7"`, then ran `make test`, which reaches pytest through `uv run`. `uv run` re-syncs the environment to `uv.lock` before executing, silently reverting the pin.

The signature is visible in [run 31662019515](https://github.com/redis/redis-vl-python/actions/runs/31662019515): the 5.x, 6.x and 7.x cells for the same Python and Redis image returned byte-identical results — `2070 passed, 143 skipped, 2 xfailed, 210 warnings` — which cannot happen when 48 tests are gated on `redis-py >= 7.1.0`.

The other two axes were unaffected. `REDIS_IMAGE` is read by `tests/docker-compose.yml` and the Python version comes from `setup-python`; neither passes through uv.

## Fix

- `UV_NO_SYNC: "1"` on the `test` job, so `uv run` uses the environment as installed. Scoped to that job because it is the only one that pins a redis-py version. `make test` is unchanged, so local runs keep auto-sync.
- The install `if/elif/else` becomes a `case` that fails on an unhandled matrix value. The previous `else` was a catch-all installing 7.x for anything unrecognised, so a typo passed for the wrong reason.
- Each cell logs the redis-py version it resolved.
- `--frozen` on `uv sync --all-extras`, matching `lint.yml` and the `docs` job.
- Notebook validation moves into its own `notebooks` job. It was two steps constrained to one matrix cell by duplicated `if:` conditions, so any matrix rename would have silently disabled it.

## Dropping redis-py 5.x

The matrix goes from 45 to 30 cells. redis-py's `VectorField` rejects `SVS-VAMANA` before 6.3.0, so `tests/unit/test_fields.py` fails on 5.x — 10 failures on 5.3.1 locally, all in that file.

We're adding redis-py 8.x support shortly and view dropping 5.x as part of that transition, so removing the cells is preferable to gating the tests around a version we're retiring.

`pyproject.toml` still declares `redis>=5.0,<8.0`; that range belongs to the 8.x work and isn't touched here. Note the effective floor is **6.3.0**, not 6.0 — the `6.x` cells pass only because the range resolves to the latest 6.x, and 6.0–6.2 remain untested.

## Notebook job

Cell 20 of `04_vectorizers.ipynb` is now `NBVAL_SKIP`, which let the Google Cloud auth step and the `GOOGLE_CREDENTIALS` / `GCP_LOCATION` / `GCP_PROJECT_ID` secrets come out of the job. That cell made a live Vertex AI call through `VertexAIVectorizer` — deprecated in code — while its supported replacement (`GoogleGenAIVectorizer`, cell 22) was already skipped. The VertexAI integration tests still run in `service-tests` under `--run-api-tests`, so no coverage is lost.

The job now waits for Redis to accept connections before running nbval. `docker run -d` returns before the server is ready, and the SVS probe in `make test-notebooks` suppresses errors, so an unready server was indistinguishable from "Redis < 8.2" and silently dropped `09_svs_vamana.ipynb`.

Two behavioural changes: notebooks are no longer gated on `make test` passing, since they now run in parallel with the matrix; and the job pins no redis-py version, because notebooks only need to work against a default install.

## Verification

All 33 jobs pass, including the 15 `6.x` cells that had never actually executed. Every cell logs the major it advertises — `6.x` cells resolve to 6.4.0, `7.x` to 7.4.0.

The skip counts now diverge by axis, with no overlap:

| redis-py | `redis:8.2` | `redis:8.4` / `latest` |
|---|---|---|
| 6.x | 1896 passed / 199 skipped | 1895 / 200 |
| 7.x | 1936 / 159 | 1952 / 143 |

(Python 3.10–3.13; 3.14 skips ~114 more via the existing HF gate.) If 6.x and 7.x ever report the same counts again, the axis has regressed.

Two useful corroborations: `7.x [redis:8.4]` on 3.11 reproduces `2070 passed / 143 skipped` — exactly the figure every cell used to report — and `6.x [redis:8.4]` matches a local run at 6.4.0 to the test.

**Cost went down, not up.** Wall clock 9.5 → 9.2 min; total compute 153 → 110 job-minutes across 47 → 33 jobs. Per-cell time is unchanged, because the old cells were already running the full suite, just at the wrong version. The critical path is now `Service Tests` (3.8 min) plus the slowest cell (5.2 min); moving nbval into a parallel job took ~2.3 min off the longest cell.

## Follow-ups

- Raise the `pyproject.toml` floor to `redis>=6.3.0` (part of the 8.x work).
- `Makefile:73`'s SVS probe hides errors with `2>/dev/null`, so local `make test-notebooks` can still drop a notebook silently.
- No lockfile freshness check in CI. `--locked` isn't a drop-in: it fails on release-bump commits that don't relock.
- Fork PRs run no unit tests — `test` and `notebooks` both `needs: service-tests`, which is fork-gated. The `test` job now uses no secrets, which is the precondition for changing that.
- `google-github-actions/auth@v1` is EOL, and action versions drift between workflows.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -60,11 +60,16 @@ jobs:
     needs: service-tests
     env:
       HF_HOME: ${{ github.workspace }}/hf_cache
+      REDIS_IMAGE: ${{ matrix.redis-image }}
+      REDIS_PY_VERSION: ${{ matrix.redis-py-version }}
+      # Required: without it `uv run` re-syncs the venv to uv.lock and reverts
+      # the redis-py pin set below.
+      UV_NO_SYNC: "1"
     strategy:
       fail-fast: false
       matrix:
         python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
-        redis-py-version: ["5.x", "6.x", "7.x"]
+        redis-py-version: ["6.x", "7.x"]
         redis-image: ["redis:8.2", "redis:8.4", "redis:latest"]
     steps:
       - name: Check out repository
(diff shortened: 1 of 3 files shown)


memburra (approved):
lgtm!

nkanu17 (approved):

PR #692, opened 2026-08-16
feat(voyageai): add contextualized embeddings and refresh model support

fzowl:
## What

Refreshes the VoyageAI integration against the current VoyageAI catalog (text, contextualized, multimodal, reranker).

- **Contextualized embeddings**: `voyage-context-*` models (e.g. `voyage-context-4`) route to VoyageAI's contextualized embeddings API. `VoyageAIVectorizer` accepts a flat `list[str]` of chunks and calls `contextualized_embed` with `enable_auto_chunking=True` and `chunk_size=32000`, so each input document resolves to a single chunk and `embed_many` keeps its one-embedding-per-input contract (and cache alignment).
- **Multimodal `embed_many` fix**: previously the multimodal branch wrapped the whole batch as `[batch]`, which VoyageAI treats as a single multimodal input with multiple parts — collapsing N requested contents into one embedding. Each content is now sent as its own input (`[[item] for item in batch]`), so the returned count matches the request count.
- **Current model catalog**: `_get_batch_size` now routes the `voyage-4` family into the same token-limit tiers as their `voyage-3.5` counterparts (`voyage-4-lite` → 30, `voyage-4` → 10); `voyage-4-large`, `voyage-code-4` and `voyage-context-*` keep the conservative default, matching their 120K-token tier. Every VoyageAI model id already works because the vectorizer passes `model` strings straight through — this refreshes the throughput heuristic and documents the current families (`voyage-4-*`, `voyage-code-4`, `voyage-context-*`, `voyage-multimodal-*`) on the class docstring.
- **Dependency**: bump the `voyageai` floor to `>=0.5.0` (latest), where contextualized auto-chunking (`enable_auto_chunking`, `chunk_size`) and the `voyage-4` series are available.
- **Docs/examples**: refresh notebook examples to current models — `rerank-2.5` for the reranker and a `voyage-context-4` contextualized example for the vectorizer guide.

## Why

Contextualized (`voyage-context-*`) chunk embeddings were not exposed, current models weren't reflected in the docs or batch heuristics, and the multimodal batching bug meant `embed_many(["a", "b"])` on a multimodal model could return a single embedding.

## Notes

- Adds mocked (no-API-key) tests covering context-model routing, auto-chunking args, async parity, the multimodal no-collapse fix, and the per-model batch-size tiers; these run under `make test`.
- `contextualized_embed` calls carry `# type: ignore[attr-defined]` so type-checking works without vendored stubs; the mypy config uses `ignore_missing_imports` and does not set `warn_unused_ignores`, so the ignores stay valid whether or not the client is typed.
- API-dependent tests remain gated behind `requires_api_keys`.
- **Housekeeping**: `uv.lock` shows marker churn beyond the `voyageai` bump (nvidia/grpcio/etc. environment markers) because it was regenerated with a newer uv resolver than the one that produced the base lock. It pins `voyageai==0.5.0` and is self-consistent, and CI installs via `uv sync --frozen`. Happy to regenerate it against the repo's pinned uv (0.12.3) for a minimal lock diff if preferred.
- **Validation**: `mypy redisvl/utils/vectorize/voyageai.py` is clean, `isort --profile black` + `black --target-version py311` report no changes across `redisvl/` and `tests/`, and the mocked VoyageAI routing/batch-size tests pass against `voyageai==0.5.0`. The full `pytest` invocation pulls in a session-scoped, Docker-backed Redis fixture that cannot start in this environment, so the mocked tests were also executed standalone (bypassing that fixture) and pass end-to-end. GitHub Actions is not enabled on this fork, so upstream CI / `make check` should still be the gate on merge.

---

> [!NOTE]
> **Medium Risk**
> Changes embedding request routing, batching, and multimodal input shape, plus a breaking voyageai floor bump to 0.5.0. Incorrect routing or wrapping can change embedding counts or cache keys.
> 
> **Overview**
> Adds `voyage-context-*` support on `VoyageAIVectorizer` by routing those models to `contextualized_embed`. Each input is treated as its own auto-chunked document (queries skip chunking) so `embed_many` still returns one embedding per string.
> 
> Fixes multimodal `embed_many` wrapping: each item is now `[[item] for item in batch]` instead of collapsing the whole batch into one multimodal input.
> 
> Plain-text embeds use token-aware batching against per-model token limits, with a fallback to item-count batches if the tokenizer is unavailable. Batch-size tiers now include the `voyage-4` family.
> 
> Bumps the extra to `voyageai[video]>=0.5.0` (ffmpeg-python is optional in 0.5). Docs add a context-model example and switch the reranker demo to `rerank-2.5`. Mocked tests cover routing, auto-chunking, multimodal counts, and batching.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f625ef074d172230251f4501cc2d0c83bc5cb268. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/user_guide/04_vectorizers.ipynb
+++ b/docs/user_guide/04_vectorizers.ipynb
@@ -600,6 +600,39 @@
       "cell_type": "markdown",
       "metadata": {},
       "source": [
+        "#### Contextualized embeddings\n",
+        "\n",
+        "VoyageAI's `voyage-context-*` models produce *contextualized* chunk embeddings: each chunk is embedded with awareness of the other chunks in the same request, which improves retrieval quality for chunked documents. The `VoyageAIVectorizer` automatically routes `voyage-context-*` models to the contextualized embeddings API — just pass your list of chunks to `embed_many` and you get one embedding back per chunk."
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "# Contextualized embeddings (voyage-context-* models) embed each chunk with\n",
+        "# awareness of the surrounding chunks. Pass a list of chunk strings and the\n",
+        "# vectorizer returns one embedding per chunk. Requires voyageai>=0.5.0.\n",
+        "context_vo = VoyageAIVectorizer(\n",
+        "    model=\"voyage-context-4\",  # See https://docs.voyageai.com/docs/contextualized-chunk-embeddings\n",
+        "    api_config={\"api_key\": api_key},\n",
+        ")\n",
+        "\n",
+        "chunks = [\n",
+        "    \"That is a happy dog\",\n",
+        "    \"That is a happy person\",\n",
+        "    \"Today is a sunny day\",\n",
+        "]\n",
+        "context_embeddings = context_vo.embed_many(chunks, input_type=\"document\")\n",
+        "print(\"Number of embeddings:\", len(context_embeddings))\n",
+        "print(\"Vector dimensions:\", len(context_embeddings[0]))"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
         "### Mistral AI\n",
         "\n",
         "[Mistral](https://console.mistral.ai/) offers LLM and embedding APIs for you to implement into your product. The `MistralAITextVectorizer` makes it simple to use RedisVL with their embeddings model.\n",
--- a/docs/user_guide/04_vectorizers.ipynb
+++ b/docs/user_guide/04_vectorizers.ipynb
@@ -602,7 +602,7 @@
       "source": [
         "#### Contextualized embeddings\n",
         "\n",
-        "VoyageAI's `voyage-context-*` models produce *contextualized* chunk embeddings: each chunk is embedded with awareness of the other chunks in the same request, which improves retrieval quality for chunked documents. The `VoyageAIVectorizer` automatically routes `voyage-context-*` models to the contextualized embeddings API — just pass your list of chunks to `embed_many` and you get one embedding back per chunk."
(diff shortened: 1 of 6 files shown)


vishal-bala (commented):
Code looks good! I think service tests are failing due to an issue we've since resolved on `main` - if you can rebase from `origin`, we should be good to go.

fzowl:
Thanks for the review! I've brought the branch up to date with the latest `main` (merged it in rather than rebasing, since the branch is already published), which pulls in the fix you mentioned. Service Tests should be green now.

fzowl:
@vishal-bala one thing worth a look before merge, since the service-test rerun won't surface it: with a `voyage-context-*` model, `_context_embed_kwargs` only enables auto-chunking when `input_type == "document"`, and `SemanticRouter` / `SemanticCache` call `embed()` / `embed_many()` without an `input_type`. That sends a flat `list[str]` to `contextualized_embed` with `enable_auto_chunking=False` (the case Bugbot flagged as high-severity), and it isn't covered by the mocked tests — only `document` and `query` are. A quick test for the omitted-`input_type` path (or defaulting it) would confirm context models actually work through those extensions.

Separately, the new token-aware batching calls `Client.tokenize()` at init (via the dims probe) and on every plain-text embed, which pulls a HuggingFace tokenizer — so init can now fail offline, or for model ids without a HF tokenizer, for all VoyageAI users rather than just context models. A fallback to item-count batching when `tokenize` isn't available would preserve the previous init behavior.

Also note the branch was merged from `main` rather than rebased, so the diff is larger than the actual change set.

fzowl:
@vishal-bala thanks! The branch is already up to date with the latest `main` — the current `main` HEAD (`0f64097`) is merged in (`git merge-base --is-ancestor upstream/main HEAD` passes), so `git diff main...HEAD` is limited to the 6 intended files (vectorizer, tests, two notebooks, `pyproject.toml`, `uv.lock`). I brought it current via merge rather than rebase since the branch is already published.

I also just pushed one small robustness fix on top: token-aware batching relied on `Client.tokenize()` (which loads a HuggingFace tokenizer), which had made the HF Hub a hard dependency of `__init__` and every plain-text `embed_many`. It now falls back to item-count batching if the tokenizer is unavailable, with sync/async + init-resilience tests. That should also re-trigger the service tests.

vishal-bala (approved):

PR #694, opened 2026-08-17
Fix reranked scores not matching reranked document order in HFCrossEncoderReranker

### LINKED ISSUE #693: `HFCrossEncoderReranker` returns scores in original input order after reranking
Codex found this bug while I was making a plan - double checked with Grok and it found the same bug.

---

# `HFCrossEncoderReranker` returns scores in original input order after reranking

## Description

`HFCrossEncoderReranker.rank(..., return_score=True)` returns reranked documents in descending score order, but returns the accompanying score array in the original input order.

As a result, the returned documents and scores are misaligned:

```python
reranked_docs[i]
```

does not necessarily correspond to:

```python
scores[i]
```

This affects RedisVL 0.23.0 and appears to remain present in current releases/main.

## Reproduction

Given several documents whose cross-encoder scores are not already in descending input order:

```python
from redisvl.utils.rerank import HFCrossEncoderReranker

reranker = HFCrossEncoderReranker(
    model="cross-encoder/ms-marco-MiniLM-L-6-v2",
)

docs = [
    "irrelevant document",
    "highly relevant document",
    "somewhat relevant document",
]

reranked_docs, scores = reranker.rank(
    query="relevant document",
    docs=docs,
    limit=len(docs),
    return_score=True,
)

for doc, score in zip(reranked_docs, scores):
    print(score, doc)
```

The documents are returned in reranked order, but the scores remain ordered according to the original `docs` input.

## Cause

The implementation first associates documents with their scores and correctly sorts them:

```python
scores = self._client.predict(...)
scores = [float(score) for score in scores]

docs_with_scores = list(zip(doc_subset, scores))
docs_with_scores.sort(key=lambda x: x[1], reverse=True)

reranked_docs = [doc for doc, _ in docs_with_scores[:limit]]
```

However, the returned scores are then taken from the original unsorted `scores` list:

```python
scores = scores[:limit]

if return_score:
    return reranked_docs, scores
```

Therefore only the documents reflect the sort operation.

## Expected behavior

Documents and scores should be returned from the same sorted sequence, such that:

```python
reranked_docs[i]
```

always corresponds to:

```python
scores[i]
```

and the scores are ordered consistently with the reranked documents.

## Proposed fix

Extract both documents and scores from the sorted `docs_with_scores` list:

```python
ranked = docs_with_scores[:limit]

reranked_docs = [doc for doc, _ in ranked]
scores = [score for _, score in ranked]

if return_score:
    return reranked_docs, scores
```

Alternatively:

```python
reranked_docs = [doc for doc, _ in docs_with_scores[:limit]]
scores = [score for _, score in docs_with_scores[:limit]]
```

## Impact

Consumers that attach the returned reranker score to each result will associate incorrect scores with documents whenever the reranker changes the input ordering.

Using `return_score=False` avoids exposing the incorrect association, but does not solve the issue for applications that need reranker scores for downstream ranking, observability, debugging, or response metadata.

## Related behavior

`HFCrossEncoderReranker` also defaults to:

```python
limit=3
```

so callers that intend to rerank and return their full candidate set must explicitly provide an appropriate `limit`, such as `limit=len(candidates)`.

That appears to be intentional API behavior rather than part of this bug.

adityaanikam:
Fixes #693.

HFCrossEncoderReranker.rank() sorted docs_with_scores by score to build reranked_docs, but then took scores from the original unsorted scores list instead of from the same sorted sequence. So reranked_docs[i] and scores[i] didn't correspond to the same document once the reranking actually changed the order. arank() delegates to rank(), so this fixes both.

Added a regression test that asserts the returned scores are sorted descending, matching the returned docs. Confirmed it fails with the old code (scores come back in the original order, not sorted) and passes with the fix, using a real cross-encoder model against the exact reproduction from the issue.

---

> [!NOTE]
> **Low Risk**
> Small, localized bugfix in reranker output pairing with a regression test; no auth, data, or API contract changes beyond correcting score alignment.
> 
> **Overview**
> **Fixes misaligned scores in `HFCrossEncoderReranker.rank()`** when reranking changes document order (#693).
> 
> After sorting `(doc, score)` pairs by score, the method now builds both `reranked_docs` and the returned `scores` from that same sorted slice. Previously it sliced the sorted docs but still returned scores from the original prediction order, so index `i` in docs and scores could refer to different hits.
> 
> Adds an integration regression test that checks returned scores are in descending order when `return_score` is true (the failure mode from the old code).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 919bddc4a9463a36bab90c73aa141e96ea57b16c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/utils/rerank/hf_cross_encoder.py
+++ b/redisvl/utils/rerank/hf_cross_encoder.py
@@ -120,8 +120,8 @@ class HFCrossEncoderReranker(BaseReranker):
         scores = [float(score) for score in scores]
         docs_with_scores = list(zip(doc_subset, scores))
         docs_with_scores.sort(key=lambda x: x[1], reverse=True)
-        reranked_docs = [doc for doc, _ in docs_with_scores[:limit]]
-        scores = [score for _, score in docs_with_scores[:limit]]
+        reranked_docs_tuple, scores_tuple = zip(*docs_with_scores[:limit])
+        reranked_docs, scores = list(reranked_docs_tuple), list(scores_tuple)
 
         if return_score:
             return reranked_docs, scores  # type: ignore
--- a/redisvl/utils/rerank/hf_cross_encoder.py
+++ b/redisvl/utils/rerank/hf_cross_encoder.py
@@ -121,7 +121,7 @@ class HFCrossEncoderReranker(BaseReranker):
         docs_with_scores = list(zip(doc_subset, scores))
         docs_with_scores.sort(key=lambda x: x[1], reverse=True)
         reranked_docs = [doc for doc, _ in docs_with_scores[:limit]]
-        scores = scores[:limit]
+        scores = [score for _, score in docs_with_scores[:limit]]
 
         if return_score:
             return reranked_docs, scores  # type: ignore
--- a/tests/integration/test_cross_encoder_reranker.py
+++ b/tests/integration/test_cross_encoder_reranker.py
@@ -33,6 +33,23 @@ async def test_async_rank_documents(reranker):
     assert all(isinstance(score, float) for score in scores)
 
 
+def test_rank_scores_align_with_reranked_docs(reranker):
+    # https://github.com/redis/redis-vl-python/issues/693
+    docs = [
+        "irrelevant document",
+        "highly relevant document",
+        "somewhat relevant document",
+    ]
+    query = "relevant document"
+
+    reranked_docs, scores = reranker.rank(query, docs, limit=len(docs))
+
+    # reranked_docs is sorted by score descending, so scores must be too,
+    # this catches the case where scores are taken from the original
+    # unsorted list instead of the sorted one.
+    assert scores == sorted(scores, reverse=True)
+
+
 def test_bad_input(reranker):
     with pytest.raises(ValueError):
         reranker.rank("", [])  # Empty query
--- a/tests/integration/test_cross_encoder_reranker.py
+++ b/tests/integration/test_cross_encoder_reranker.py
@@ -47,7 +47,7 @@ def test_rank_scores_align_with_reranked_docs(reranker):
     # reranked_docs is sorted by score descending, so scores must be too,
     # this catches the case where scores are taken from the original
     # unsorted list instead of the sorted one.
-    assert scores == sorted(scores, reverse=True)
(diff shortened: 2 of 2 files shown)


redisvl/utils/rerank/hf_cross_encoder.py
```diff
@@ -121,7 +121,7 @@ def rank(
         docs_with_scores = list(zip(doc_subset, scores))
         docs_with_scores.sort(key=lambda x: x[1], reverse=True)
         reranked_docs = [doc for doc, _ in docs_with_scores[:limit]]
-        scores = scores[:limit]
+        scores = [score for _, score in docs_with_scores[:limit]]
```
vishal-bala: Good catch! I'd prefer to use one small optimization here since right now we'd be iterating over `docs_with_scores` twice just to split up its tuples. We can use the splat operator to unzip the list of tuples tidily 

```python
scores = self._client.predict([(query, text) for text in texts])
scores = [float(score) for score in scores]
docs_with_scores = list(zip(doc_subset, scores))
docs_with_scores.sort(key=lambda x: x[1], reverse=True)
reranked_docs, scores = zip(*docs_with_scores[:limit])
```

vishal-bala (commented):
Thanks for the fix! Just one very minor optimization and I think we're good to go

redisvl/utils/rerank/hf_cross_encoder.py
adityaanikam: Applied it. Had to add a list() conversion after the unzip since zip gives tuples but rank's return type is List and two existing tests check isinstance(reranked_docs, list). That tripped mypy too, reassigning the same names from tuple to list mid scope isn't allowed, so I used separate variable names for the tuple result first. Still one pass overall like you suggested.

vishal-bala:
Hi, this looks good! You just need to rebase onto `origin/main` to pull in some CD/CD-related fixes we made recently.

adityaanikam:
> Hi, this looks good! You just need to rebase onto `origin/main` to pull in some CD/CD-related fixes we made recently.

done

vishal-bala (approved):

PR #697, opened 2026-08-20
fix: EmbeddingsCache batch writes silently dropped on async Redis Cluster

vishal-bala:
## Problem

On an async Redis Cluster client, `EmbeddingsCache.amset()` returned every cache key while writing **nothing**. Callers had no way to tell — no error, no partial result.

Queueing a command on a redis-py async pipeline is synchronous and returns the pipeline itself. So `await pipeline.hset(...)` awaits the pipeline, which calls `ClusterPipeline.__await__` → `initialize()` → `self._command_queue = []`. The queue is wiped, `execute()` returns `[]`, and nothing is ever sent.

Standalone pipelines tolerate the `await` (their `__await__` just returns `self`), which is why this was cluster-only and went unnoticed.

The same pattern in `amexists_by_keys()` made it return `[]` instead of one bool per key.

## Fix

Drop the `await` on the two queueing calls; only `execute()` is awaited.

## Also included

- **TTL is now pipelined with the write** in `mset`/`amset`. TTLs were applied in a sequential loop after `execute()` — N extra round trips, plus a window where a crash left entries with no expiry at all. `HSET` alone does not touch a key's TTL, so nothing would ever have reclaimed them.
- **Stale test kwargs fixed.** Two tests in the cluster file passed `text=`, renamed to `content=` in #452, so they had been raising `TypeError` for ~8 months. `requires_cluster` tests run in no CI job, so nobody saw it.
- Notebook prose named the same renamed parameter; `CONTRIBUTING.md` now documents how to run the cluster tests.

## Testing

Verified against a 3-primary Redis 8.4 cluster: pre-fix `scan_iter` finds **0 of 10** written keys, post-fix **10**. TTLs confirmed for explicit, cache-default, and no-TTL cases, and refreshed on rewrite.

The new unit test is hermetic and runs in default CI — deliberately, since the cluster suite does not. It uses a fake pipeline rather than redis-py internals, because CI matrixes redis-py 6.x/7.x/8.x and the queue attribute differs between them. Both it and the cluster test fail when the `await` is restored.

**This unit test is the only guard against a regression.** An earlier revision of this description claimed that removing a bare `# type: ignore` made `make check-types` catch a reintroduced `await`. That was measured against redis-py 7.4.0 locally and is wrong for the version CI uses: under the locked redis-py 8.1.0 the `await` produces no mypy diagnostic at all. Those ignores are also still required under 8.1.0, where `hset` types `mapping` strictly, so they are restored — scoped as `# type: ignore[arg-type]` rather than bare, so the suppression names what it covers.

Net zero new cluster tests: the batch coverage folds into the existing async cluster test.

## Follow-up (not in this PR)

No CI job passes `--run-cluster-tests`, so ~17 cluster tests have never run. A single non-blocking job would close that gap; `tests/cluster-compose.yml` already exists for it.


limjoobin (approved):
LGTM, this PR fixes a bug causing silent data loss on an async Redis Cluster client.

PR #700, opened 2026-08-24
fix: replace shared mutable connection_kwargs defaults with None sentinels

harshadkhetpal:
## Summary

All five extension constructors declare `connection_kwargs: dict[str, Any] = {}` (ruff B006). Python evaluates defaults once at import, so a single dict object is shared by every cache/history instance created without explicit kwargs — a latent cross-instance state-leak hazard for anything that mutates or stores it:

- `extensions/cache/base.py` (`BaseCache`)
- `extensions/cache/embeddings/embeddings.py`
- `extensions/cache/llm/semantic.py`
- `extensions/message_history/message_history.py`
- `extensions/message_history/semantic_history.py`

Each becomes `dict[str, Any] | None = None` with `connection_kwargs = connection_kwargs or {}` as the first statement after the docstring, so every existing caller — including ones passing `None` or `{}` explicitly — behaves identically.

Also included: `type(top_k) != int` → `type(top_k) is not int` in `message_history.py` (E721; identity is the intended semantics, and this preserves the exact-type check that rejects `bool`).

## Testing
`python -m py_compile` passes on all five files; `ruff check --select B006,E721` on the touched files goes clean. Scoped deliberately to the extension constructors — other B006 instances elsewhere in the repo are left for a follow-up if maintainers want them.

---

> [!NOTE]
> **Low Risk**
> Constructor default and type-check cleanup only; behavior for existing callers is unchanged. No auth, data-path, or Redis protocol changes.
> 
> **Overview**
> Stops cache and message-history constructors from sharing one default `connection_kwargs` dict (B006). `BaseCache`, `EmbeddingsCache`, `SemanticCache`, `MessageHistory`, and `SemanticMessageHistory` now take `None` and assign `connection_kwargs = connection_kwargs or {}` so each instance gets its own dict.
> 
> Also switches `MessageHistory.get_recent` from `type(top_k) != int` to `type(top_k) is not int` (E721), keeping the exact-type check that rejects `bool`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6039d82c81241792252ed0d05b940d5a19a31a3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/extensions/cache/base.py
+++ b/redisvl/extensions/cache/base.py
@@ -31,7 +31,7 @@ class BaseCache:
         redis_client: SyncRedisClient | None = None,
         async_redis_client: AsyncRedisClient | None = None,
         redis_url: str = "redis://localhost:6379",
-        connection_kwargs: dict[str, Any] = {},
+        connection_kwargs: dict[str, Any] | None = None,
     ):
         """Initialize a base cache.
 
@@ -45,6 +45,7 @@ class BaseCache:
             connection_kwargs (Dict[str, Any]): The connection arguments
                 for the redis client. Defaults to empty {}.
         """
+        connection_kwargs = connection_kwargs or {}
         self.name = name
         self._ttl: int | None = None
         self.set_ttl(ttl)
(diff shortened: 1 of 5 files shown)


vishal-bala (approved):
LGTM! Thanks for cleaning up this code 👍

PR #709, opened 2026-08-31
fix: use whitespace intersection for filtered text queries

### LINKED ISSUE #708: Filtered TextQuery and AggregateHybridQuery treat literal AND as a required search term when stopwords are disabled
### Summary
`TextQuery` and `AggregateHybridQuery` join full-text queries with `filter_expression` using a literal `"AND"` in the following places:

- `TextQuery._build_query_string()` in `redisvl/query/query.py`

- `FullTextQueryHelper.build_query_string()` in `redisvl/utils/full_text_query_helper.py`, which is used by `AggregateHybridQuery` and the native `HybridQuery`

For example, `TextQuery` builds the final query with:
```
if filter_expression and filter_expression != "*":
    text += f" AND {filter_expression}"
```
Hence, given a generated full-text clause "`<text-query>`" and a filter clause "`<filter-expression>`", the final query is therefore constructed as:
```
<text-query> AND <filter-expression>
```
instead of a query compatible with Redis Search (that uses whitespace to intersect the two clauses):
```
(<text-query>) (<filter-expression>)
```

Redis Search does not use `AND` for intersection, and instead intersects expressions by separating them with whitespace. As a result, Redis parses `AND` as an ordinary, fieldless full-text search term. 

This bug is normally hidden because `and` is part of Redis Search’s default stopword list and is discarded during query parsing. However, it becomes visible when an index is created with either:

- `STOPWORDS 0`, represented in RedisVL by `stopwords=[]`
- A custom stopword list that does not include `"and"`.

Under these configurations, `"AND"` becomes an unintended required term in filtered full text queries, causing these queries to commonly return no results even when documents satisfy both the text query and the supplied filter.

### Reproduction
Start Redis:
```
docker run -d --rm --name redisvl-repro -p 6379:6379 redis:8.4
```
Run the following using RedisVL:
```
from redisvl.query import TextQuery
from redisvl.query.filter import Tag

index = SearchIndex.from_dict(
    {
        "index": {
            "name": "repro",
            "prefix": "repro",
            "stopwords": [],  # Emit STOPWORDS 0
        },
        "fields": [
            {"name": "text", "type": "text"},
            {"name": "team", "type": "tag"},
        ],
    },
    redis_url="redis://localhost:6379",
)

index.create()
index.load([{"text": "the quick brown fox", "team": "a"}])

unfiltered = TextQuery(
    text="fox",
    text_field_name="text",
    stopwords=None,
)

filtered = TextQuery(
    text="fox",
    text_field_name="text",
    stopwords=None,
    filter_expression=Tag("team") == "a",
)

print(index.query(unfiltered))
print(str(filtered))
print(index.query(filtered))
```
#### Observed output:
```
[<matching document>]
@text:(fox) AND @team:{a} ...
[]
```
The unfiltered query matches, but the filtered query does not.
#### Expected Behaviour
The equivalent query using Redis Search’s intersection syntax works and returns the document as expected:
```
redis-cli FT.SEARCH repro '(@text:(fox)) (@team:{a})' DIALECT 2
```
However, the query generated by RedisVL returns nothing:
```
redis-cli FT.SEARCH repro '@text:(fox) AND @team:{a}' DIALECT 2
```

Using `FT.EXPLAIN` to validate the inclusion of `"and"` in the query plan on the `STOPWORDS 0` index:
```
redis-cli FT.EXPLAIN repro '@text:(fox) AND @team:{a}' DIALECT 2
```
The query plan includes `"and"` as a required term, producing:
```
INTERSECT {
  @text:UNION {
    @text:fox
    @text:+fox(expanded)
  }
  UNION {
    and
    +and(expanded)
  }
  TAG:@team {
    a
  }
}
```

### Root cause
There are two query-construction paths that append the literal token.

- `redisvl/query/query.py:1572` - `TextQuery._build_query_string()`:
```
if filter_expression and filter_expression != "*":
    text += f" AND {filter_expression}"
```

- `redisvl/utils/full_text_query_helper.py:65` - `FullTextQueryHelper.build_query_string()`, used by `AggregateHybridQuery` and `HybridQuery`:
```
if filter_expression and filter_expression != "*":
    query += f" AND {filter_expression}"
```

### Suggested fix
Join the expressions using whitespace instead of `"AND"`, making the intended intersection and operator precedence explicit. For example:

- In `TextQuery._build_query_string()`:
```
text += f" ({filter_expression})"
```

- For `FullTextQueryHelper.build_query_string()`:
```
query += f" ({filter_expression})"
```

limjoobin:
Fixes #708.

### Summary
When an index is created without including `"and"` as a stopword, the literal `"AND"` is included in the full-text query generated by `TextQuery`, `AggregateHybridQuery` and `HybridQuery`. This causes Redis Search to parse `AND` as an ordinary, fieldless full-text search term, returning no results even when documents satisfy both the text query and the supplied filter.

### Fixes
This PR modifies `TextQuery`, `AggregateHybridQuery` and `HybridQuery` to use whitespace instead of `"AND"` to join full-text queries and filter expressions.

### Tests
Added regression tests covering filtered text and aggregate hybrid queries on an index created with stopwords disabled.

---

> [!NOTE]
> **Medium Risk**
> Changes Redis Search query syntax for all filtered text/hybrid/multi-vector paths; behavior should match Redis intersection semantics but any caller depending on the old `AND` string could differ, especially on indexes without default stopwords.
> 
> **Overview**
> Fixes filtered full-text and hybrid searches on indexes where **stopwords are disabled** (e.g. `STOPWORDS 0`). Generated query strings no longer insert the literal token `AND` between the text clause and a filter; they use **whitespace intersection** and parenthesize the filter instead (e.g. `@text:(terms) (@team:{docs})`).
> 
> **`TextQuery`**, **`FullTextQueryHelper`** (shared by **`AggregateHybridQuery`** and **`HybridQuery`**), and **`MultiVectorQuery`** are updated consistently. **`MultiVectorQuery`** also joins multiple vector range clauses with spaces rather than `AND`.
> 
> Integration tests cover filtered **`TextQuery`**, raw-string union filters (operator precedence with `|`), and filtered **`AggregateHybridQuery`** on a no-stopwords index; unit expectations are aligned with the new syntax.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 070a5b2bfc7c320412af97f0fc6eccb04cac1c06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -1569,5 +1569,5 @@ class TextQuery(BaseQuery):
             text = "(" + " | ".join(field_queries) + ")"
 
         if filter_expression and filter_expression != "*":
-            text += f" AND {filter_expression}"
+            text += f" {filter_expression}"
         return text
--- a/redisvl/query/query.py
+++ b/redisvl/query/query.py
@@ -1569,5 +1569,5 @@ class TextQuery(BaseQuery):
             text = "(" + " | ".join(field_queries) + ")"
 
         if filter_expression and filter_expression != "*":
-            text += f" {filter_expression}"
+            text += f" ({filter_expression})"
         return text
--- a/redisvl/utils/full_text_query_helper.py
+++ b/redisvl/utils/full_text_query_helper.py
@@ -62,7 +62,7 @@ class FullTextQueryHelper:
         query = f"(~@{text_field_name}:({self._tokenize_and_escape_query(text)})"
 
         if filter_expression and filter_expression != "*":
-            query += f" {filter_expression}"
+            query += f" ({filter_expression})"
 
         return query + ")"
 
--- a/redisvl/utils/full_text_query_helper.py
+++ b/redisvl/utils/full_text_query_helper.py
@@ -62,7 +62,7 @@ class FullTextQueryHelper:
         query = f"(~@{text_field_name}:({self._tokenize_and_escape_query(text)})"
 
         if filter_expression and filter_expression != "*":
-            query += f" AND {filter_expression}"
+            query += f" {filter_expression}"
 
         return query + ")"
 
--- a/tests/integration/test_stopwords_integration.py
+++ b/tests/integration/test_stopwords_integration.py
@@ -136,6 +136,11 @@ def filtered_queries_stopwords_disabled_index(redis_url, redis_test_name):
                 "team": "support",
                 "embedding": array_to_buffer([1.0, 0.0], "float32"),
             },
+            {
+                "text": "quarterly summary",
+                "team": "legal",
+                "embedding": array_to_buffer([1.0, 0.0], "float32"),
+            },
         ]
     )
 
@@ -253,6 +258,32 @@ def test_filtered_text_query_with_stopwords_disabled(
         text="handbook",
         text_field_name="text",
         filter_expression=Tag("team") == "docs",
+        return_fields=["text", "team"],
+        stopwords=None,
+    )
+
+    results = filtered_queries_stopwords_disabled_index.query(query)
+
+    assert len(results) == 1
+    assert results[0]["text"] == "reference handbook"
+    assert results[0]["team"] == "docs"
+
+
+def test_filtered_text_query_with_raw_string_union_filter(
+    filtered_queries_stopwords_disabled_index,
+):
+    """A raw-string union filter must stay grouped inside the intersection.
+
+    Redis binds '|' more loosely than whitespace intersection, so without
+    parenthesization this query parses as (handbook docs) | legal and wrongly
+    returns the legal document, which matches neither the text nor the first
+    tag. FilterExpression unions self-parenthesize, so only a raw string
+    exercises this.
+    """
+    query = TextQuery(
+        text="handbook",
+        text_field_name="text",
+        filter_expression="@team:{docs} | @team:{legal}",
         return_fields=["text", "team"],
         stopwords=None,
     )
--- a/tests/integration/test_stopwords_integration.py
+++ b/tests/integration/test_stopwords_integration.py
@@ -3,8 +3,11 @@
 import pytest
 
 from redisvl.index import SearchIndex
-from redisvl.query import FilterQuery
+from redisvl.query import AggregateHybridQuery, FilterQuery, TextQuery
+from redisvl.query.filter import Tag
+from redisvl.redis.utils import array_to_buffer
 from redisvl.schema import IndexSchema
+from tests.conftest import skip_if_redis_version_below
 
 
 @pytest.fixture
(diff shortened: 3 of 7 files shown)


redisvl/query/query.py
```diff
@@ -1569,5 +1569,5 @@ def _build_query_string(self) -> str:
             text = "(" + " | ".join(field_queries) + ")"
 
         if filter_expression and filter_expression != "*":
-            text += f" AND {filter_expression}"
+            text += f" {filter_expression}"
```
vishal-bala: Could you wrap this in parentheses — `text += f" ({filter_expression})"`, as your issue proposed?

Whitespace alone fixes the `AND` term but leaves a second precedence bug live. Redis binds `|` more loosely than whitespace intersection, so an unparenthesised multi-clause filter escapes the intersection. Unlike the `AND` bug, this one is **not** masked by the default stopword list, so it affects every user rather than only `STOPWORDS 0` indexes:

```
FT.CREATE demo ON HASH PREFIX 1 e: SCHEMA text TEXT a TAG b TAG
HSET e:1 text "the quick brown fox" a x b n
HSET e:2 text "nothing here"        a n b y

FT.SEARCH demo '@text:(fox) @a:{x} | @b:{y}'   DIALECT 2   ->  e:1, e:2
FT.SEARCH demo '@text:(fox) (@a:{x} | @b:{y})' DIALECT 2   ->  e:1
```

`e:2` contains neither `fox` nor `a:{x}`. `FT.EXPLAIN` on the first form shows why:

```
UNION {
  INTERSECT { @text:fox, TAG:@a { x } }
  TAG:@b { y }
}
```

To be clear, this is pre-existing rather than something you're introducing — on a default-stopwords index the old string parsed to the same plan once `AND` was dropped. But since the join is being rewritten anyway it's worth closing both at once, and the failure mode is nastier than the one you fixed: documents that match neither clause come back, which is harder to spot than zero results.

Only raw-string filters are affected. `FilterExpression.format_expression` already parenthesises composites, so `(Tag("a") == "x") | (Tag("b") == "y")` was never at risk.

redisvl/utils/full_text_query_helper.py
```diff
@@ -62,7 +62,7 @@ def build_query_string(
         query = f"(~@{text_field_name}:({self._tokenize_and_escape_query(text)})"
 
         if filter_expression and filter_expression != "*":
-            query += f" AND {filter_expression}"
+            query += f" {filter_expression}"
```
vishal-bala: Same parenthesisation here — `query += f" ({filter_expression})"`.

The impact is lower than in `TextQuery`, because the text clause is `~`-optional so `INTERSECT{~T, X}` returns the same document set as `X` whichever way the union binds. But the query plan is still wrong, and this helper is shared by both `AggregateHybridQuery` and the native `HybridQuery`, so I'd rather the two sites stay consistent than have one carry a latent precedence bug.

tests/integration/test_stopwords_integration.py
```diff
@@ -190,3 +243,46 @@ def test_stopwords_disabled_allows_searching_common_words(
...
+
+
+def test_filtered_text_query_with_stopwords_disabled(
+    filtered_queries_stopwords_disabled_index,
+):
+    """Filtered text queries should not add AND as a full-text search term."""
+    query = TextQuery(
+        text="handbook",
+        text_field_name="text",
+        filter_expression=Tag("team") == "docs",
```
vishal-bala, line 260: This filter is a single clause, so the test passes with or without the parenthesisation fix. Could you add one case with a union filter, so the precedence behaviour is pinned rather than incidental?

Something like `filter_expression=(Tag("team") == "docs") | (Tag("team") == "legal")` — though note that a `FilterExpression` union is self-parenthesising, so to exercise the actual bug the filter needs to be a raw string: `filter_expression="@team:{docs} | @team:{legal}"`, with a third document that matches neither the text nor either tag. Without the fix that document comes back; with it, it doesn't.

vishal-bala (commented):
Thanks for this — the diagnosis in #708 is exactly right, and the `FT.EXPLAIN` output showing `and` as a required term is what made this easy to act on. The two sites you picked are the two that matter most, and the new `STOPWORDS 0` integration tests are the right shape: two documents differing only in the filtered field, so they actually distinguish "filter applied" from "filter ignored".

Two things I'd like to see before this merges. Both are on lines you're already touching, and both leave a real bug live otherwise.

The first is parenthesising the filter, which I've commented on inline. Worth noting your own issue proposed `f" ({filter_expression})"` — I think the issue had it right and the PR is the weaker version of it.

The second can't be an inline comment because the file isn't in this diff. There's a third `AND` join site: `MultiVectorQuery._build_query_string` in `redisvl/query/aggregate.py` has two of them.

```python
383:  range_query = " AND ".join(range_queries)
390:  return f"({range_query}) AND ({filter_expression})"
```

Line 383 is the same failure mode you diagnosed, on a `STOPWORDS 0` index:

```
INTERSECT {
  VECTOR { ... @v1 }
  UNION { and, +and(expanded) }
  VECTOR { ... @v2 }
  ...
}
```

Line 390 is easy to miss — `grep '" AND "'` doesn't match it, because the literal there is `) AND (`. It needs the `AND` deleted separately; the groups are already parenthesised, so the parenthesisation change won't cover it. Both are one-line edits, and after them `grep -rn ' AND ' redisvl/query/ redisvl/utils/` should come back empty, which seems like the right bar for a PR closing #708.

For scope: we have follow-up work queued for the remaining call sites — `VectorQuery`'s KNN pre-filter (where a bare multi-clause pre-filter is a hard syntax error rather than silently wrong), `VectorRangeQuery`, and a shared helper so the join logic lives in one place instead of four. We'll open that as a downstream PR crediting this one, so please don't take any of it on here.

vishal-bala (approved):

PR #716, opened 2026-09-02
fix(mcp): fail closed when the tool scope gate cannot reach its auth config

vishal-bala:
## Motivation

The MCP scope gate can be disabled by a rename, silently. Four tool wrappers each carry an identical three-line prologue that reads `auth_config` off the server, pulls a scope name from it, and hands that to `ensure_tool_scope`. Every step of that chain tolerates a miss: `getattr(server, "auth_config", None)` yields `None`, the conditional expression then yields a `None` scope, and `ensure_tool_scope` returns early on a `None` scope. So renaming the server's `auth_config` attribute stops read and write scopes being enforced on every tool at once, with no error, no warning and no failing test.

The prologue is also redundant. `ensure_tool_scope` performs the same `getattr` lookup itself, so the caller's copy exists only to name which scope field applies.

### The scope name is resolved inside `auth.py`

`ensure_read_scope(server)` and `ensure_write_scope(server)` replace the prologue at all four call sites. Each resolves its own scope field from the server and delegates to `ensure_tool_scope`, so a wrapper no longer needs to know that a read tool reads `read_scope`.

```python
        auth_config = getattr(server, "auth_config", None)
        read_scope = auth_config.read_scope if auth_config is not None else None
        ensure_tool_scope(server, read_scope)
```

becomes

```python
        ensure_read_scope(server)
```

The field lookup is a bare `getattr(auth_config, attribute)` with no default, so renaming a field on `MCPAuthConfig` raises `AttributeError` rather than resolving to `None` and turning the gate into a no-op.

### An unreachable auth config now fails closed

`ensure_tool_scope` checks `_auth_enabled` first and, when auth is wired, treats a missing `auth_config` as an internal inconsistency rather than a reason to skip the check:

```python
    if not getattr(server, "_auth_enabled", False):
        return

    auth_config = getattr(server, "auth_config", None)
    if auth_config is None:
        raise RedisVLMCPError(..., code=MCPErrorCode.INTERNAL_ERROR, retryable=False)
```

The tokenless-request exit is unchanged and still returns early. That one is correct: an authenticated HTTP transport rejects tokenless requests before a tool body runs, so a missing token means stdio, where there is no scope to check.

### Secondary changes

- Four call sites reduced to one line each: `tools/search.py`, `tools/upsert.py`, `tools/list_indexes.py`, `tools/profiles.py`. No wrapper reads an `auth_config` attribute any more.
- Four tests in `tests/unit/test_mcp/test_auth_scope.py` covering scope resolution, the auth-disabled no-op, the unreachable-config failure and the renamed-field failure.

## Notes

Both new guards were mutation-checked: reverting the fail-closed raise alone fails exactly one test, and weakening the bare `getattr` to a defaulted one fails exactly one other. Neither guard is decorative.

`ensure_tool_scope` keeps its signature and stays public, because it is the right entry point for a caller that has a scope name in hand rather than a server to resolve one from. The new helpers are the preferred call-site form and its docstring says so.

## Next Steps

1. Verify the MCP suites:

```bash
uv run pytest tests/unit/test_mcp tests/integration/test_mcp -q
```

---

> [!NOTE]
> **Medium Risk**
> Changes authentication gating behavior for MCP tools when auth is enabled; misconfiguration now blocks tools rather than running ungated, which is safer but could surface new errors in production if wiring regresses.
> 
> **Overview**
> **Hardens MCP tool scope enforcement** so auth cannot be bypassed silently when server wiring is wrong, and **centralizes read/write scope checks** at the tool entry points.
> 
> When auth is enabled but `auth_config` is missing on the server, `ensure_tool_scope` now raises an **internal error** and refuses to run the tool instead of treating a missing config like “no scope configured.” Scope names are resolved via new **`ensure_read_scope`** / **`ensure_write_scope`** helpers (using strict attribute lookup on `MCPAuthConfig`), replacing duplicated three-line prologues in **list-indexes**, **search**, **profiles**, and **upsert-records**. Unit tests cover helper resolution, fail-closed unreachable config, and loud failure when config field names are renamed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 586d9d4cc46173601322e991b0ff0b10b25e424e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


limjoobin (approved):
LGTM. This PR replaces the scope prologue at the MCP tool wrappers to make sure that the gate fails closed when auth is enabled but its config can't be reached.

PR #724, opened 2026-09-04
fix: validate Geo filter arguments before they reach the query string

### LINKED ISSUE #723: GeoRadius coordinates are interpolated into the query string unvalidated (inconsistent with Num)
### Summary
`Geo` filter coordinates are formatted into the RediSearch query string without any type check or coercion. A coordinate supplied as a string can close its own geo clause and contribute arbitrary query syntax to the surrounding expression, including a `|` that detaches an enclosing filter.

This is the same containment gap that `Num` values had before #721 , but it is reachable with a plain string rather than requiring a crafted `int`/`float` subclass.

### Current behavior
`Geo.SUPPORTED_VAL_TYPES` is `(GeoSpec, type(None))`, so `_set_value` does refuse a raw operand — `Geo("location") == "anything"` correctly raises `TypeError`. But the check stops at the wrapper object.

`GeoSpec.__init__`  validates only `unit`:
```
def __init__(self, longitude: float, latitude: float, unit: str = "km"):
    if unit.lower() not in self.GEO_UNITS:
        raise ValueError(f"Unit must be one of {self.GEO_UNITS}")
    self._longitude = longitude
    self._latitude = latitude
    self._unit = unit.lower()
```
The coordinates are stored exactly as passed, `get_args` returns them unchanged, and `Geo.__str__` formats them through the `%s` slots of `@%s:[%s %s %i %s]`.

`radius` is unaffected, but only incidentally: its `%i` slot raises on a string. `longitude` and `latitude` have no equivalent protection.

### Reproduction

No Redis required; this is visible in the rendered filter string.
```
from redisvl.query.filter import Geo, GeoRadius

# The operand type is checked, so this is correctly refused:
Geo("location") == "anything"          # TypeError

# The coordinates inside an accepted GeoSpec are not checked:
lon = "-122.4194 37.7749 10 km] | @tenant_id:{victim} @location:[-122.4194"
print(Geo("location") == GeoRadius(lon, 37.7749, 10, "km"))
```

#### Expected behavior
A non-numeric coordinate should raise, consistently with how `Num` now coerces its values in #721 and with how a raw `Geo` operand is already rejected.

#### Actual behavior

The value is accepted and rendered as live query syntax:
```
@location:[-122.4194 37.7749 10 km] | @tenant_id:{victim} @location:[-122.4194 37.7749 10 km]
```

The injected text closes the geo clause and opens a second one that the template's own trailing `37.7749 10 km]` completes, so the result is syntactically valid and RediSearch parses it without error.

### Impact

The rendered `|` sits at the top level of the expression, and under `DIALECT 2` it binds looser than the implicit intersection. Any filter combined with the geo clause therefore stops constraining half the query. Verified against Redis 8.4.6: a query scoped with `Tag("tenant_id") == "acme"` and combined with a `GeoRadius` carrying the payload above returned rows from a second tenant as well as the intended one.

Coordinates are the parameter most likely to arrive from outside the application — query-string values for a proximity search are the obvious case, and they arrive as strings. Nothing in the signature, the type hints, or the docs indicates that the caller must coerce them first.

### Suggested fix

Coerce in the constructors, so a bad value fails at the call site and any operator that later formats the object is safe by construction: `longitude` and `latitude` in `GeoSpec.__init__`, `radius in GeoRadius.__init__`.

`Num._coerce_numeric` (added in #721 ) is the natural helper, but it is a `Num` classmethod and `GeoSpec` is not a `FilterField`, so sharing it means lifting it to a module-level function that takes the owner name as an argument rather than reading `cls.__name__`.

Two notes:

- Bounding longitude to `[-180, 180]` and latitude to `[-90, 90]` while the types are being checked costs almost nothing and additionally rejects `inf`, which a NaN-only check lets through and which renders literally into the query.
- If the %i in OPERATOR_MAP is changed to `%s` (see below), that edit must land together with the coercion, not before it. `%i` is currently the only thing refusing a string in the radius slot.

For tests, the `_StrOverridingFloat` pattern from #721  ports directly onto a coordinate, and a numpy-scalar case would confirm `numbers.Real` is wide enough here too.

limjoobin:
Fixes #723 

### Summary
`GeoSpec` validated only its `unit`. `longitude` and `latitude` were stored as given despite their `float` annotations, and every one of the four arguments is interpolated into a RediSearch query by `Geo.__str__`. 

A `str `coordinate carrying a `]` closed the geo clause and had its remainder parsed as syntax. Similar to #721, an injected `|` binds looser than the implicit space-AND, so it lifts to the root of the parse tree and any filter sharing the query stops constraining it:
```
crafted = "-122.4194 37.7749 10 km] | @secret:{leaked}"
str(Geo("location") == GeoRadius(crafted, 37.7749, 1))
# before: @location:[-122.4194 37.7749 10 km] | @secret:{leaked} 37.7749 1 km]
# after:  TypeError: GeoRadius longitude must be an int, a float, or another
#         numbers.Real; got str
```

The `unit` had the same shape by a different route, and this one is not in the issue as filed. `unit.lower()` not in `self.GEO_UNITS` is an equality test the caller controls, while `self._unit = unit.lower()` stored the caller's object. A `str` subclass is neutralised by `str.lower()` returning a builtin, but an object whose `lower()` returns `self` and whose `__eq__` matches `"km"` passed the check and then rendered its own `__str__`:
```
class Kilometres:
    def lower(self): return self
    def __eq__(self, other): return other == "km"
    def __hash__(self): return hash("km")
    def __str__(self): return "km] | @secret:{leaked}"

str(Geo("geo_field") == GeoRadius(1.0, 2.0, 3, Kilometres()))
# before: @geo_field:[1.0 2.0 3 km] | @secret:{leaked}]
# after:  @geo_field:[1.0 2.0 3 km]
```

Coordinates were also unbounded, so GeoRadius(9999, -9999, 1) and a NaN coordinate both rendered queries the server rejects, failing far from the line that caused them.

### Changes

1. Coordinates are coerced and range-checked at the constructor

`GeoSpec.__init__` routes both coordinates through `_coerce_to_number_within`, which coerces to a builtin and requires a finite value inside an inclusive range — `LONGITUDE_RANGE = (-180.0, 180.0)` and `LATITUDE_RANGE = (-90.0, 90.0)`, class attributes beside the existing `GEO_UNITS` so the numeric and unit domains share a namespace and a subclass can widen either.

Constructor-time rather than render-time is deliberate. The error names the offending argument at the caller's line instead of surfacing from inside `__str__`, and `str()` is on the query path so it should not re-validate on every call. The bounds are inclusive because the antimeridian and the poles are real places.

`isfinite` is checked separately rather than left to the range comparison, so an infinite bound could not admit an infinite value. That much is unreachable through this class's own finite ranges and is covered by a subclass test. The ordering is load-bearing for a second and entirely reachable reason, though: `isfinite` raises `OverflowError` on an int too large to convert to a float, so the range is compared first and rejects `10**400` as the documented `ValueError`.

2. The unit stores `GEO_UNITS' own spelling

`_canonical_unit` returns the matched element of `GEO_UNITS` rather than the value that matched it. The caller's `__eq__` still decides whether a unit matches; it no longer supplies what renders. There is deliberately no `isinstance(unit, str)` check, which would reject a legitimate `str` subclass — canonicalisation makes the identity of `lower()`'s return value irrelevant, since it is only ever compared, never rendered. `GEO_UNITS `stays the single source of truth for both the lookup and the error message.

Unit validation still runs before the coordinates, so a caller passing both a bad unit and a bad coordinate sees the error they saw before.

3. The radius is guarded explicitly, and stops being truncated

`%i` was doing two jobs badly. As a type guard it worked by accident, raising `TypeError` on a `str`, which left a format specifier load-bearing where nothing documented it. As a formatter it was simply wrong: it truncates toward zero, so a fractional radius silently queried a smaller circle than the caller asked for, and a sub-unit radius rendered `0`, which the server rejects outright.

Measured on 8.4.5 against a document 1.5 km from the centre:

| Caller writes | Rendered before | Hits | Rendered now | Hits |
|---|---|---|---|---|
| `radius=1.9, unit="km"` | `1 km` | 1 | `1.9 km` | 2 |
| `radius=0.5, unit="km"` | `0 km` | `Invalid GeoFilter radius` | `0.5 km` | 1 |

So the radius now renders through `%s`, and `GeoRadius` coerces it — that coercion is the type guard, explicitly, with a test that fails without it. An integral float renders as an int, because `repr` switches to exponent form at 1e16 and `1e+16` is a syntax error at DIALECT 1, the Redis 8 server default.

The radius is also bounded now: zero, negative and infinite are refused at the constructor. Measured, the server answers `Invalid GeoFilter radius` to zero and to a negative, and an infinite radius previously escaped as an `OverflowError` from inside `Geo.__str__`. The bound is a chained comparison rather than `math.isfinite`, which raises `OverflowError` on an int too large to convert — a radius has no upper bound to reject such a value first, unlike a coordinate.

4. The numeric coercion moves to module scope

`Num._coerce_numeric` from #721  becomes the module-level `_coerce_to_number`, because `GeoSpec` is not a `FilterField` and could not inherit it. `Num._coerce_numeric` stays as a one-line delegator binding `cls.__name__`, so error text is unchanged and `Timestamp` still reports its own name through `cls`. 

### Tests
30 geo cases in `tests/unit/test_filter.py`, sharing a `_geo_radius(**overrides)` builder so each row names only the argument it changes. `_StrOverridingInt`/`_StrOverridingFloat` are reused from #721  rather than duplicated per filter type. Every guard was mutation-checked by reverting it alone, including both `OPERATOR_MAP` templates separately, since `__ne__` is otherwise asserted nowhere.

The geo filter in `tests/integration/test_query.py` now uses a fractional radius, so a truncation regression fails against a real server rather than only against a string comparison. Verified: reverting the template alone turns that test red with `Invalid GeoFilter radius`.

## Release Notes

Geo filter arguments are now validated where they are supplied, rather than reaching the query string unchecked. A `str` coordinate could previously close the geo clause and have its remainder parsed as RediSearch syntax, so a filter built from untrusted input could be widened past the scope it was meant to enforce. Upgrade if any geo filter argument in your application originates from user input.

One change moves results silently and deserves attention before you upgrade. A fractional radius used to be truncated toward zero, so `GeoRadius(lon, lat, 1.9, "km")` queried a 1 km circle; it now queries 1.9 km and returns the rows it always should have. Nobody gets an error from this, only different rows.

Three changes are backwards-incompatible, all of them turning a value that used to build a broken query into an error at the line that supplied it. `Decimal` coordinates now raise `TypeError`, having rendered successfully before, so a coordinate arriving from a `NUMERIC` database column or from `json.loads(payload, parse_float=Decimal)` has to be passed through `float()` first. A zero, negative or infinite radius now raises `ValueError`, where zero and negative previously built a query the server rejected with `Invalid GeoFilter radius` and infinite raised `OverflowError` from inside the formatter. And a coordinate outside its range now raises `ValueError` instead of building a query the server refuses.

### DIFF
--- a/redisvl/query/filter.py
+++ b/redisvl/query/filter.py
@@ -237,16 +237,100 @@ class Tag(FilterField):
         )
 
 
-class GeoSpec:
-    GEO_UNITS = ["m", "km", "mi", "ft"]
+def _coerce_to_number(value: Any, owner: str, name: str) -> int | float:
+    """Return a numeric filter value as a plain int or float.
 
-    # class for the operand for FilterExpressions with Geo
+    Coercion rather than the type check is the guard: every numeric value is
+    formatted into the query string, so a subclass overriding __str__ would
+    satisfy isinstance and inject syntax. int() and float() return builtins
+    regardless, which strips the override.
+
+    A non-`numbers.Real` is rejected rather than converted, which is why
+    `Decimal` does not pass despite `float(Decimal("1.5"))` working: its exact
+    decimal arithmetic is different semantics, not a different spelling.
+
+    ``owner`` and ``name`` only label the error -- pass the class that took the
+    value and the parameter it arrived as.
+    """
+    if isinstance(value, numbers.Integral):
+        return int(value)
+    if isinstance(value, numbers.Real):
+        coerced = float(value)
+        if math.isnan(coerced):
+            # Renders `@field:[nan ...]`, which RediSearch rejects outright.
+            raise ValueError(f"{owner} {name} cannot be NaN")
+        return coerced
+    raise TypeError(
+        f"{owner} {name} must be an int, a float, or another "
+        f"numbers.Real; got {type(value).__name__}"
+    )
+
+
+def _coerce_to_number_within(
+    value: Any, owner: str, name: str, bounds: tuple[float, float]
+) -> int | float:
+    """Coerce, then require a finite value inside ``bounds``, both ends included.
+
+    Separate from _coerce_to_number because `Num` renders `-inf` and `+inf` by
+    design -- they are literal text in its own operator templates -- while a
+    coordinate has a domain and does not.
+
+    `isfinite` is checked separately rather than left to the range, so an
+    infinite bound could not admit an infinite value.
+    """
+    minimum, maximum = bounds
+    coerced = _coerce_to_number(value, owner, name)
+    if not math.isfinite(coerced) or not minimum <= coerced <= maximum:
+        raise ValueError(
+            f"{owner} {name} must be a finite number in "
+            f"[{minimum}, {maximum}]; got {coerced!r}"
+        )
+    return coerced
+
+
+class GeoSpec:
+    """The operand for a FilterExpression on a Geo field.
+
+    Every argument is formatted into the query string, so each is coerced and
+    range-checked here, at the constructor: the caller's own value selects or
+    scopes, and never itself renders. A `str` longitude used to interpolate raw,
+    so a value carrying `]` closed the geo clause and had its remainder parsed
+    as query syntax -- and an injected `|` binds looser than the implicit
+    space-AND, so it lifts to the root of the parse tree and any surrounding
+    filter stops constraining the query.
+    """
+
+    GEO_UNITS = ["m", "km", "mi", "ft"]
+    LONGITUDE_RANGE = (-180.0, 180.0)
+    LATITUDE_RANGE = (-90.0, 90.0)
+
     def __init__(self, longitude: float, latitude: float, unit: str = "km"):
-        if unit.lower() not in self.GEO_UNITS:
-            raise ValueError(f"Unit must be one of {self.GEO_UNITS}")
-        self._longitude = longitude
-        self._latitude = latitude
-        self._unit = unit.lower()
+        # Unit first, as before, so a caller passing both a bad unit and a bad
+        # coordinate still sees the error it saw yesterday.
+        self._unit = self._canonical_unit(unit)
+        owner = type(self).__name__
+        self._longitude = _coerce_to_number_within(
+            longitude, owner, "longitude", self.LONGITUDE_RANGE
+        )
+        self._latitude = _coerce_to_number_within(
+            latitude, owner, "latitude", self.LATITUDE_RANGE
+        )
+
+    @classmethod
+    def _canonical_unit(cls, unit: str) -> str:
+        """Return the library's own spelling of ``unit``.
+
+        `str.lower` already returns a builtin `str`, so a `str` subclass
+        overriding `__str__` cannot reach the query string. An object whose
+        `lower()` and `__eq__` merely *match* a known unit still could, and
+        returning the matched element of GEO_UNITS closes that -- without an
+        isinstance check, which would reject a legitimate `str` subclass.
+        """
+        requested = unit.lower()
+        for known in cls.GEO_UNITS:
+            if known == requested:
+                return known
+        raise ValueError(f"Unit must be one of {cls.GEO_UNITS}")
 
 
 class GeoRadius(GeoSpec):
@@ -262,16 +346,24 @@ class GeoRadius(GeoSpec):
         """Create a GeoRadius specification (GeoSpec)
 
         Args:
-            longitude (float): The longitude of the center of the radius.
-            latitude (float): The latitude of the center of the radius.
-            radius (int, optional): The radius of the circle. Defaults to 1.
+            longitude (float): The longitude of the center of the radius, in
+                degrees, from -180 to 180.
+            latitude (float): The latitude of the center of the radius, in
+                degrees, from -90 to 90.
+            radius (int, optional): The radius of the circle, in ``unit``.
+                Defaults to 1.
             unit (str, optional): The unit of the radius. Defaults to "km".
 
         Raises:
-            ValueError: If the unit is not one of "m", "km", "mi", or "ft".
+            TypeError: If a coordinate or the radius is not an ``int``, a
+                ``float``, or another ``numbers.Real``. numpy scalars qualify;
+                ``Decimal`` and ``str`` do not.
+            ValueError: If a coordinate is NaN, infinite, or outside its range,
+                if the radius is NaN, or if the unit is not one of "m", "km",
+                "mi", or "ft".
         """
         super().__init__(longitude, latitude, unit)
-        self._radius = radius
+        self._radius = _coerce_to_number(radius, type(self).__name__, "radius")
 
     def get_args(self) -> list[float | int | str]:
         return [self._longitude, self._latitude, self._radius, self._unit]
@@ -465,25 +557,13 @@ class Num(FilterField):
 
     @classmethod
     def _coerce_numeric(cls, value: Any, name: str = "value") -> int | float:
-        """Return a numeric filter value as a plain int or float.
+        """Bind this class's name to the shared numeric coercion.
 
-        Coercion rather than the type check is the guard: every numeric value is
-        formatted into the query string, so a subclass overriding __str__ would
-        satisfy isinstance and inject syntax. int() and float() return builtins
-        regardless, which strips the override.
+        The seam a subclass would widen, and what lets ``Timestamp`` report its
+        own name through ``cls``. See ``_coerce_to_number`` for why coercion,
+        rather than the type check, is the guard.
         """
-        if isinstance(value, numbers.Integral):
-            return int(value)
-        if isinstance(value, numbers.Real):
-            coerced = float(value)
-            if math.isnan(coerced):
-                # Renders `@field:[nan ...]`, which RediSearch rejects outright.
-                raise ValueError(f"{cls.__name__} {name} cannot be NaN")
-            return coerced
-        raise TypeError(
-            f"{cls.__name__} {name} must be an int, a float, or another "
-            f"numbers.Real; got {type(value).__name__}"
-        )
+        return _coerce_to_number(value, cls.__name__, name)
 
     def _set_value(
         self,
--- a/redisvl/query/filter.py
+++ b/redisvl/query/filter.py
@@ -255,7 +255,15 @@ def _coerce_to_number(value: Any, owner: str, name: str) -> int | float:
     if isinstance(value, numbers.Integral):
         return int(value)
     if isinstance(value, numbers.Real):
-        coerced = float(value)
+        try:
+            coerced = float(value)
+        except OverflowError:
+            # A `Real` too large to convert, such as `Fraction(10**400, 1)`.
+            # Finite, but unrepresentable, so it leaves by the documented door
+            # rather than as an `OverflowError` from the conversion.
+            raise ValueError(
+                f"{owner} {name} is too large to represent as a float"
+            ) from None
         if math.isnan(coerced):
             # Renders `@field:[nan ...]`, which RediSearch rejects outright.
             raise ValueError(f"{owner} {name} cannot be NaN")
@@ -276,11 +284,14 @@ def _coerce_to_number_within(
     coordinate has a domain and does not.
 
     `isfinite` is checked separately rather than left to the range, so an
-    infinite bound could not admit an infinite value.
+    infinite bound could not admit an infinite value. It is checked *second*
+    because `isfinite` itself raises `OverflowError` on an int too large to
+    convert, while the comparison handles one fine -- so the range rejects
+    `10**400` before finiteness is ever asked.
     """
     minimum, maximum = bounds
     coerced = _coerce_to_number(value, owner, name)
-    if not math.isfinite(coerced) or not minimum <= coerced <= maximum:
+    if not minimum <= coerced <= maximum or not math.isfinite(coerced):
         raise ValueError(
             f"{owner} {name} must be a finite number in "
             f"[{minimum}, {maximum}]; got {coerced!r}"
@@ -292,21 +303,24 @@ class GeoSpec:
     """The operand for a FilterExpression on a Geo field.
 
     Every argument is formatted into the query string, so each is coerced and
-    range-checked here, at the constructor: the caller's own value selects or
-    scopes, and never itself renders. A `str` longitude used to interpolate raw,
-    so a value carrying `]` closed the geo clause and had its remainder parsed
-    as query syntax -- and an injected `|` binds looser than the implicit
-    space-AND, so it lifts to the root of the parse tree and any surrounding
-    filter stops constraining the query.
+    checked at the constructor -- the coordinates and unit here, the radius in
+    `GeoRadius` -- so the caller's own value selects or scopes, and never itself
+    renders. Unchecked, a value carrying `]` would close the geo clause and
+    have its remainder parsed as query syntax, and an injected `|` binds looser
+    than the implicit space-AND, so it would lift to the root of the parse tree
+    and any surrounding filter would stop constraining the query.
     """
 
-    GEO_UNITS = ["m", "km", "mi", "ft"]
+    # Immutable because it is public, shared, and interpolated into the error
+    # message, so a mutable default is state any caller could corrupt for every
+    # other. Annotated loosely so a subclass can still widen it.
+    GEO_UNITS: tuple[str, ...] = ("m", "km", "mi", "ft")
     LONGITUDE_RANGE = (-180.0, 180.0)
     LATITUDE_RANGE = (-90.0, 90.0)
 
     def __init__(self, longitude: float, latitude: float, unit: str = "km"):
-        # Unit first, as before, so a caller passing both a bad unit and a bad
-        # coordinate still sees the error it saw yesterday.
+        # Unit first, so which error a caller sees when more than one argument
+        # is bad is fixed rather than incidental.
         self._unit = self._canonical_unit(unit)
         owner = type(self).__name__
         self._longitude = _coerce_to_number_within(
@@ -326,7 +340,15 @@ class GeoSpec:
         returning the matched element of GEO_UNITS closes that -- without an
         isinstance check, which would reject a legitimate `str` subclass.
         """
-        requested = unit.lower()
+        # Looked up rather than caught, so an `AttributeError` raised from
+        # inside a caller's own `lower()` is not misreported as a bad unit.
+        # Still no isinstance check: anything that spells itself lowercase is
+        # welcome, and one that cannot is a bad value rather than an internal
+        # error, so it leaves by the documented door.
+        lower = getattr(unit, "lower", None)
+        if lower is None:
+            raise ValueError(f"Unit must be one of {cls.GEO_UNITS}")
+        requested = lower()
         for known in cls.GEO_UNITS:
             if known == requested:
                 return known
@@ -340,7 +362,7 @@ class GeoRadius(GeoSpec):
         self,
         longitude: float,
         latitude: float,
-        radius: int = 1,
+        radius: float = 1,
         unit: str = "km",
     ):
         """Create a GeoRadius specification (GeoSpec)
@@ -350,8 +372,9 @@ class GeoRadius(GeoSpec):
                 degrees, from -180 to 180.
             latitude (float): The latitude of the center of the radius, in
                 degrees, from -90 to 90.
-            radius (int, optional): The radius of the circle, in ``unit``.
-                Defaults to 1.
+            radius (float, optional): The radius of the circle, in ``unit``,
+                greater than 0. Fractional radii are sent as given, so 0.5 with
+                a unit of "km" is half a kilometre. Defaults to 1.
             unit (str, optional): The unit of the radius. Defaults to "km".
 
         Raises:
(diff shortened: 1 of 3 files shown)


redisvl/query/filter.py
```diff
@@ -251,16 +337,24 @@ def __init__(
...
+            TypeError: If a coordinate or the radius is not an ``int``, a
+                ``float``, or another ``numbers.Real``. numpy scalars qualify;
+                ``Decimal`` and ``str`` do not.
+            ValueError: If a coordinate is NaN, infinite, or outside its range,
+                if the radius is NaN, or if the unit is not one of "m", "km",
+                "mi", or "ft".
         """
         super().__init__(longitude, latitude, unit)
-        self._radius = radius
+        self._radius = _coerce_to_number(radius, type(self).__name__, "radius")
```
vishal-bala: **Must fix.** The radius is the one argument left unfinished, and `_coerce_to_number_within` is defined ten lines up in this same diff.

Three separate consequences, all measured on Redis 8.4.5:

- `GeoRadius(-122.4194, 37.7749, 1.9, "km")` renders `1 km`, returning 1 hit where the requested `1.9 km` returns 2. Silent.
- `radius=0` and `radius=-5` construct and then fail server-side with `Invalid GeoFilter radius`, which is the ergonomic the rest of this PR exists to remove.
- `radius=float("inf")` constructs and raises `OverflowError: cannot convert float infinity to integer` from inside `Geo.__str__`. The server accepts `inf` happily (2 hits), so that ceiling is ours, not RediSearch's.

Two ways to close it, and they are not equivalent. Switching `%i` to `%s` in `OPERATOR_MAP` keeps fractional radii working, which the server supports and a caller converting metres to kilometres will want; it is safe now precisely because the value is a coerced builtin. Alternatively reject a non-integral radius, which keeps the rendering as-is but makes `0.5, "km"` unexpressible. Either way `radius` also wants a finiteness and non-negativity check here, and a `RADIUS_RANGE` sibling would match the pattern the coordinates already use.

No test covers a fractional radius, which is why CI is green on this.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
+            f"[{minimum}, {maximum}]; got {coerced!r}"
+        )
+    return coerced
+
+
 class GeoSpec:
+    """The operand for a FilterExpression on a Geo field.
+
+    Every argument is formatted into the query string, so each is coerced and
+    range-checked here, at the constructor: the caller's own value selects or
```
vishal-bala: **Must fix, alongside the radius comment.** "each is coerced and range-checked here, at the constructor" is not true of the radius, which is coerced only, and in `GeoRadius` rather than here. This sentence is the reason the gap is easy to miss: the next person auditing geo reads it and concludes the argument is guarded. Worth correcting even once the radius is bounded, since the check would then live in the subclass.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
 
+def _coerce_to_number(value: Any, owner: str, name: str) -> int | float:
+    """Return a numeric filter value as a plain int or float.
+
+    Coercion rather than the type check is the guard: every numeric value is
+    formatted into the query string, so a subclass overriding __str__ would
+    satisfy isinstance and inject syntax. int() and float() return builtins
+    regardless, which strips the override.
+
+    A non-`numbers.Real` is rejected rather than converted, which is why
```
vishal-bala, line 248: **Needs a release note.** `Decimal` coordinates render on `main` and raise here. The rejection is defensible, and the reasoning about exact decimal arithmetic being different semantics is fair: a `Decimal` subclass overriding `__str__` was genuinely injectable through `%s` before this change.

But `Decimal` arrives naturally from a `NUMERIC` column through psycopg or SQLAlchemy, and from `json.loads(payload, parse_float=Decimal)`, so this will break working code. #721 made the same call for `Num.between` and documented it under `## Release Notes` in the PR body, which is where `auto` picks it up. Adding that section here would keep the two consistent.

If the intent is to keep `Decimal` working, widening to `numbers.Number` and coercing through `float()` strips a `__str__` override exactly as it does on a `float` subclass, so the guard survives.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
+    Every argument is formatted into the query string, so each is coerced and
+    range-checked here, at the constructor: the caller's own value selects or
+    scopes, and never itself renders. A `str` longitude used to interpolate raw,
+    so a value carrying `]` closed the geo clause and had its remainder parsed
+    as query syntax -- and an injected `|` binds looser than the implicit
+    space-AND, so it lifts to the root of the parse tree and any surrounding
+    filter stops constraining the query.
+    """
+
     GEO_UNITS = ["m", "km", "mi", "ft"]
```
vishal-bala: **Worth a one-character fix.** `GEO_UNITS` is a mutable list and `_canonical_unit` calls `unit.lower()` before iterating it, so a `lower()` with a side effect can append its own payload to the allowlist and then match it. Measured: a `lower()` that does `GeoSpec.GEO_UNITS.append("km] | @secret:{leaked} @loc:[1.0 2.0 3 km")` and returns that element renders output byte-identical to the pre-fix exploit, and the list stays polluted process-wide for every later `GeoSpec`.

This works on `main` too, so it is not a regression, and reaching it needs an arbitrary object with attacker-chosen method bodies, and anyone who can pass that can mutate the list directly anyway. Raising it only because `_canonical_unit`'s docstring specifically claims to close the deceptive-unit route, and a tuple closes this last corner of it for free.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
+    Separate from _coerce_to_number because `Num` renders `-inf` and `+inf` by
+    design -- they are literal text in its own operator templates -- while a
+    coordinate has a domain and does not.
+
+    `isfinite` is checked separately rather than left to the range, so an
+    infinite bound could not admit an infinite value.
+    """
+    minimum, maximum = bounds
+    coerced = _coerce_to_number(value, owner, name)
+    if not math.isfinite(coerced) or not minimum <= coerced <= maximum:
```
vishal-bala: Minor. `math.isfinite(coerced)` runs before the range comparison, and it raises for a large `int` because the `numbers.Integral` branch returns the int unnarrowed. `GeoRadius(10**400, 0, 1)` gives `OverflowError: int too large to convert to float` rather than the documented `ValueError`. Swapping to `if not minimum <= coerced <= maximum or not math.isfinite(coerced)` gets the documented exception by short-circuit and still rejects `inf` under infinite bounds, so the subclass test keeps passing.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
+    def _canonical_unit(cls, unit: str) -> str:
+        """Return the library's own spelling of ``unit``.
+
+        `str.lower` already returns a builtin `str`, so a `str` subclass
+        overriding `__str__` cannot reach the query string. An object whose
+        `lower()` and `__eq__` merely *match* a known unit still could, and
+        returning the matched element of GEO_UNITS closes that -- without an
+        isinstance check, which would reject a legitimate `str` subclass.
+        """
+        requested = unit.lower()
```
vishal-bala: Minor. `unit=None` and `unit=5` raise `AttributeError: 'NoneType' object has no attribute 'lower'`, which the new `Raises:` block does not name. Pre-existing, but this method was rewritten here and now carries an explicit contract, and a caller guarding config with `except (TypeError, ValueError)` will let it through. Re-raising as `ValueError` keeps the deliberate absence of an `isinstance` check.

redisvl/query/filter.py
```diff
@@ -226,16 +228,100 @@ def __str__(self) -> str:
...
+    scopes, and never itself renders. A `str` longitude used to interpolate raw,
+    so a value carrying `]` closed the geo clause and had its remainder parsed
+    as query syntax -- and an injected `|` binds looser than the implicit
+    space-AND, so it lifts to the root of the parse tree and any surrounding
+    filter stops constraining the query.
+    """
+
     GEO_UNITS = ["m", "km", "mi", "ft"]
+    LONGITUDE_RANGE = (-180.0, 180.0)
+    LATITUDE_RANGE = (-90.0, 90.0)
```
vishal-bala, line 319: Minor, documentation only. This correctly mirrors the query parser, and the comment about the poles being real places is server-accurate: `@loc:[0 90 3 km]` parses and `90.00001` does not. Worth knowing that Redis cannot index above 85.05112878 degrees ([GEOADD](https://redis.io/docs/latest/commands/geoadd/)), and that limit silently governs query centres too. Measured, a centre at 85.0511288 returns 0 hits against a document at 85 degrees at any radius, including 20000 km. So `(85.05112878, 90]` renders clauses the server accepts and that can never match. Not worth narrowing the rejection, since nobody has a working query in that band, but a sentence in the docstring would save someone the experiment.

tests/unit/test_filter.py
```diff
@@ -321,6 +380,154 @@ def test_geo_filter(operation, expected):
     assert str(getattr(geo_f, operation)(geo_radius)) == expected
 
 
+def _geo_radius(**overrides) -> GeoRadius:
```
vishal-bala, line 419: Optional. The `_geo_radius(**overrides)` builder is a good pattern and the injection rows earn their place. A few rows re-assert one line through a second door, though, and the same matrix is already covered by `test_numeric_between_coerces_a_real_endpoint` and `test_numeric_refuses_an_unrenderable_value`: `latitude_nan` duplicates `longitude_nan` (`radius_nan` does earn its place as the other call site), `latitude_infinite` duplicates `longitude_infinite`, `longitude_none` and `longitude_list` share a `raise` with each other while `longitude_decimal` is the documented boundary, and one of the `latitude_above_range`/`latitude_below_range` pair differs from the longitude pair only in its constants.

`str_overriding_subclasses` is also half vacuous: `%i` converts through `__index__`/`__int__` and never consults `__str__`, so the radius half of that row passes with the radius coercion deleted. Making the radius fractional or infinite would make it load-bearing. And `radius_injects` names a vulnerability that never existed, given `%i`; the comment above it is accurate, but the id reads as evidence of a third vector.

vishal-bala (changes requested):
The injection is closed, and the reasoning in the description holds up. I reproduced the original hole against Redis 8.4.5 (search module 80410): a crafted latitude gives a `UNION` root under `DIALECT 2` and leaks 3 documents where 1 is correct, and this branch refuses it at the constructor. A separate pass tried nineteen bypasses of the new guard, including `numbers.Real` subclasses overriding `__str__`/`__repr__`/`__format__`, a nondeterministic `__float__` aimed at the range check, `__index__`-only integrals, numpy scalars, and `str` subclasses overriding `lower()`. Every one either rendered a plain numeric literal or raised. Validating in `GeoSpec.__init__` is also the right layer: the spec holds all three arguments before any field is in scope, so the constructor is the only place that can own the check.

Two things should change before this merges, and the first is the argument the description is most confident about.

**The radius is coerced but never bounded, and `%i` truncates it.** That combination turns a documented value class into silently wrong results. Point 3 of the description is right that the guard should not rest on a format specifier, but the fix makes the constructor accept floats that `%i` cannot render. Measured against a document planted roughly 1.5 km north of the centre:

| Caller writes | Renders as | Server returns |
|---|---|---|
| `radius=1.9, unit="km"` | `1 km` | 1 hit, losing the 1.5 km document |
| the query that was asked for | `1.9 km` | 2 hits |
| `radius=0.5, unit="km"` | `0 km` | `Invalid GeoFilter radius` |
| `radius=-5` | `-5 km` | `Invalid GeoFilter radius` |

The first row is the one that matters: no exception, no warning, just a narrower result set. RediSearch accepts a fractional radius, so this is the template discarding precision rather than a server limit.

**Second, rejecting `Decimal` is an unflagged behaviour break.** `Geo("l") == GeoRadius(Decimal("-122.4194"), Decimal("37.7749"), 1)` renders `@l:[-122.4194 37.7749 1 km]` on `main` and raises `TypeError` here. `Num` already rejected `Decimal`, so this is Geo-only, and #721 treated the same rejection as a breaking change and said so under `## Release Notes`. This one needs either the same note or a widened guard.

Everything else is optional. Longitude range, the unit list, case-insensitive unit handling, the inclusive endpoints and the `DIALECT 2` premise all check out against the documentation and the server, and the `LONGITUDE_RANGE`/`LATITUDE_RANGE` endpoints mirror the query parser exactly (`@loc:[180 0 3 km]` parses, `180.00001` does not).

vishal-bala:
Taking this over with Joo Bin's permission so it can ship today. Rebased onto `main` and pushed the fixes in 54271a0; his authorship is preserved on 0869a1f and the commit is co-authored.

The rebase needed `--onto`, because the branch carried the pre-squash version of #721 underneath the geo commit. A plain rebase would have replayed stale hunks over the merged fix.

**Both must-fix items are closed.**

The radius no longer truncates. `%i` became `%s`, which is safe because the coercion this PR added is the real type guard. Measured on 8.4.5 against a document 1.5 km out, `1.9 km` now returns 2 hits where it returned 1. An integral float renders as an int, because `repr` switches to exponent form at 1e16 and `1e+16` is a syntax error at DIALECT 1, the server default — without that, the flip would have regressed large radii, since `%i` rendered them digit by digit.

The radius is bounded: zero, negative and infinite are refused at the constructor. `GeoSpec`'s docstring no longer claims the radius is checked there.

`Decimal` is still rejected, matching #721, and the break is now documented under `## Release Notes` in the body along with the silent result-set change, which is the one nobody can catch with a try/except.

**The optional items are in too**, plus four things the second review round turned up:

- `GEO_UNITS` is an annotated tuple. The bare tuple broke subclass widening under mypy, which the body advertises.
- A non-string unit leaves as the documented `ValueError`, found by lookup rather than by catching, so an `AttributeError` from inside a caller's own `lower()` still propagates.
- `_coerce_to_number_within` compares the range before `math.isfinite`, and `float()` is guarded for a `Real` that is finite but unrepresentable, e.g. `Fraction(10**400, 1)`.
- The pole limit moved to the `Geo` class docstring. On `latitude` it contradicted the "-90 to 90" bullet it sat in, and it applies to every geo query rather than one argument.

**Two corrections to my own review, both found by mutation testing rather than reading.**

I was wrong to call `latitude_below_range` redundant. Widening `LATITUDE_RANGE` to `(-180.0, 90.0)` passed the whole suite without it, so it was the sole killer of that bound. Restored.

And the `__ne__` template's radius specifier was unpinned — reverting only that one stayed green, because `test_geo_filter[ne]` used a whole radius. It now uses `0.5`, so both templates are pinned and the dedicated render row is gone.

All eleven guards are mutation-checked by reverting each alone. The integration geo filter uses a fractional radius, so a truncation regression fails against a real server — verified: reverting the template turns it red with `Invalid GeoFilter radius`. 157 unit tests pass, mypy clean.

redisvl/query/filter.py
```diff
@@ -256,21 +363,53 @@ def __init__(
...
+                f"{owner} radius must be a finite number greater than 0; "
+                f"got {radius!r}"
+            )
+        if isinstance(radius, float) and radius.is_integer():
+            # `repr` switches to exponent form at 1e16, and `@loc:[... 1e+20 km]`
+            # is a syntax error at DIALECT 1 -- the Redis 8 server default, which
+            # is what a rendered filter meets if it is run outside a RedisVL
+            # query class. Every float that large is integral, so an int renders
+            # the same value without an exponent.
+            radius = int(radius)
```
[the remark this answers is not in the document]
vishal-bala: Not reproducible — measured on Redis 8.4.5 (search module 80410), a negative-exponent radius parses at **both** dialects:

```
DIALECT 1  radius=1e-05     => 0
DIALECT 1  radius=5e-324    => 0
DIALECT 1  radius=1e+16     => Syntax error at offset 24 near 1e
DIALECT 2  radius=1e-05     => 0
```

Only the positive-exponent form fails, which is why the normalisation targets integral floats: every float at or above 1e16 is integral, so that form cannot escape it. The `0` hit counts are the correct answer for a sub-millimetre radius, not a parse failure — the non-exponent spelling `0.0001` returns `0` too. Tiny coordinates are fine for the same reason: `@loc:[1e-05 2e-07 3 km]` parses at both dialects.

No code change. Added a clause to the comment in 6b53a4b noting that only a positive exponent is a problem, since the asymmetry is not obvious and is worth stating where someone would otherwise reach for a fix.

redisvl/mcp/tools/search.py
```diff
@@ -425,6 +430,21 @@ def _reject_escapable_filter(caller: FilterExpression) -> None:
...
+            # skip at the first *unescaped* quote is what keeps this failing
+            # closed: a value that did reach the rendering raw ends the phrase
+            # there, and the injected remainder is counted as before.
+            end = _find_unescaped(rendered, position + 1, '"')
+            if end is None:
+                # Unterminated phrase: the remainder is unparsable.
+                escaped = True
+                break
+            position = end + 1
+            continue
```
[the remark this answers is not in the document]
vishal-bala: Stale — this was raised against `c584314`, the pre-rebase head, which carried an orphaned early version of #721 underneath the geo commit. That branch has since been rebased onto `main`, and this PR does not touch `redisvl/mcp/` at all (`git diff origin/main --stat -- redisvl/mcp/` is empty).

The bug itself is real but already fixed, in #721 as `1866750` on `main`. Verified on this branch:

```
ACCEPTED  @content:("ends with a backslash\")
ACCEPTED  @content:("two backslashes\\")
```

`_reject_escapable_filter` deliberately does not honour a backslash as escaping the closing quote, because RediSearch does not either — `@f:("x\")` closes the phrase and yields the term `x\`. 529 tests pass across `test_filter.py` and `test_mcp/`.

No change needed.

vishal-bala (approved):

PR #730, opened 2026-09-06
fix(migration): support RESP3 key enumeration

### LINKED ISSUE #713: Migration readiness checks read FT.INFO with string keys, so under RESP3 they always report a healthy index
On a RESP3 connection the migration executor's readiness check concludes that every index is fully built with zero indexing failures, whatever the server actually reports. It therefore skips the SCAN fallback that exists precisely to avoid missing documents. This is a silent wrong answer rather than a crash, which is what makes it worth fixing ahead of louder failures: the check passes, the migration proceeds, and documents the server could not index are never enumerated.

## Reproduction

Measured on Redis 8.4.6 with redis-py 8.1.0, RedisVL at `edcc78e`. The index covers five hashes: three index cleanly, and two carry a malformed vector blob, so the server reports `hash_indexing_failures` of 2.

```
=== RESP2 ===
  actual hash_indexing_failures on server: 2
  info.get("hash_indexing_failures", 0) as the executor reads it: 2
  keys enumerated (5): ['f:bad0', 'f:bad1', 'f:ok0', 'f:ok1', 'f:ok2']

=== RESP3 ===
  actual hash_indexing_failures on server: 2
  info.get("hash_indexing_failures", 0) as the executor reads it: 0
  keys enumerated: none, raises KeyError: slice(1, None, None)
```

Under RESP2 the guard fires, logs `Index 'fail_idx' has 2 indexing failures. Using SCAN for complete enumeration.`, and enumerates all five keys including the two that failed to index. Under RESP3 the guard reads zero failures, skips SCAN, and hands control to the aggregate path.

## Cause

`FT.INFO` comes back with bytes keys under RESP3. Measured on both redis-py 6.3.0 and 8.1.0:

```python
[b'index_name', b'index_options', b'index_definition', b'attributes', b'num_docs', ...]
```

`redisvl/migration/executor.py:254` reads that dict with string keys:

```python
failures = int(info.get("hash_indexing_failures", 0) or 0)
percent_indexed = float(info.get("percent_indexed", 1.0) or 1.0)
```

Both lookups miss, so both defaults apply. The defaults were chosen to mean "nothing is wrong", so the guard concludes the index is complete and failure-free and takes the fast path. `redisvl/migration/async_executor.py:107` and `:108` are the same code, and `async_executor.py` repeats the shape around line 222, where `_enumerate_with_scan` reads prefixes out of the same raw `info` reply.

The proximate cause is that these sites call `client.ft(index_name).info()` directly rather than going through RedisVL's own wrapper. `SearchIndex._info` in `redisvl/index/index.py` passes the reply through `convert_bytes`, so `SearchIndex.info()` returns string keys and the right value on the same RESP3 client:

```python
>>> index.info()["hash_indexing_failures"]
2
```

Measured on the same build and the same index that produced the zero above. The planner already uses that wrapper, since `redisvl/migration/planner.py:216` calls `index.info()`, so its readiness checks at lines 157 and 174 read correctly. Only the executor's raw calls are wrong, in both the sync and async variants.

## Suggested fix

Route the executor's readiness checks through `convert_bytes`, or through `SearchIndex.info()`, so that the executor and the planner read `FT.INFO` the same way. Reading one field under two conventions in two files of the same package is the underlying problem, and normalising at a single point fixes every call site at once. The defaults are worth revisiting too: a lookup that misses should not be indistinguishable from a server reporting good news.

## Related

The path this bug hands control to is itself broken on RESP3: see #714, where the aggregate cursor read raises `KeyError`, which is the exception shown in the reproduction above. The two cannot both surface under RESP2, because a correct readiness check diverts to SCAN before the aggregate code runs. Under RESP3 this bug removes the guard and #714 then raises.

### LINKED ISSUE #714: Migration executor cannot read an FT.AGGREGATE cursor over a RESP3 connection, raising KeyError
`MigrationExecutor._enumerate_with_aggregate` cannot read an aggregate cursor over a RESP3 connection. It raises `KeyError: slice(1, None, None)`, and because the caller catches only `ResponseError` the exception escapes the SCAN fallback and aborts the enumeration outright.

## Reproduction

Measured on Redis 8.4.6 with redis-py 8.1.0, RedisVL at `edcc78e`.

```
KeyError: slice(1, None, None)
  redisvl/migration/executor.py:335, in _enumerate_with_aggregate
    for item in results_data[1:]:
```

`redisvl/migration/executor.py:317` issues the command through the client rather than through the search helper:

```python
result = client.execute_command(
    "FT.AGGREGATE", index_name, "*", "LOAD", "1", "__key",
    "WITHCURSOR", "COUNT", str(batch_size), "MAXIDLE", "300000",
)
```

Raw `execute_command` on the client gets no `Search` module response callbacks, because redis-py registers those on the object returned by `client.ft()`. The reply therefore arrives in whatever shape the protocol delivered it. Measured replies for one index holding three indexed documents:

`protocol=2`

```python
[[1, [b'__key', b'f:ok0'], [b'__key', b'f:ok1'], [b'__key', b'f:ok2']], 0]
```

`protocol=3`, and byte for byte the same with no explicit `protocol` argument on redis-py 8

```python
[{b'attributes': [], b'format': b'STRING',
  b'results': [{b'extra_attributes': {b'__key': b'f:ok0'}, b'values': []}, ...],
  b'total_results': 3, b'warning': []}, 0]
```

Both are two-element lists, so `results_data, cursor_id = result` at line 332 succeeds either way. The failure lands one line later: under RESP3 `results_data` is a dict, and `dict[1:]` raises `KeyError` with the slice object as the key. `redisvl/migration/async_executor.py:184` and `:187` are the same code and fail identically.

Worth noting that this is not a `TypeError`. `dict.__getitem__` accepts a slice quite happily and reports it as a missing key, which is a weaker diagnostic than a type error would have been and makes the traceback read as though an index were absent.

## Why this is more than a corner case

redis-py 8 negotiates RESP3 by default, so a caller who constructs a client with no protocol argument at all gets the failing shape. That is measured: the second reply above came from `Redis.from_url(url)` with no protocol keyword.

RedisVL currently masks the problem by forcing `kwargs.setdefault("protocol", 2)` at five points in `redisvl/redis/connection.py` (lines 640, 681, 734, 760 and 770), but that default reaches only clients RedisVL builds itself. A caller-supplied client keeps whatever protocol it negotiated. The default is also transitional, and this bug blocks removing it: the moment RedisVL stops pinning RESP2, every migration that reaches aggregate enumeration hits this path.

## The fallback does not catch it

`_enumerate_indexed_keys` wraps the aggregate call in `except ResponseError` at around line 286 of `executor.py`. `KeyError` is not a `ResponseError`, so the SCAN fallback that exists to rescue a failed aggregate enumeration never runs and the caller sees the raw traceback.

## Suggested fix

Issue the command through `client.ft(index_name).aggregate(...)` so that redis-py's own callbacks normalise the reply. If the raw command must stay, branch on the reply shape and read `results` and `extra_attributes` when handed a dict. Widening the caller's `except` clause beyond `ResponseError` is worth doing as well, though on its own it would only convert a crash into a quiet fall back to SCAN, which hides a genuine defect rather than fixing it.

## Related

The readiness check guarding this same enumeration has an independent RESP3 bug, #713, and the two are worth fixing in one pass. They cannot both surface under RESP2, because a correct readiness check diverts to SCAN before the aggregate code runs. Under RESP3 the first removes the guard and this one then raises.

## A smaller adjacent gap

`_parse_sentinel_url` in `redisvl/redis/connection.py:929` reads only `netloc` and `path`, never `query`, so a Sentinel URL that specifies a protocol has it silently ignored:

```python
>>> RedisConnectionFactory._parse_sentinel_url("redis+sentinel://host:26379/mymaster?protocol=3")
([('host', 26379)], 'mymaster', None, None, None)
```

Measured on the same build. The `protocol=3` parameter is dropped without a warning. This is a separate defect from the cursor bug and could reasonably be split out, but it belongs to the same cluster of inconsistent protocol handling across connection paths.

Kuang-xianxin:
Closes #713
Closes #714

On RESP3 connections, migration key enumeration reads byte-keyed `FT.INFO` maps as string-keyed dictionaries, treating failed or incomplete indexes as ready. Healthy indexes then hit a `KeyError` when the aggregate cursor's result map is sliced as a RESP2 row list.

The missing prefix also has a data-safety consequence: with RESP3 and `decode_responses=False`, if `FT.AGGREGATE` raises `ResponseError`, the SCAN fallback can use `*` instead of the source index's prefix. A field-rename migration can then rewrite a same-named field on unrelated applications' keys. This cross-prefix rewrite was reproduced on Redis 8.2.7 [during review](https://github.com/redis/redis-vl-python/pull/730#discussion_r3978238483).

Changes:
- Normalize `FT.INFO` before checking readiness and extracting SCAN prefixes, keeping fallback enumeration scoped to the source index.
- Preserve a real zero for `percent_indexed` in enumeration, both executors' `apply()` guards, and the planner. Schema-only migrations of zero-progress indexes enumerate the stored keys and pass their count to validation; planning also emits the still-building warning.
- Share aggregate key extraction between sync and async executors, handling RESP2 rows and RESP3 result maps while retaining cursor pagination and cleanup.

Validation:
- Before the review follow-up, three new regression cases failed at `d55117f`: sync and async `apply()` omitted the source count, and the planner omitted its warning for `percent_indexed=0.0`. Nine controls passed. All 12 pass after the fix.
- After merging current main (`2f8d3d3`), all 162 related enumeration, executor, planner, validation/cluster-scan, pattern-escaping and batch-migration unit tests pass with redis-py 8.1.0 and separately with 6.3.0. The SCAN mocks bind redis-py's real `scan_iter` implementations to exercise the updated traversal.
- Full source mypy passes (120 files); repository-wide Black/isort and scoped codespell pass.
- These tests use mocked Redis responses. Run with `pytest --confcutdir=tests/unit` to exclude the root Docker startup fixture. The Redis-backed integration suite was not run locally because Docker is unavailable; the Redis 8.2.7 reproduction above is the reviewer's evidence.

Suggested release label: auto:patch.

---

> [!NOTE]
> **Medium Risk**
> Changes core migration enumeration and index-readiness logic; incorrect behavior could skip documents or fail mid-migration, though scope is limited to wire-format handling and existing SCAN fallbacks.
> 
> **Overview**
> Fixes migration **key enumeration** on RESP3 (and mixed byte/string) Redis responses so drop/recreate migrations no longer mis-read index readiness or crash when paging `FT.AGGREGATE` cursors.
> 
> **FT.INFO** is normalized with `convert_bytes` before reading `hash_indexing_failures`, prefixes, and **`percent_indexed`**. A real **0** progress value is preserved (only `None` defaults to fully indexed), so unfinished indexes correctly trigger **SCAN** fallback instead of using the aggregate fast path.
> 
> Aggregate cursor pages share **`_extract_aggregate_keys`**, which reads document keys from RESP2 row lists or RESP3 `results` maps. Sync and async executors and the planner use the same **`percent_indexed`** rules when deciding enumeration and warnings.
> 
> Unit tests cover RESP2/RESP3 aggregate paging, SCAN fallback for failed/partial indexes, cursor cleanup, and schema-only migrations that must count keys when `num_docs` is misleading.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit affa15a69c19d599ec2bdf37a39d604a18c5ed50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


redisvl/migration/executor.py
```diff
@@ -250,9 +268,10 @@ def _enumerate_indexed_keys(
         # condition means FT.AGGREGATE would miss documents, so fall
         # back to SCAN for complete enumeration.
         try:
-            info = client.ft(index_name).info()
+            info = convert_bytes(client.ft(index_name).info())
             failures = int(info.get("hash_indexing_failures", 0) or 0)
-            percent_indexed = float(info.get("percent_indexed", 1.0) or 1.0)
+            progress = info.get("percent_indexed")
+            percent_indexed = float(progress) if progress is not None else 1.0
```
vishal-bala, line 274: The same expression survives at three sibling sites, all on the RESP3 path:

- `redisvl/migration/executor.py:1116-1119`
- `redisvl/migration/async_executor.py:828-831`
- `redisvl/migration/planner.py:173-175`

Those read `stats_snapshot`, which comes from `SearchIndex.info()` and is already normalised by `convert_bytes`, so the byte keys are not the problem there. The value type is. Measured on Redis 8.2.7 across all four protocol and `decode_responses` combinations, `percent_indexed` is the string `'1'` on RESP2 and the float `1.0` on RESP3, so a genuine zero is truthy on RESP2 and falsy on RESP3, where `or 1.0` reads it as fully built.

`executor.py:1119` feeds `needs_exact_count`, which gates `needs_enumeration` at `:1120-1125`; when false, the migration skips key enumeration altogether and validation falls back to the weaker `num_docs` comparison. That is the outer guard for the same condition this line now catches. `planner.py:175` suppresses the "Source index is still building" warning in `rvl migrate plan`.

```python
progress = snapshot.stats_snapshot.get("percent_indexed")
source_percent_indexed = float(progress) if progress is not None else 1.0
```

redisvl/migration/executor.py
```diff
@@ -144,6 +145,7 @@ def _map_keys_prefix(
 
 def _extract_prefixes_from_info(info: Any) -> List[str]:
     """Extract Redis Search index prefixes from dict or list FT.INFO shapes."""
+    info = convert_bytes(info)
```
vishal-bala, line 148: This is the most consequential change in the diff, and the description does not mention it. Worth adding, because it changes how urgent the fix looks to anyone triaging.

Measured pre-PR on RESP3 with `decode_responses=False`, `_extract_prefixes_from_info` returned `[]` against the byte-keyed `FT.INFO`. `build_scan_match_patterns([])` returns `["*"]` (`utils.py:86-92`), so `_enumerate_with_scan` enumerated the entire keyspace as the index's documents. `_rename_field_in_hash` at `:640` does not filter the key list it is handed, so an unrelated key carrying a field of the same name gets rewritten:

```
otherapp:user:9  before: {'email': 'x@y.z', 'title': 'someone elses data'}
                 after : {'email': 'x@y.z', 'headline': 'someone elses data'}
```

Post-fix, prefix extraction returns the real prefix in all four combinations. Reaching the bad path needs `FT.AGGREGATE` to fail with a `ResponseError` first, which the `_enumerate_with_aggregate` docstring treats as routine cursor expiry rather than an exotic condition.

redisvl/migration/executor.py
```diff
@@ -216,6 +218,22 @@ def _checkpoint_identity_matches(
     )
 
 
+def _extract_aggregate_keys(results_data: Any) -> Generator[str, None, None]:
+    """Read keys from raw RESP2 rows or a RESP3 aggregate result map."""
+    results_data = convert_bytes(results_data)
+    if isinstance(results_data, dict):
+        keys = (row["extra_attributes"]["__key"] for row in results_data["results"])
+    else:
+        # RESP2 starts with the row count, followed by field/value pairs.
```
vishal-bala: Measured on Redis 8.2.7, the leading element is not the row count: a two-row page came back as `[1, [b'__key', b'a:0'], [b'__key', b'a:1']]`, while the same page reported `total_results: 2` on RESP3. `[1:]` discards it either way, so only the comment is wrong.

tests/unit/test_migration_enumeration.py
```diff
@@ -0,0 +1,173 @@
...
+
+@pytest.mark.parametrize("decode_responses", [False, True])
+@pytest.mark.parametrize(
+    "readiness",
+    [
+        {"hash_indexing_failures": 2, "percent_indexed": 1.0},
+        {"hash_indexing_failures": 0, "percent_indexed": 0.5},
+        {"hash_indexing_failures": 0, "percent_indexed": 0.0},
+    ],
+    ids=["failed-documents", "partial-index", "empty-index"],
```
vishal-bala: Measured on Redis 8.2.7, a freshly created index with `num_docs: 0` reports `percent_indexed` of `'1'` on RESP2 and `1.0` on RESP3, with `indexing: 0`, so an empty index takes the aggregate fast path rather than the fallback. A `percent_indexed` of `0.0` means the background build has not started.

Your PR body already has the right word: "zero-progress". The id carries weight here because, with the executors reverted, `empty-index` is the only readiness value that fails at `decode_responses=True`, making it the one case that guards this change.

tests/unit/test_migration_enumeration.py
```diff
@@ -0,0 +1,173 @@
...
+        {**readiness, "index_definition": {"prefixes": ["doc:", "archive:"]}},
+        decode_responses,
+    )
+    client.scan.side_effect = [
+        (5, _wire(["archive:1"], decode_responses)),
+        (0, _wire(["archive:1", "archive:failed"], decode_responses)),
+        (0, _wire(["doc:pending"], decode_responses)),
+    ]
+    # The fast path would omit failed/pending documents, even if it did not crash.
+    client.execute_command.return_value = [[1, [b"__key", b"archive:1"]], 0]
```
vishal-bala: Dead: `assert_not_called()` at `:131` makes this `return_value` unreachable by construction. Worth dropping the assignment and moving the comment down beside that assertion, where a reader meets it after the thing it justifies.

Two redundant axes while you are here. With the executors reverted, all three `readiness` values fail at `decode_responses=False`, all catching the same byte-key bug, so two of the three earn nothing there. And `test_closing_enumeration_releases_cursor` fails only at `protocol=3`, where cursor release reads `result[1]` identically in both response models, so that axis is already covered by `test_enumerate_aggregate_cursor_pages`.

vishal-bala (changes requested):
Thanks for your contribution! This looks good, but it looks like there are a couple more spots where the fix should be applied and a couple details that could be clarified.

redisvl/migration/executor.py
Kuang-xianxin: Fixed all three sites in affa15a. Both apply guards and the planner now default only when progress is None. The new schema-only apply tests use num_docs=0 with two stored keys and assert that expected_source_count=2 reaches validation. These sync/async cases and the planner's numeric-zero case failed before the fix; RESP2 string-zero and complete/null controls passed. After syncing main's scan_iter change, all 162 related tests pass on both redis-py 8.1.0 and 6.3.0.

redisvl/migration/executor.py
Kuang-xianxin: Expanded the description with the cross-prefix field-rename consequence and its trigger: RESP3 byte-keyed FT.INFO plus an aggregate ResponseError could turn the fallback into SCAN *. It now explicitly attributes the Redis 8.2.7 reproduction to your review. The existing sync/async fallback tests continue to assert the source-prefix match after the aggregate error.

redisvl/migration/executor.py
Kuang-xianxin: Corrected the comment to call this leading metadata. The RESP2 page fixture now starts with 1 even for a two-row page, so pagination also verifies that extraction does not treat it as the row count.

tests/unit/test_migration_enumeration.py
Kuang-xianxin: Renamed the case to zero-progress. The new apply regressions likewise distinguish zero background-build progress from a fully built index with no indexed documents.

tests/unit/test_migration_enumeration.py
Kuang-xianxin: Removed the unreachable return_value and moved its explanation beside assert_not_called(). Kept one byte-key readiness case plus the three decoded semantic cases, and restricted the cursor-close test to RESP3; the pagination test still covers both protocols. This reduces the enumeration file from 28 to 22 cases while preserving those distinct regressions.

Kuang-xianxin:
Hi @vishal-bala, following up on `affa15a`: both apply guards and the planner now preserve numeric zero progress, and the remaining description, terminology, and test-cleanup points are addressed in the inline replies. The September 11 validation recorded 162 related passing tests on each of redis-py 8.1.0 and 6.3.0.

The head is unchanged, and Lint and Test Suite still require workflow approval. Could you take another look at the revisions, and could a maintainer approve the pending runs? Thanks!

